PR #2, opened 2021-04-22
[NRH-13]  DonationPage modeling

michael-caktus:
This chunk of work does the following:
- Creates a new app, `pages`
- Models the following:
  - `DonationPage`
  - `Template`
  - `Style`
  - `DonorBenefit`
  - `BenefitTier`
  - `Benefit`
- Stubs new app, `organizations`, including stubs for models (needed for relationships):
  - `Organization`
  - `RevenueProgram`
- Configures basic admin functionality, and two more advanced tools:
  - Make template from existing page
  - Make page from existing template
- Defines basic serializers for new models, including distinct Detail vs List serializers for some.
- Exposes resources in `pages` app via REST API.
- Tests, coverage @ >95% (Existing user model and manager will get tested during Auth work, and `organizations` app will get tested in next ticket. Should bring us closer to 100% coverage)


apps/organizations/admin.py
```diff
@@ -0,0 +1,15 @@
+from django.contrib import admin
+
+from apps.organizations.models import Organization, RevenueProgram
+
```
copelco, line 4: For ModelAdmins, I think it's helpful to, by default, add relevant fields to`list_display`, `list_filter`, `ordering`, and `search_fields`.

apps/pages/models.py
```diff
@@ -0,0 +1,159 @@
...
+        page.save()
+        return page_model.objects.get(pk=page.pk)
+
+
+class DonationPage(AbstractPage):
+    """
+    A DonationPage represents a single instance of a Donation Page.
+    """
+
+    name = models.CharField(max_length=255, null=True, blank=True)
```
copelco, line 66:
```suggestion
    name = models.CharField(max_length=255, blank=True)
```

apps/pages/models.py
```diff
@@ -0,0 +1,159 @@
...
+    class Meta:
+        unique_together = (
+            "name",
+            "organization",
+        )
+
+
+class DonorBenefit(models.Model):
+    name = models.CharField(max_length=255)
+    blurb = models.TextField(null=True, blank=True)
```
copelco, line 124:
```suggestion
    blurb = models.TextField(blank=True)
```

apps/pages/models.py
```diff
@@ -0,0 +1,159 @@
...
+    class Meta:
+        unique_together = (
+            "name",
+            "organization",
+        )
+
+
+class BenefitTier(models.Model):
+    name = models.CharField(max_length=255)
+    description = models.CharField(max_length=255, null=True, blank=True)
```
copelco, line 140:
```suggestion
    description = models.CharField(max_length=255, blank=True)
```

apps/pages/tests/test_views.py
```diff
@@ -0,0 +1,151 @@
...
+            list_url, {"title": "New DonationPage", "slug": "new-page", "organization": self.org.pk}
+        )
+        self.assertEqual(response.status_code, 201)
+        self.assertEqual(DonationPage.objects.count(), self.page_count + 1)
+
+    def test_page_update_updates_page(self):
+        page = self.pages[0]
+        old_page_title = page.title
+        old_page_pk = page.pk
+        detail_url = f"/api/v1/pages/{old_page_pk}/"
```
copelco, line 41: Can we use `reverse` here?

revengine/settings/base.py
```diff
@@ -148,7 +149,6 @@
 
 STATIC_ROOT = os.path.join(BASE_DIR, "public/static")
 STATIC_URL = "/static/"
-STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
```
copelco, line 151: We should sync on what Disco did with their project or revert to what Traffic Stops does here.

copelco (commented):
Added a few initial comments!

benjaminEwhite:
Initial comment: you need to cd into `spa` before running `npm install`. README.md should be updated to reflect this.

apps/pages/tests/test_views.py
```diff
@@ -0,0 +1,151 @@
...
+        new_title = "Old DonationPage With New Title"
+        response = self.client.patch(detail_url, {"title": new_title})
+        page = DonationPage.objects.filter(pk=old_page_pk).first()
```
benjaminEwhite, line 44:
```suggestion
        page = DonationPage.objects.get(pk=old_page_pk)
```

apps/pages/tests/test_views.py
```diff
@@ -0,0 +1,151 @@
...
+        template = self.templates[0]
+        old_template_pk = template.pk
+        detail_url = f"/api/v1/templates/{old_template_pk}/"
+        response = self.client.delete(detail_url)
+
+        templates = Template.objects.all()
+        self.assertEqual(len(templates), self.template_count - 1)
+        self.assertNotIn(old_template_pk, [p.pk for p in templates])
+
+    def test_template_list_uses_list_serializer(self):
```
benjaminEwhite, line 122: Unclear to me why it's important to test that template_list uses a particular implementation (and a side effect of that implementation, at that).

apps/pages/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class PageViewSet(viewsets.ModelViewSet):
+    queryset = DonationPage.objects.all()
+
+    def get_serializer_class(self):
+        if self.action == "retrieve":
+            return serializers.DonationPageDetailSerializer
+        return serializers.DonationPageListSerializer
+
```
benjaminEwhite, line 14: Nitpicking, but:

```suggestion
    def get_serializer_class(self):
        return serializers.DonationPageDetailSerializer if self.action == "retrieve" else serializers.DonationPageListSerializer

```

benjaminEwhite (commented):
Left a bunch of comments. Overall this looks good, but a little hard to wrap my head around which features I should be able to interact with in order to see how it all fits together. 

Looking at the AC for NRH-13 in JIRA, there's some bits about permissions for modification, and as best I can tell, this chunk of work doesn't address that. If this is meant to ship only a portion of the work described in that ticket, that's fine (or if the ticket should be rescoped, also fine).

apps/organizations/admin.py
```diff
@@ -0,0 +1,15 @@
...
+
+@admin.register(Organization)
+class OrganizationAdmin(admin.ModelAdmin):
+    class Meta:
+        model = Organization
```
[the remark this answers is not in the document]
copelco: Where? https://docs.djangoproject.com/en/2.2/ref/contrib/admin/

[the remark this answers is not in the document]
benjaminEwhite: Oh I may have misunderstood your comment: I thought you were saying to just remove `class`, not the entire block.

apps/pages/models.py
michael-caktus: I can't see why, in retrospect. I'll make it required.

apps/pages/models.py
```diff
@@ -0,0 +1,159 @@
...
+
+        return (instance, created)
+
+
+class Style(models.Model):
+    """
+    Ties a set of styles to a page. Discoverable by name, belonging to an Organization.
+    """
+
+    name = models.CharField(max_length=255)
```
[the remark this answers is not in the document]
michael-caktus: How do I make a field more required than the default `null=false, blank=false`? DRF serializers will treat this the same way as plain django Forms-- that is, will return a validation error if not provided.

apps/pages/tests/test_models.py
```diff
@@ -0,0 +1,82 @@
...
+        template, _ = self.instance.save_as_template(name=new_name)
+
+        self.assertEqual(template.name, new_name)
+
+
+class StyleTest(TestCase):
+    def setUp(self):
+        self.instance = StyleFactory()
+
+    def test_to_string(self):
```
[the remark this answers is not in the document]
michael-caktus: To me, the test says "This is what `to_string` should return for this model". Such that, if we depend on to_string returning a particular thing somewhere in the code, somebody changes it and we run tests, we'll see where/why it has broken.

apps/pages/models.py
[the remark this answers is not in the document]
benjaminEwhite: Oh wow. I didn't realize that fields are required by default in Django. Carry on!

apps/pages/tests/test_views.py
michael-caktus: Here I want to ensure that the correct serializer is being used based on the request method sent. It tests the method at apps.pages.views:10, and others like it.

apps/pages/models.py
```diff
@@ -0,0 +1,176 @@
...
+    elements = models.JSONField(null=True, blank=True)
+
+    show_benefits = models.BooleanField(default=False)
+
+    # lookup with donor_benefit.donationpages / donor_benefit.templates
+    donor_benefits = models.ForeignKey(
+        "pages.DonorBenefit",
+        null=True,
+        blank=True,
+        related_name="%(class)ss",
```
benjaminEwhite, line 27: This would be a more contemporary way of doing the string formatting here (and to my mind, more readable):

```suggestion
        related_name=f"{class}s",
```

apps/pages/models.py
```diff
@@ -0,0 +1,159 @@
...
+        template_exists = Template.objects.filter(name=template.name).exists()
+        created = False
+
+        if not template_exists:
+            template.save()
+            created = True
+
+        instance = Template.objects.filter(name=template.name).first()
+
+        return (instance, created)
```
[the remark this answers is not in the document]
benjaminEwhite: Any particular reason you don't want to use get_or_create? In addition to being more succinct, according to docs, it also protects against a possible race condition:  https://docs.djangoproject.com/en/3.2/ref/models/querysets/#get-or-create

apps/pages/models.py
```diff
@@ -0,0 +1,176 @@
...
+            "organization",
+        )
+
+
+class BenefitTier(IndexedTimeStampedModel):
+    name = models.CharField(max_length=255)
+    description = models.CharField(max_length=255, blank=True)
+    benefits = models.ManyToManyField("pages.Benefit")
+    organization = models.ForeignKey(
+        "organizations.Organization", related_name="benefittiers", on_delete=models.CASCADE
```
benjaminEwhite, line 156:
```suggestion
        "organizations.Organization", related_name="benefit_tiers", on_delete=models.CASCADE
```

apps/pages/models.py
```diff
@@ -0,0 +1,176 @@
...
+            "organization",
+        )
+
+
+class DonorBenefit(IndexedTimeStampedModel):
+    name = models.CharField(max_length=255)
+    blurb = models.TextField(blank=True)
+    tiers = models.ManyToManyField("pages.BenefitTier")
+    organization = models.ForeignKey(
+        "organizations.Organization", related_name="donorbenefits", on_delete=models.CASCADE
```
benjaminEwhite, line 138:
```suggestion
        "organizations.Organization", related_name="donor_benefits", on_delete=models.CASCADE
```

apps/pages/models.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    A DonationPage represents a single instance of a Donation Page.
+    """
+
+    slug = models.SlugField(unique=True)
+
+    revenue_program = models.ForeignKey(
+        "organizations.RevenueProgram",
+        null=True,
+        related_name="donationpages",
```
benjaminEwhite, line 79:
```suggestion
        related_name="donation_pages",
```

benjaminEwhite (commented):
A couple small suggestions re: using snake case for related names, and question re: get_or_create, but this looks fine.

michael-caktus:
I took out the underscore for the related names to match the pattern of organization.donationpages that is set using related_name="%(class)ss". It felt confusing to have organizations.donationpages but revenue_programs.donation_pages.

As for the `get_or_create`, what I needed was somethign that didn't need to be updated every time a field was added to AbstractPage. I tried to user get_or_create originally, but I couldn't get it to fit the need.

apps/pages/models.py
michael-caktus: That won't actually work here. It expects a `class` variable to be defined, whereas %(class)s seems to be lazy.

PR #6, opened 2021-04-26
Styles inline api explorer

michael-caktus:
Fixes two issues discovered in demo today:
1. "styles" inline missing from DonationPage and Template admins
2. API explorer not hooked up.


JeremyGibson (approved):
:shipit:

PR #11, opened 2021-05-09
Nrh 14  plans features

JeremyGibson:
Adds two types of feature, but the boolean is meant as a kind of thought experiment. Since we don't have a concrete business case, it is kind of a place holder.

The Features _feature_ takes the approach that there will pretty much always need to be a code change when one of these is added, and that each feature is essentially unique on the feature properties. This assumes that Features will most likely be fairly static (i.e. tied to pricing tiers), and that there probably won't be many of them (greater than 50). This solves the problem of needing to find all features when a call is made, since each feature will reasonably be targeted at specific functions of the site like page creation. Therefore a LIMIT_PAGE FeatureType can have a value, say `3` that really only needs to be checked if an organization is creating a new page. 

This potentially **could** be a Hub admin type system where Features are defined on the fly, but that, I think requires a much more detailed examination of the business cases and would, IMHO, create a much more complex system prone to bugs caused by mismatches in Dashboard assumptions and API reality.


apps/pages/models.py
```diff
@@ -81,10 +84,28 @@ class DonationPage(AbstractPage):
...
+    @property
+    def total_pages(self):
+        return DonationPage.objects.filter(organization=self.organization).count()
+
     @property
     def is_live(self):
         return bool(self.published_date and self.published_date <= timezone.now())
 
+    def save(self, *args, **kwargs):
+        if limit := self.has_page_limit():
```
benjaminEwhite, line 102: TIL about the walrus operator!

apps/pages/models.py
```diff
@@ -81,10 +84,28 @@ class DonationPage(AbstractPage):
     def __str__(self):
         return f"{self.title} - {self.slug}"
 
+    def has_page_limit(self):
+        try:
+            return self.organization.plan.features.filter(feature_type=Feature.FeatureType.PAGE_LIMIT).first()
+        except AttributeError as e:
+            pass
```
benjaminEwhite, line 91:
```suggestion
        return Feature.objects.filter(feature_type=Feature.FeatureType.PAGE_LIMIT, plans__organization=self.id).first()
```

benjaminEwhite: Believe ^ is more "djangonic". Also, in the case where there is > 1, probably want an explicit order by here.

apps/pages/models.py
```diff
@@ -81,10 +84,28 @@ class DonationPage(AbstractPage):
     def __str__(self):
         return f"{self.title} - {self.slug}"
 
+    def has_page_limit(self):
```
benjaminEwhite, line 87: Might make sense to make this a property instead (`page_limit`)?

apps/pages/models.py
```diff
@@ -81,10 +84,28 @@ class DonationPage(AbstractPage):
...
+    def total_pages(self):
+        return DonationPage.objects.filter(organization=self.organization).count()
+
     @property
     def is_live(self):
         return bool(self.published_date and self.published_date <= timezone.now())
 
+    def save(self, *args, **kwargs):
+        if limit := self.has_page_limit():
+            if self.total_pages + 1 > int(limit.feature_value):
```
benjaminEwhite, line 103: Woudn't this block saving when updating an existing page if already at limit?

apps/pages/models.py
benjaminEwhite: Probably should test this method.

benjaminEwhite:
One more general comment: I was wondering if perhaps there should be "soft" delete model across the models here (and others more generally) whereby instances are made inactive instead of fully deleted. Seems like it might be important for auditing and/or offboarding purposes.

apps/organizations/models.py
```diff
@@ -56,23 +71,26 @@ def __str__(self):
...
     def save(self, *args, **kwargs):
         if not self.id:
-            slug = normalize_slug(slugify(self.name, allow_unicode=True))
-            self.slug = f"{self.organization.slug}-{slug}"
-        super(RevenueProgram, self).save(*args, **kwargs)
+            self.slug = normalize_slug(
+                self.name,
+                self.slug,
+            )
+            self.slug = f"{self.organization.slug}-{self.slug}"
```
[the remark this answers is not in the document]
benjaminEwhite: Should this be wrapped in `normalize_slug` to ensure that the org slug + new name/slug slug aren't together greater than max length 100?

apps/pages/models.py
JeremyGibson: Order by shouldn't be an issue here, but we'll find that out later. However you are right that is more django-y

apps/pages/models.py
JeremyGibson: Point one. This is such a fuzzy area, but to me `has_page_limit` is asking a question? Maybe it does maybe it doesn't. Since it is asking for information with a complex answer rather than a fact like `is_live` or `total_pages`, I tend to think of those as methods rather than properties.

Point two: This is tested in `test_feature_limits_page_creation`. It is not an explicit test of this function, but more of an integration test. Happy to make more explicit if you think its more helpful.

apps/organizations/tests/test_org_views.py
```diff
@@ -158,3 +157,40 @@ def test_list_returns_expected_count(self):
...
+                FeatureFactory(feature_type=Feature.FeatureType.BOOLEAN, feature_value=v)
+            except django.core.exceptions.ValidationError as e:
+                self.fail(f"Save raised a validation error on expected valid inputs: {e.message}")
+
+    def test_validation_error_on_bad_input(self):
+        invalid = ["-1", "S"]
+        for v in iter(invalid):
+            with self.assertRaises(django.core.exceptions.ValidationError) as cm:
+                FeatureFactory(feature_type=Feature.FeatureType.BOOLEAN, feature_value=v)
+            self.assertEqual(cm.exception.message, "The feature type 'Boolean' requires one of the following [1,0,t,f]")
```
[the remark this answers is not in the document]
JeremyGibson: Not sure the issue here. Since the set of invalid values is vast the idea of the test is to test that some invalid values are being caught with both a string and an integer outside of the valid range. However this did make me think that I should add a positive integer and maybe a decimal.

JeremyGibson:
@benjaminEwhite Regarding your last general comment. I agree that might be a good idea, but probably should be discussed with NRH and planned out in another follow on ticket.

I took the requirement in the AC as meaning a feature is "activated" or "deactivated" on a plan. Meaning the feature is not deleted just removed from the plan attached to the organization.

apps/pages/models.py
```diff
@@ -99,7 +96,8 @@ def is_live(self):
         return bool(self.published_date and self.published_date <= timezone.now())
 
     def save(self, *args, **kwargs):
-        if limit := self.has_page_limit():
+        limit = self.has_page_limit()
```
benjaminEwhite, line 99: But I only just met the walrus operator :(

benjaminEwhite (approved):
LGTM! :shipit:

PR #17, opened 2021-05-18
[NRH-18] Quarantine

michael-caktus:
Integrates NRH [BadActor API](https://github.com/newsrevenuehub/sjp).

Makes the checkout form on the frontend a little bit more dynamic, for testing.

In order to implement the BadActor check AFTER we collect info from the donor but BEFORE we confirm the payment, we need to force an interstitial step. Stripe PaymentIntents can be created with `capture_method` `manual` or `automatic`. `automatic` is the default, and it will finish the transaction without any further steps necessary on our end. `manual` requires that we manually call PaymentIntent.capture or PaymentIntent.cancel.


apps/contributions/payment_intent.py
```diff
@@ -0,0 +1,193 @@
...
+
+class PaymentIntent:
+    serializer_class = None
+    bad_actor_score = None
+    bad_actor_response = None
+    validated_data = None
+    flagged = None
+    contribution = None
+
+    def __init__(self, data=None, contribution=None):
```
benjaminEwhite, line 26: I may have missed it, but didn't see any use of this with a contribution vs. data.

apps/contributions/payment_intent.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def get_or_create_contributor(self):
+        contributor, _ = Contributor.objects.get_or_create(email=self.validated_data["email"])
+        return contributor
+
+    def should_flag(self):
+        """
+        BadActor API returns an "overall_judgement", between 0-5.
+        """
+        return self.bad_actor_score >= 3
```
benjaminEwhite, line 74: worthwhile to put 3 in a magic constant somewhere?

apps/contributions/tests/test_payment_intent.py
```diff
@@ -0,0 +1,276 @@
...
+    def test_validate_fail(self):
+        bad_data = self.data.copy()
+        bad_data.pop("referer")
+        pi = self._instantiate_pi_with_data(data=bad_data)
+        with self.assertRaises(ValidationError) as v_error:
+            pi.validate()
+        self.assertIn("referer", v_error.exception.detail)
+        self.assertEqual(str(v_error.exception.detail["referer"][0]), "This field is required.")
+
+    def test_calling_ba_before_validate_throws_error(self):
```
benjaminEwhite, line 85: More verbose name here would be quite helpful -- `test_calling_get_bad_actor_score_before_validate_throws_error`.  IMO, you can't go wrong with long descriptive test names, which can obviate the need for doc strings, and make intent clear.  

Even having a good understanding of context, I had to think for a moment to realize what "ba" meant here.

apps/contributions/tests/test_payment_intent.py
```diff
@@ -0,0 +1,276 @@
...
+
+    @responses.activate
+    def test_get_bad_actor_score_error(self):
+        self._create_mock_ba_response(status_code=500)
+        pi = self._instantiate_pi_with_data()
+        self.assertIsNone(pi.validate())
+        pi.get_bad_actor_score()
+        self.assertFalse(pi.flagged)
+
+    def test_calling_create_pi_before_ba_throws_error(self):
```
benjaminEwhite, line 123: seems like a good convention might be `test_<method_name>_when_<scenario>`. So `test_create_payment_intent_when_get_bad_actor_score_not_called`

apps/contributions/tests/test_payment_intent.py
```diff
@@ -0,0 +1,276 @@
...
+        pi = self._instantiate_pi_with_data()
+        with self.assertRaises(ValueError) as e:
+            pi.create_payment_intent()
+        self.assertEqual(
+            str(e.exception), "PaymentIntent must call 'get_bad_actor_score' before creating payment intent"
+        )
+
+    @responses.activate
+    @patch("stripe.PaymentIntent.create", side_effect=MockPaymentIntent)
+    def test_create_payment_intent_single_flagged(self, mock_stripe_create_pi):
```
benjaminEwhite, line 133:
```suggestion
    def test_create_payment_intent_when_single_flagged_contribution(self, mock_stripe_create_pi):
```

apps/contributions/tests/test_payment_intent.py
```diff
@@ -0,0 +1,276 @@
...
+        mock_stripe_create_pi.assert_called_once_with(
+            amount=self.amount,
+            currency="usd",
+            payment_method_types=["card"],
+            api_key=fake_api_key,
+            stripe_account=self.organization.stripe_account_id,
+            capture_method="manual",
+        )
+        # New contribution is created...
+        new_contribution = Contribution.objects.filter(amount=1099).first()
```
benjaminEwhite, line 152: There's an edge case where the ContributionFactory initially creates a contribution with this amount in the setUp step, in which case .first() would be false positive for evidence of a new contribution being created. Perhaps adding a contribution.reason and filtering on that might make more sense? Or else, asserting that no 1099 contribution was originally there -- or that the count of 1099 contributions has increased by one.

apps/contributions/tests/test_payment_intent.py
```diff
@@ -0,0 +1,276 @@
...
+        mock_stripe_create_pi.assert_called_once_with(
+            amount=self.amount,
+            currency="usd",
+            payment_method_types=["card"],
+            api_key=fake_api_key,
+            stripe_account=self.organization.stripe_account_id,
+            capture_method="automatic",
+        )
+        # New contribution is created...
+        new_contribution = Contribution.objects.filter(amount=1099).first()
```
benjaminEwhite, line 178: This test will fail if ContributionFactory chooses 1099 for amount.

apps/contributions/payment_intent.py
```diff
@@ -104,7 +104,7 @@ def create_contribution(self, organization, payment_intent):
             bad_actor_response=self.bad_actor_response,
         )
 
-    def create_payment_intent(self):
+    def ensure_bad_actor_score(self):
```
benjaminEwhite, line 107: LOL

apps/contributions/tests/test_payment_intent.py
```diff
@@ -130,14 +130,18 @@ def test_calling_create_pi_before_ba_throws_error(self):
...
         data = self.data
         data["payment_type"] = StripePaymentIntentSerializer.PAYMENT_TYPE_SINGLE[0]
         pi = self._instantiate_pi_with_data(data=data)
         pi.validate()
         self._create_mock_ba_response(target_score=4)
         pi.get_bad_actor_score()
 
+        # Assert that this contribution is not there yet
+        new_contribution = Contribution.objects.filter(amount=1099).first()
+        self.assertIsNone(new_contribution)
```
benjaminEwhite, line 143:
```suggestion
        self.assertFalse(Contribution.objects.filter(amount=1099).exists()
```

benjaminEwhite (approved):
Left one  suggestion, but I'm 👍

PR #19, opened 2021-05-18
NRH-31  - Password reset flow for org admins

benjaminEwhite:
Adds password reset flow (minimally styled) that ends with OrgAdmins getting redirected to SPA, and HubAdmins getting redirected to Django Admin login page.


michael-caktus:
I'm wondering if it wouldn't be easier to follow the rule: if the password-reset came from django admin login form, redirect to django admin. If it comes from the SPA login, redirect to SPA login. There's an entry in the django docs about [adding `/admin` specific forgot-pw](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#adding-a-password-reset-feature).  Just a thought-- seems like this works fine. I think the benefit of including the `'admin_password_reset'` named view as documented is that it causes the "Forgot password?" link to appear in django admin automatically.

apps/users/views.py
```diff
@@ -11,6 +13,13 @@
 user_model = get_user_model()
 
 
+class UserTypeSensitivePasswordResetConfirm(PasswordResetConfirmView):
+    def form_valid(self, form):
+        if self.user.organizations.exists():
+            self.success_url = reverse_lazy("orgadmin_password_reset_complete")
+        return super().form_valid(form)
+
+
```
michael-caktus, line 22: I wonder if it would be wise to do a bit of "session invalidation" here. `super()` will handle the django-admin session invalidation. On the SPA side, we potentially have an Authorization cookie with a now-stale JWT. So let's invalidate that cookie--
```suggestion
    def dispatch(self, *args, **kwargs):
        response = super().dispatch(*args, **kwargs)
        # Invalidate Authorization cookie in client browser by setting its value to gibersih and it's max_age to 0
        response.set_cookie(settings.AUTH_COOKIE_KEY, "NoTaVaLiDtOkEn", max_age=0)
        return response
```

michael-caktus (changes requested):

michael-caktus:
Could we also take this time to make sure that a "Forgot password?" link appears in both the Django-admin login form and the SPA login form?

michael-caktus (approved):

PR #25, opened 2021-05-24
[NRH-59] ThankYou Page

michael-caktus:
Adds a basic default Thank You page.  Adds some fields to models:

- Organization.default_thank_you_redirect
- Organization.default_post_thank_you_redirect
- DonationPage.thank_you_redirect
- DonationPage.post_thank_you_redirect

For `thank_you_redirect`, frontend sends donors there after a payment if the field is defined, else sends them to the basic default.  The `post_thank_you_redirect` is an optional link that appears on the thank you page to take them back to the website they came from (the orgs website). In the future, we can send them there automatically after a timeout, if we wish...


apps/contributions/admin.py
```diff
@@ -32,7 +32,9 @@ def lookups(self, request, model_admin):
         return Contribution.BAD_ACTOR_SCORES
 
     def queryset(self, request, queryset):
-        return queryset.filter(bad_actor_score=self.value())
+        if self.value():
```
michael-caktus, line 35: Having previously left this check out, Contribution list-view was always filtered on blank bad_actor_scores

apps/contributions/payment_intent.py
```diff
@@ -73,17 +86,27 @@ def should_flag(self):
         """
         return self.bad_actor_score >= settings.BAD_ACTOR_FAIL_ABOVE
 
-    def get_organization(self):
+    def get_revenue_program(self):
```
michael-caktus, line 89: Since we don't provide org_slug in the payment intent create view anymore, we need to find rev program here, and then use it to get org.  It's possible that rev program is already defined on an instance, so check that first.

apps/contributions/serializers.py
```diff
@@ -24,8 +24,8 @@ class StripePaymentIntentSerializer(serializers.Serializer):
     amount = serializers.IntegerField()
     reason = serializers.CharField(max_length=255)
 
-    organization_slug = serializers.SlugField()
-    donation_page_slug = serializers.SlugField()
+    revenue_program_slug = serializers.SlugField()
+    donation_page_slug = serializers.SlugField(required=False)
```
michael-caktus, line 28: This must be optional, since it's optional on the front-end

apps/contributions/views.py
```diff
@@ -34,8 +36,12 @@ def stripe_payment_intent(request):
     # Performs request to BadActor API
     stripe_pi.get_bad_actor_score()
 
-    # Create payment intent with Stripe, associated local models
-    stripe_payment_intent = stripe_pi.create_payment_intent()
+    try:
```
michael-caktus, line 39: Without this try-except, this view can return a misleading success status, causing the frontend to hang

apps/organizations/views.py
```diff
@@ -45,6 +52,35 @@ class OrganizationViewSet(OrganizationLimitedListView, viewsets.ReadOnlyModelVie
     permission_classes = [IsAuthenticated, UserBelongsToOrg, ReadOnly]
     serializer_class = serializers.OrganizationSerializer
 
+    @action(detail=False, methods=["get"], permission_classes=[], authentication_classes=[])
+    def stripe_account_id(self, request):
```
michael-caktus, line 56: New action on OrganziationViewSet, `stripe_account_id`. When a donor visits a donation page, Stripe needs the stripe account id of the org for that donation. We can only depend on having the slug of the revenue program there, so here we create a very small and light endpoint to fetch the stripe_account_id by rev program slug. 

In the future, this could probably be generalized as `payment_provider_account_id`. Not worth it here.

apps/pages/models.py
```diff
@@ -111,6 +117,13 @@ def save(self, *args, **kwargs):
                 raise ValidationError(f"Your organization has reached its limit of {limit.feature_value} pages")
 
         self.slug = normalize_slug(self.name, self.slug)
+
+        if not self.thank_you_redirect:
+            self.thank_you_redirect = self.organization.default_thank_you_redirect
+
+        if not self.post_thank_you_redirect:
+            self.post_thank_you_redirect = self.organization.default_post_thank_you_redirect
```
michael-caktus, line 125: Pages inherit default thank yous from org, if not defined on self.

spa/src/ajax/axios.js
```diff
@@ -6,7 +6,7 @@ export const apiBaseUrl = `/api/${apiVersion}/`;
 
 const Axios = axios.create({
   baseURL: apiBaseUrl,
-  timeout: 5000,
+  timeout: 7000,
```
michael-caktus, line 9: Sometimes creating a payment takes a bit-- it has to call BadActor API and then call Stripe api.

spa/src/components/MainLayout.js
```diff
@@ -1,39 +1,45 @@
...
-          </ProtectedRoute>
-
-          <Route path="/:revProgramSlug/:pageSlug">
-            <DonationPageRouter live />
-          </Route>
-          <Route path="/:revProgramSlug">
-            <DonationPageRouter live />
-          </Route>
-        </Switch>
+        <React.Suspense fallback={<GlobalLoading />}>
```
michael-caktus, line 23: Here, we've converted most of these routes to lazy-loading routes. The javascript bundles needed to render these routes have been split out, and will only be sent to the user if they navigate to one of these urls.  `fallback` is the component that is displayed while that bundle is being fetched.

We also utilize Sentry.ErrorBoundary to catch failed bundles (or any other uncaught errors). That will display a default error page and send us a sentry error.

spa/src/components/TEMP/TemporaryStripeCheckoutTest.js
```diff
@@ -1,29 +1,46 @@
...
-  );
+  const [stripe, setStripe] = useState();
+  const params = useParams();
+
+  const setOrgStripeAccountId = useCallback(async () => {
+    const { data } = await axios.get(ORG_STRIPE_ACCOUNT_ID, {
+      params: { revenue_program_slug: params.revProgramSlug }
+    });
+    const stripeAccount = data.stripe_account_id;
+    setStripe(loadStripe(process.env.REACT_APP_HUB_STRIPE_API_PUB_KEY, { stripeAccount }));
```
michael-caktus, line 32: This is what should fix staging. We can't just use the logged in user anymore-- that only worked by coincidence up until now. So now we also have to "loadStripe" asynchronously, which adds some fiddly state-management-- may be unavoidable.

spa/src/components/TEMP/TemporaryStripeCheckoutTest.js
```diff
@@ -43,6 +60,8 @@ function TemporaryStripeCheckoutTest() {
 export default TemporaryStripeCheckoutTest;
 
 function CheckoutForm() {
+  const { page } = { page: { thank_you_redirect: '', post_thank_you_redirect: 'https://www.caktusgroup.com' } }; // usePageContext()
```
michael-caktus, line 63: When "page" is a full object of page properties, we'll pull "page" from context

spa/cypress/integration/donation-page.spec.js
```diff
@@ -44,4 +46,23 @@ describe('Donation page', () => {
       cy.getByTestId('live-donation-page').should('exist');
     });
   });
+
+  describe('Thank you page', () => {
```
benjaminEwhite, line 50: Should there be a test for what happens when a thank you page URL is blank?

spa/cypress/support/index.js
```diff
@@ -18,3 +18,8 @@ import './commands';
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+Cypress.on('uncaught:exception', (err, runnable) => {
```
benjaminEwhite, line 21: Why is this desirable behavior?

apps/organizations/models.py
```diff
@@ -76,6 +76,9 @@ class Organization(IndexedTimeStampedModel):
         help_text='A fully verified Stripe Connected account should have "charges_enabled: true" in Stripe',
     )
 
+    default_thank_you_redirect = models.URLField(blank=True)
+    default_post_thank_you_redirect = models.URLField(blank=True)
```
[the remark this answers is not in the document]
benjaminEwhite: NVM -- i see that the SPA knows what to do (spa/src/components/donationPage/live/thankYou/GenericThankYou.js)

spa/src/components/donationPage/live/thankYou/GenericThankYou.js
```diff
@@ -0,0 +1,25 @@
...
+
+function GenericThankYou() {
+  const { state: routedState } = useLocation();
+
+  return (
+    <SegregatedStyles>
+      <S.GenericThankYou data-testid="generic-thank-you">
+        <S.ThankYou>
+          <h1>Thank You</h1>
+          <h4>for your donation.</h4>
```
benjaminEwhite, line 15: Are you splitting into distinct h1/h4 here to achieve styling ends? NBD, but this seems to go slightly vs. being semantic with markup.

benjaminEwhite (commented):
A few small comments, but overall looks good.

PR #29, opened 2021-05-26
[NRH-68] Rename + refactor for subscriptions

michael-caktus:
PaymentIntent on our side is now PaymentManager. This is better in a number of ways:
- Disambiguates stripe.PaymentIntent with this class
- Better describes what it does

No longer trying to use a single view for both recurring and one-time payments.

Frontend beings to implement out subscriptions flow on frontend.


apps/contributions/admin.py
```diff
@@ -136,7 +137,7 @@ def _process_flagged_payment(self, request, queryset, reject=False):
         failed = []
         for contribution in queryset:
             try:
-                payment_intent = contribution.get_payment_intent_instance()
+                payment_intent = contribution.get_payment_manager_instance(StripeOneTimePaymentSerializer)
```
benjaminEwhite, line 140: As long as you're preparing the way for supporting distinct one-time and recurring payments, would it make sense to go ahead and create something like a `get_stripe_payment_serializer` function that only returns one-time for now?

apps/contributions/payment_managers.py
```diff
@@ -173,17 +170,10 @@ def create_one_time_payment_intent(self):
...
+    def create_subscription(self):
+        self.ensure_validated_data()
         self.ensure_bad_actor_score()
-        if self.validated_data["payment_type"] == StripePaymentIntentSerializer.PAYMENT_TYPE_SINGLE[0]:
-            stripe_payment_intent = self.create_one_time_payment_intent()
-        else:  # pragma: no cover
-            stripe_payment_intent = self.create_recurring_payment_intent()
-
-        return stripe_payment_intent
+        pass
```
benjaminEwhite, line 176: Is this passing because subscription creation is going to happen NRH-70?

apps/contributions/serializers.py
```diff
@@ -15,7 +15,8 @@ class Meta:
         fields = "__all__"
 
 
-class StripePaymentIntentSerializer(serializers.Serializer):
+class AbstractPaymentSerializer(serializers.Serializer):
+    # These are the fields required for the BadActor API
```
benjaminEwhite, line 19: Out of scope for this PR, but based on our TT convo about edge cases, there's possibly an edge case here where a bad actor could do fraudulent payments with an email of a legit contributor, and cause that email address to get a bad actor score. Seems like a very remote possibility, as there shouldn't be any public knowledge that we use this API. That said, it could happen as an unintended side effect of a bad actor spamming the system with CC and email addresses.

apps/contributions/tasks.py
```diff
@@ -42,7 +43,7 @@ def auto_accept_flagged_contributions():
         payment_state=Contribution.FLAGGED[0], flagged_date__lte=now - settings.FLAGGED_PAYMENT_AUTO_ACCEPT_DELTA
     )
     for contribution in eligible_flagged_contributions:
-        payment_intent = contribution.get_payment_intent_instance()
+        payment_intent = contribution.get_payment_manager_instance(StripeOneTimePaymentSerializer)
```
benjaminEwhite, line 46: Ditto about wondering if it's right time to abstract out retrieval of payment serializer.

apps/contributions/tests/test_payment_managers.py
```diff
@@ -56,97 +56,99 @@ def setUp(self):
...
             "referer": faker.url(),
         }
         self.contribution = ContributionFactory()
         self.contribution.organization = self.organization
         self.contribution.save()
 
-    def _instantiate_pi_with_data(self, data=None):
-        return StripePaymentIntent(data=data if data else self.data)
+    def _instantiate_payment_manager_with_data(self, data=None):
+        return StripePaymentManager(StripeOneTimePaymentSerializer, data=data if data else self.data)
```
benjaminEwhite, line 67: And here. By the way, I don't have an opinion either way, just asking.

apps/contributions/tests/test_views.py
```diff
@@ -62,7 +62,7 @@ def _create_request(self, email=None, rev_slug=None, page_slug=None):
         return request
 
     def _post_valid_payment_intent(self, *args, **kwargs):
-        return stripe_payment_intent(self._create_request(*args, **kwargs))
+        return stripe_one_time_payment(self._create_request(*args, **kwargs))
```
benjaminEwhite, line 65: Maybe rename this method in line 64 to reflect change in function name?

benjaminEwhite (commented):
Looks good to me. Left a few comments, but nothing that needs to block merging this.

apps/contributions/serializers.py
michael-caktus: Good point. I have to assume that the BadActor API is responsible for how tightly its judgement is tied to a specific email address. We send a whole mess of info over, so I think it has the ability to make a smarter determination.

At any rate, yeah-- not authenticating the email address provided means that this will always be an issue. It's worth mentioning that getting a BadActor score only means that the contribution is flagged for further inspection. The rightful owner of the email address would only notice (if anything) a delay between payment and capture of funds.

PR #30, opened 2021-05-27
[NRH-70] Stripe Customers **WIP

michael-caktus:
WIP

Each Organization will want to have a list of its own Customers.  The big challenge here is that RevEngine only wants one contributor per unique email, whereas Stripe has a Customer per organization.  It would be easy enough if the Customer was essentially metadata, but in fact the Customer stores a default payment method that is required to capture each payment on a recurring donation.

### DIFF
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -1,6 +1,9 @@
 from django.db import models
 
+import stripe
+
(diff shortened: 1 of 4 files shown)


michael-caktus:
Closing this. Went a different direction and implemented on NRH-37

PR #34, opened 2021-06-03
[NRH-37] Recurring payments

michael-caktus:
Implements Recurring payments with Stripe.

We've opted to use Stripe's Subscriptions feature for this. As a brief aside, there is a scenario where we use the PaymentIntent API, set off_session=True, store donor credentials using PaymentMethods, and handle the payment scheduling ourselves. While this would have some benefits, the complexity and room for error recommend against it.

Stripe Subscriptions require a number of related models. With PaymentIntents (one-time payments), we have no need to store any references to payment information. With Subscriptions, we must. So the flow for recurring payments looks something like this:
1. Front end creates a Stripe PaymentMethod using the card info provided, sends its ID to the back end to get stored. 
2. On the backend, Stripe "Customer" is created for each transaction.
3. Contribution object is created on our side, storing necessary related Stripe models
4. Checks with BadActor, flagging if necessary.
5. If flagged, Subscription is not created, but info necessary to create it later is stored on Contribution.
6. Completing a recurring payment, either because it wasn't flagged, or because it was accepted, means finally creating the stripe Subscription instance.
7. This model needs the customer, the payment method, a Stripe Price, a Stripe Product.
8. The Stripe Price is created inline with the Subscription.
9. The Stripe Product is, for now, created when the org adds stripe as their payment provider. One product, "RevEngine Donation", created per Organization.
10. Contributions created in this way are "processing" until Stripe sends a payment_intent.succeeded/failed. Internally, Stripe creates a PaymentIntent for every attempted payment of a recurring donation. In our case, that means that as soon as the Subscription is created, Stripe will attempt its first PaymentIntent.
11. Subsequent attempts to capture a recurring payment will follow this flow. In response to payment_intent events, recurring Contributions will update their fields accordingly. If `failed`, Contribution state will indicate that, and the Event JSON will be stored on the model of inspection.

It's worth noting that it is very difficult to test the asynchronous aspects of contributions. The interval for the contribution is shortened based on the `USE_DEBUG_INTERVALS` setting, but we can only shorten it a little bit. The shortest interval Stripe will accept for a subscription is `daily`.  See docs [here](https://stripe.com/docs/api/prices/object#price_object-recurring-interval).  It would be really great to find a way to test this a little better than waiting a day...


apps/contributions/payment_managers.py
```diff
@@ -151,59 +169,169 @@ def complete_payment(self, **kwargs):  # pragma: no cover
...
     def create_subscription(self):
+        """
+        Unlike PaymentIntents, Stripe Subscriptions do not have a concept of a "capture_method".
+        So far, it seems like the best way to delay the "capture" of a flagged contribution is to
+        simply not create it until it's been approved.
+        """
         self.ensure_validated_data()
         self.ensure_bad_actor_score()
-        pass
+        org = self.get_organization()
```
JeremyGibson, line 214: organization

apps/contributions/serializers.py
```diff
@@ -25,17 +25,21 @@ class AbstractPaymentSerializer(serializers.Serializer):
...
+    INTERVAL_CHOICES = [
+        Contribution.INTERVAL_ONE_TIME,
+        Contribution.INTERVAL_MONTHLY,
+        Contribution.INTERVAL_YEARLY,
+    ]
+    interval = serializers.ChoiceField(choices=INTERVAL_CHOICES, default=Contribution.INTERVAL_ONE_TIME[0])
+
+    @classmethod
+    def convert_cents_to_amount(self, cents):
+        return str(float(cents / 100))
```
JeremyGibson, line 42: f"{(cents/100):.2f}"

apps/contributions/views.py
```diff
@@ -40,12 +40,20 @@ def stripe_one_time_payment(request):
...
+            stripe_payment_intent = stripe_payment.create_one_time_payment()
+            response_body = {"detail": "Success", "clientSecret": stripe_payment_intent["client_secret"]}
+        else:
+            # Create subscription with Stripe, associated local models
+            stripe_payment.create_subscription()
+            response_body = {"detail": "Success"}
     except PaymentBadParamsError:
         return Response({"detail": "There was an error processing your payment."}, status=status.HTTP_400_BAD_REQUEST)
+    except PaymentProviderError as pp_error:
+        return Response({"detail": str(pp_error)}, status=status.HTTP_400_BAD_REQUEST)
```
JeremyGibson, line 54: logging?

PR #40, opened 2021-06-14
[NRH-111] Calculated fee

michael-caktus:
Subtask of NRH-67.

Donors have the option to pay stripe fees. That fee is calculated based on the amount selected. Fortunately, Stripe recently flattened their rate for all payment types (card, apple, google). This significantly decreases complexity here.  So as it is we just need to calculate the fee based on a flat fee of $0.30 and a flat rate of 0.29% of the total.


x110dc:
Unfortunately it's not that simple for nonprofits. For nonprofits Stripe charges 2.2% + 0.30 for all cards except Amex. For Amex they charge 3.5% (no additional flat fee).

x110dc:
My guess is that you'll have difficulty determining the type of card for Google, Apple and Microsoft Pay transactions.

michael-caktus:
@x110dc 
Of course! I didn't think about the non-profit angle.

Aside from the technical difficulty of determining the card type, I think there are some UX pains here too-- the donation amount widget could potentially be above the fold when the user is entering their card info. The change in the fee would be invisible, aside from the text in the "Pay" button changing.

I think the best way to go with this for now is what we briefly discussed before. We should choose a rate to target and change the language on the toggle to make it clear (for legal purposes?). Is that acceptable?  If so, any thoughts on the right rate to choose? And the right language to use?

spa/cypress/integration/donation-page.spec.js
```diff
@@ -96,6 +96,26 @@ describe('Donation page', () => {
...
+          cy.getByTestId('pay-fees').find('label').contains(calculatedFee);
+        });
+      });
+    });
+  });
+
+  describe('Resulting request', () => {
+    it('should send a request with the expected amount', () => {
+      //
+    });
```
benjaminEwhite, line 118: Remove this?

spa/cypress/integration/donation-page.spec.js
```diff
@@ -96,6 +96,26 @@ describe('Donation page', () => {
         }
       });
     });
+
+    it('should render the correct fee base on frequency and amount', () => {
```
benjaminEwhite, line 100: based, not base

benjaminEwhite (approved):
Minor cleanup items noted, but otherwise :+1:

PR #41, opened 2021-06-14
[NRH-107] Styled payment widget

michael-caktus:
Subtask of NRH-67.

The payment widget needed a good deal of hooking up as well. This does that, plus styles the thing.


spa/src/utilities/scrollIfNotVisible.js
```diff
@@ -0,0 +1,17 @@
+function isInViewport(el) {
+  var bounding = el.getBoundingClientRect();
```
benjaminEwhite, line 2: use `const` instead?

benjaminEwhite (approved):
LGTM

PR #44, opened 2021-06-15
[NRH-121] Edit interface, element removal

michael-caktus:
Now you can remove an element from the list of page elements by selecting it and clicking the trash icon in the top right. You cannot remove certain elements, as they are required for the page to function.  Also adds a "disabled" state to CircleButton and disables the "save" button until changes have been made.


spa/src/elements/modal/GlobalConfirmationModal.js
```diff
@@ -35,11 +35,11 @@ function GlobalConfirmationModal({ message, onConfirm, onDecline, isOpen, closeM
...
-  message: PropTypes.string.isRequired,
-  onConfirm: PropTypes.func.isRequired,
+  message: PropTypes.string,
+  onConfirm: PropTypes.func,
   onDecline: PropTypes.func,
-  isOpen: PropTypes.bool.isRequired,
-  closeModal: PropTypes.func.isRequired
+  isOpen: PropTypes.bool,
+  closeModal: PropTypes.func
 };
```
benjaminEwhite, line 43: Why are these becoming non-required?

benjaminEwhite (approved):
LGTM. 

Question: If a user deletes a non-required element, is there an easy way to bring it back?

spa/src/elements/modal/GlobalConfirmationModal.js
michael-caktus: Because of the way the modal is being passed this context, there are cases where the modal is "instantiated" before it is displayed, leading to errors here.

michael-caktus:
Currently, no way to bring back deleted page elements. Hopefully risk is mitigated by the steps required-- remove it from the edit interface, view the page and click "save".

On the subject, a good place for improvement is "caching" user input for these fields in sessionStorate/localStorage. That wouldn't solve the above issue, but it would improve the user experience nicely.

PR #45, opened 2021-06-15
[NRH-120] Page-edit validation

michael-caktus:
Enables page validation and enables required-element validation. Any validation beyond that is possible, but it's not clear at the moment what that would need to be.

It's worth noting that I vacillated a bit on whether or not to handle the validation on the backend with a PageElementsSerializer. I think there's lots of value in that (eg I particularly don't like front-end validation), but if we do that, fully controlled page elements start to recommend themselves. Rather than Page.elements being a flexible JSON field, Page.elements would be a one-to-many to an Element model. That's fine and all, but it's probably a bit more control than we need/want for this version. 

The way we have it now, adding more features to the donation page only requires front-end changes, and that's a big plus.

### DIFF
--- a/spa/src/components/pageEditor/validatePage.js
+++ b/spa/src/components/pageEditor/validatePage.js
@@ -1,0 +1,24 @@
+import isEmpty from 'lodash.isempty';
+import * as dynamicElements from 'components/donationPage/pageContent/dynamicElements';
+
+function validatePage(page) {
+  const errors = {
+    ...validateRequiredElements(page.elements)
+  };
+  return !isEmpty(errors) && errors;
+}
+
+function validateRequiredElements(elements = []) {
+  const requiredElements = Object.keys(dynamicElements).filter((key) => dynamicElements[key].required);
+  const errors = {};
+  for (let i = 0; i < requiredElements.length; i++) {
+    const requiredElement = requiredElements[i];
+    if (!elements.find((el) => el.type === requiredElement)) {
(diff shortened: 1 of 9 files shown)


spa/src/components/pageEditor/validatePage.js
```diff
@@ -0,0 +1,24 @@
...
+  const errors = {
+    ...validateRequiredElements(page.elements)
+  };
+  return !isEmpty(errors) && errors;
+}
+
+function validateRequiredElements(elements = []) {
+  const requiredElements = Object.keys(dynamicElements).filter((key) => dynamicElements[key].required);
+  const errors = {};
+  for (let i = 0; i < requiredElements.length; i++) {
```
benjaminEwhite, line 14: Any reason not to use `for... of`?

spa/src/elements/alert/Alert.styled.js
```diff
@@ -39,3 +39,8 @@ export const Close = styled.button`
   right: 0;
   top: 0;
 `;
+
+export const ErrorsList = styled.ul`
```
benjaminEwhite, line 43: Remove?

benjaminEwhite (approved):

spa/src/components/pageEditor/validatePage.js
michael-caktus: Just a personal preference.

PR #46, opened 2021-06-15
NRH-124: Add endpoint for listing org's contributions

benjaminEwhite:


apps/contributions/views.py
```diff
@@ -149,3 +152,12 @@ def process_stripe_webhook_view(request):
         logger.error("Could not find contribution matching provider_payment_id")
 
     return Response(status=status.HTTP_200_OK)
+
+
+class ContributionsListView(generics.ListAPIView):
```
michael-caktus, line 157: Is the idea here that since we only need the `list` action for this ticket, we use `ListAPIView`? We'll eventually need `retrieve` as well. But since we don't need it yet this seems fine to me. `ReadOnlyModelViewSet` contains only `retrieve` and `list`.

apps/contributions/views.py
```diff
@@ -149,3 +152,12 @@ def process_stripe_webhook_view(request):
...
+    serializer_class = serializers.ContributionSerializer
+    model = Contribution
+    permission_classes = [IsAuthenticated, UserBelongsToOrg]
+
+    def get_queryset(self):
+        return self.model.objects.filter(organization__users=self.request.user)
```
michael-caktus, line 163: Any reason not to use `OrganizationLimitedListView`?

michael-caktus (approved):

michael-caktus:
Approved with questions/comments

apps/contributions/views.py
benjaminEwhite: It expects there to be the view instance to have a  `.action`, but since I'm using a generic CBR (in order to get default pagination behaviors without having to do any implementation), there `self` has no `action`. Suppose I could add a further condition to `OrganizationLimitedListView` to check that it has an action attribute.

apps/contributions/views.py
benjaminEwhite: That's my lack of experience with DRF more than anything else. IIUC, `ReadOnlyModelViewSet` would support listing and retrieving a single item? In that case, I can see swapping.

PR #58, opened 2021-06-24
[NRH-141] Handle no elements json

michael-caktus:
Checks for elements before trying to map on frontend, just in case. Also adds default of list to json field.


spa/src/components/donationPage/DonationPage.js
```diff
@@ -39,7 +39,7 @@ function DonationPage({ page, live = false }) {
                 {getters.getGraphicElement()}
                 <form ref={formRef} data-testid="donation-page-form">
                   <S.PageElements>
-                    {page?.elements.map((element) => getters.getDynamicElement(element, live))}
+                    {page?.elements?.map((element) => getters.getDynamicElement(element, live))}
```
benjaminEwhite, line 42: Seems like `.elements` should always be an array, but if for some reason `page.elements` is undefined, wouldn't this render as just the string "undefined" (note: i'm not actually certain this is how react handles passing `{undefined}` in this context.

benjaminEwhite (approved):

spa/src/components/donationPage/DonationPage.js
michael-caktus: We're just covering both bases here. Elements should always be an array by default, even for new pages, but it's possible that somebody in django admin could go in and null-out that value, which would cause this to crash. In React, `null` and `undefined` simple don't render anything.

PR #62, opened 2021-06-30
[NRH-151] Magic link auth with poc

michael-caktus:
Subtask of NRH-91. This implements everything but:
* Rate throttling
* Token invalidation after attempt to redeem
* Actual templated email sent
* Tests


spa/cypress/integration/contributor.spec.js
```diff
@@ -0,0 +1,35 @@
...
+      });
+    });
+
+    it('should display server generated error on email if non-200 status', () => {
+      const emailError = ['email error'];
+      cy.intercept(
+        { method: 'POST', url: getEndpoint(GET_MAGIC_LINK) },
+        { body: { email: emailError }, statusCode: 400 }
+      ).as('getMagicLink');
+      cy.getByTestId('magic-link-email-button').click();
```
benjaminEwhite, line 27: Is there a missing assertion here? Doesn't seem like this proves that server generated error is displayed.

spa/src/components/contributor/ContributorEntry.js
```diff
@@ -0,0 +1,71 @@
...
+  const [showConfirmation, setShowConfirmation] = useState(false);
+
+  const handleSendMagicLink = async (e) => {
+    e.preventDefault();
+    setLoading(true);
+    try {
+      const response = await axios.post(GET_MAGIC_LINK, { email });
+      if (response.status === 200) setShowConfirmation(true);
+    } catch (e) {
+      if (e?.response?.data) {
```
benjaminEwhite, line 28: Is there a scenario where error doesn't have response.data? If so, doesn't look like the user would know there's a problem.

benjaminEwhite (commented):
Overall looking good, but a few comments.

apps/contributions/models.py
```diff
@@ -14,6 +17,16 @@ def contributions_count(self):
     def most_recent_contribution(self):
         return self.contribution_set.filter(status="paid").latest()
 
+    @property
+    def is_authenticated(self):
+        """
+        Copy django.contrib.auth.models import AbstractBaseUser for request.user.is_authenticated
+
+        Always return True. This is a way to tell if the user has been
```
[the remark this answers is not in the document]
michael-caktus: Built in, called internally. I agree it sounds scary! It think it's used to differentiate between users that have an auth mechanism (based off of AbstractBaseUser) and those that don't.

PR #63, opened 2021-06-30
Nrh 42 nrh 80  one time donor recieves thank you

JeremyGibson:
This PR adds the API for sending emails using template values that respond to donation actions from DonationPages.

For this PR the only implemented donation action is a One Time Donation.

This PR does not set up the ability to create a template on the ESP (Mailgun) so those templates will need to be created by NRH staff.  If the template values that are required for those templates differ from what we have here

The type and schema of the defaults can easily be modified to handled different types. At the moment the definition is static and would require a code change to add a new ContactType.

As the first commit notes this PR switches from using SMTP with mailgun to the API.

There is a management command to create the templates and link them to the donation pages `create_default_templates`

There is some sugar in here as this PR adds a JSON widget override for the Admin making it much easier for admins to view and modify JSONFields.

- [x] Add MAILGUN_API Settings to Heroku before merge.


apps/emails/models.py
```diff
@@ -0,0 +1,109 @@
...
+
+class BaseEmailTemplate(models.Model):
+    class ContactType(models.TextChoices):
+        ONE_TIME_DONATION = "OTD", "One Time Donation"
+        RECURRING_DONATION = "RCD", "Recurring Donation"
+        FAILED_PAYMENT = "FLD", "Failed Payment"
+        CANCELLED_DONATION = "CAN", "Cancelled Donation"
+
+        @staticmethod
+        def default_schema():
```
michael-caktus, line 34: Just curious, why not @property? I think there are some subtleties here that I don't fully understand.

michael-caktus: ha. It thinks I'm trying to tag somebody named Property.

JeremyGibson:
It has to be a callable for it to work as a default.

michael-caktus (approved):

PR #68, opened 2021-07-02
[NRH-158] Hub slack integration

michael-caktus:
Adds a new app, `slack`, to manage the integration.
So we've got a nice, agnostic set of models and classes to handle Slack integration.
- `SlackIntegration` is an abstract model
- `HubSlackIntegration` is a singleton model, subclassing from `SlackIntegration`, for managing Hub tokens and values
- `OrganizationSlackIntegration` subclasses `SlackIntegration` and it tracks Organization Slack integrations

As a matter of fact, this PR enables Organization slack integrations, with a few caveats:
- Hub must manually enter the organization's Slack Bot Token and target slack channel in to a new `OrganizationSlackIntegration` instance and attached it to the org.
- Acquiring this Bot Token will likely be a bit too technical for orgs, so will likely require help from Hub


apps/contributions/admin.py
```diff
@@ -91,15 +91,15 @@ class ContributionAdmin(admin.ModelAdmin):
...
-        "updated",
+        "modified",
         "created",
     )
 
     search_fields = (
         "organization__name",
         "contributor__email",
         "donation_page__name",
-        "updated",
```
JeremyGibson, line 102: Is this relevant to this PR?

apps/slack/models.py
```diff
@@ -0,0 +1,54 @@
...
+        super().save(*args, **kwargs)
+
+    class Meta:
+        abstract = True
+
+
+class HubSlackIntegration(SlackIntegration, SingletonModel):
+    org_channel_prefix = models.CharField(
+        max_length=24,
+        default="donations-",
```
JeremyGibson, line 33: Any reason the default shouldn't start with "#"

apps/slack/models.py
```diff
@@ -0,0 +1,54 @@
...
+    return val
+
+
+class SlackNotificationTypes(models.TextChoices):
+    SUCCESS = "success", "Success"
+
+
+class SlackIntegration(models.Model):
+    bot_token = models.CharField(max_length=255)
+    channel = models.CharField(max_length=255, help_text="All donation notifications will be sent to this channel")
```
JeremyGibson, line 20: Why not add a default here as well?

JeremyGibson (changes requested):
Just a few questions.

apps/contributions/admin.py
michael-caktus: Not directly-- contribution admin was crashing because "updated" doesn't exist.

apps/slack/models.py
```diff
@@ -0,0 +1,54 @@
+from django.db import models
+
+from solo.models import SingletonModel
+
+from apps.organizations.models import Organization
+
+
+def format_channel_name(val):
+    if val and val[0] != "#":
```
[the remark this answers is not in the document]
michael-caktus: I'm operating under the understanding that if it gets to save, it's already passed serializer validation, which is where we should decide whether or not to permit empty strings.

JeremyGibson (approved):

PR #76, opened 2021-07-12
[NRH-138] PayFees changes

michael-caktus:
#### What's this PR do?
- Move's the "Pay transaction fees" button down into the payment widget.
- Fixes the calculated fee to include fee increase from original payment + fee

- Visit a live donation page
- Make sure that donation page is set up to display the "Pay transaction fees" button
- Verify that the pay transaction fees toggle is grouped with the payment widget
- Select an amount to donate and verify that the correct fee is being applied
- Complete the contribution and verify that the correct amount is recorded on the Contribution object in Django-admin

No


JeremyGibson (commented):
Cypress mp4 and image files should be ignored.

JeremyGibson:
This was not working for me before the last commit I made. When I attempted to toggle the agreePayFees, nothing would happen.
This works, but might not be Reacty.

JeremyGibson (changes requested):
Looks good. I made a commit that, for me, allowed the toggle to work. The cypress artifacts need to be removed. Then good to shipit.

JeremyGibson (approved):
:shipit:

PR #80, opened 2021-07-14
[NRH-162] (Hopefully) Fix ApplePay

michael-caktus:
#### What's this PR do?
When Organizations are first fully verified with Stripe (ie Stripe account exists, is connected to hub, and is ready to accept payments), we now verify our domain with Stripe for that account via stripe API (`stripe.ApplePayDomain.create`).  There are a ton of caveats to this, the details of which are described in [this basecamp post](https://3.basecamp.com/3120230/buckets/21497549/messages/3956684749).

See basecamp post above for testing flow.


JeremyGibson (approved):
I made one change to the wording of a doc string otherwise looks good.

JeremyGibson:
I approved but I missed that the tests were failing on cypress.

PR #85, opened 2021-07-16
[NRH-203-205] bugfix New styles bugs

michael-caktus:
#### What's this PR do?
Fixes two bugs:
1. Newly created styles wouldn't appear in the list of styles after creation
2. Could save newly created styles

Try to do both those things

Nope


JeremyGibson (approved):
:shipit:

PR #88, opened 2021-07-16
NRH-201: fix filter pagination bug

benjaminEwhite:
#### What's this PR do?

Fixes a bug. Previously, if you were on (say) the second page of donation results, and you activated a filter, if the resulting filtered results only had a single page, you would get an error.

This PR fixes the bug by resetting to the first page of results when a filter changes.

Go to the second page of donations in donations dashboard list view. Activate a filter. There should be no error, and should get the first page of (filtered) results.

No


michael-caktus (approved):
Much better than my previous attempts. Nice to find that `useControlledState` prop.

PR #89, opened 2021-07-16
[NRH-196] logs error and skips slack message if no integration

michael-caktus:
#### What's this PR do?
Logs at info level if no Slack integration exists for target, instead of crashing

Make a contribution for an Org without slack integrated. Observe a lack of explosions. Possible view logs for .info level message.

#### Before it can be merged
- [x] Need to get unit test coverage up just a skosh


JeremyGibson (approved):

JeremyGibson:
The test coverage here was dropped by a tenth of a percent. This is trivial, but I went through the report and modified coverage to exclude management commands, since these are not application critical resources. That moved us up.  The functionality in this PR is well covered by included tests.

PR #94, opened 2021-07-20
refs #NRH-220 Allows fractional other amounts.

JeremyGibson:
#### What's this PR do?
Modifies a regression where fractional OtherAmounts were disabled.

1. Go to a donation page with other amount enabled
2. Add a fractional donation. (e.g 42.42)
3. Note that the donation button reflects the fractional change

No.


benjaminEwhite (approved):
Could consider adding an integration test that submits a donation, and intercept the donation data to show it's in expected state. 

LGTM

JeremyGibson:
Since the error in the fractional amount was happening before submission, I believe the existing integration test that checks the donation button for the correct calculated amount is sufficient as a test. However I am adding an "other" with fractional amounts to the existing test fixtures. They weren't there before. That should guard against this regression again.

PR #102, opened 2021-07-29
refs #NRH-250 Refactors metadata generation

JeremyGibson:
#### What's this PR do?
Refactors the app to use first_name, last_name, and the expected `mailing` prefixes 
Adds some logic to make sure that the correct metadata is present on the Processor Object.

1. Make a one-time contribution. 

- Any metadata with a ProcessorObject of ALL or PAYMENT should be found on the PaymentIntent Object in Stripe.

2. Make a recurring donation.

- Any metadata with a ProcessorObject of ALL or CUSTOMER should be found on the Customer Object.
- Any metadata with a ProcessorObject of ALL or PAYMENT should be found on the Subscription Object

No other extraneous metadata should be found on any objects.

*NOTE* Customer creation for one-time payments still has not been implemented, so metadata will not be on those objects for one-time payments.

I have a table dump of my local ContributionMetadata that I will add to `rev-engine-test`. This should give us a common reference to understand future requests or changes for this feature.

For BAD_ACTOR this PR changes the keys sent from `given_name` to `first_name` and `family_name` to `last_name`


spa/src/components/donationPage/pageContent/DDonorInfo.js
```diff
@@ -28,21 +28,21 @@ function DDonorInfo(props) {
             <Grid.Column tablet={16} computer={8}>
               <Input
                 type="text"
-                name="given_name"
+                name="first_name"
                 label="First Name"
                 value={firstName}
                 onChange={(e) => setFirstName(e.target.value)}
-                errors={errors.given_name}
+                errors={errors.firstName}
```
michael-caktus, line 35: This and all of the other errors props need to point to the serializer key for the value, not the frontend state key-- that is, here we should have:
```jsx
errors={errors.first_name}
```
rather than
```jsx
errors={errors.firstName}
```

michael-caktus (changes requested):

michael-caktus (approved):

PR #104, opened 2021-08-02
[NRH] bugfix 503 on Page Save

michael-caktus:
#### What's this PR do?
Ensures that page screenshots are properly blobby before sending. Use (slightly) more modern API, `element.toBlob`, to ensure proper encoding and async timing. This may be the culprit for the 503 on staging. See [this Heroku article ](https://help.heroku.com/N5O6E5XP/debugging-h18-server-request-interrupted-errors-in-python-applications)for details on the particular error we were getting.

If this works, you should be able to save donation page edits without getting a 503.

In order to capture page screenshots, the env var `REACT_APP_CAPTURE_PAGE_SCREENSHOT` will need to be set to `'true'`.  That said, this feature is currently stuck "on", since we aren't able to get env vars set on the front end properly. NRH-76 exists to fix this.


michael-caktus:
This ought to be tested on staging, so I'm going to push it up.

michael-caktus:
This did NOT fix the 503 issue. For now, we're going to disable the screenshot feature so we can make progress elsewhere. This code is still good as an improvement. Will add a commit to set CAPTURE_PAGE_SCREENSHOT to env var, thereby disabling it.

PR #113, opened 2021-08-03
[NRH-260] handle string amount

michael-caktus:
#### What's this PR do?
One of those "typescript would have caught this, but is it really worth it?" times. The "amount" argument in calculateStripeFees can be an int or a string. The string wasn't being handled properly. 

Stripe fees should be calculated accurately now.

no


benjaminEwhite:
LGTM. :shipit:

PR #114, opened 2021-08-05
[NRH-236-237-238] Amount and Frequency queryparams and defaults

michael-caktus:
#### What's this PR do?
Tickets 236, 237 and 238 are implemented on this PR. We've added the ability to select default selected amounts per frequency, default selected frequency. We're also responding to the `amount` and `frequency` query parameters as outlined in the requirements.

1. Edit a donation page.
2. Edit the Frequency element
3. Observe that you can check one or zero "default" checkboxes.
4. Edit the Amount element
5. Observe that you can click on an amount (not the X) and it will highlight, indicating that it is the default for that frequency.
6. Save your changes and visit the page live.
7. Observe that the defaults, as set above, are selected (for each frequency).
8. Next, add `amount=99` and/or `frequency=monthly` or some combination of some amount and frequency, and observe that the behavior is as expected.

No

### DIFF
--- a/spa/src/components/pageEditor/elementEditors/amount/AmountEditor.js
+++ b/spa/src/components/pageEditor/elementEditors/amount/AmountEditor.js
@@ -73,7 +73,10 @@ function AmountEditor() {
 
   return (
     <S.AmountEditor data-testid="amount-editor">
-      <S.HelpText>Highlighted amounts will be selected by default</S.HelpText>
+      <S.HelpTexts>
+        <S.HelpText>Click an amount to set a default value</S.HelpText>
+        <S.HelpText>Highlighted amounts will be selected by default on live donation pages</S.HelpText>
+      </S.HelpTexts>
       <S.FrequenciesList>
         {frequencies ? (
           frequencies.map((freq) => {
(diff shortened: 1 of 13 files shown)


spa/src/components/pageEditor/elementEditors/amount/AmountEditor.js
```diff
@@ -60,52 +61,79 @@ function AmountEditor() {
...
-        <S.ToggleWrapper>
-          <S.Toggle
-            label='Include "other" option'
-            checked={elementContent?.allowOther}
-            onChange={toggleAllowOther}
-            toggle
-          />
-        </S.ToggleWrapper>
-      </S.Toggles>
+      <S.HelpText>Highlighted amounts will be selected by default</S.HelpText>
```
JeremyGibson, line 76: Suggestion add: Click amount bubble to highlight.

JeremyGibson: Suggestion: One default per frequency may be added.

JeremyGibson (approved):
The greater functionality for this works, but when an element is set as default in the editor the preview does not reflect it until save and refresh.  I think we can move this along and open that issue as an improvement.

spa/src/components/pageEditor/elementEditors/amount/AmountEditor.js
michael-caktus: Good call

PR #124, opened 2021-08-10
[NRH-290] Hub API access

michael-caktus:
NOTE: I needed some stuff from 263 and 239, so those are merged in here. Best to merge those first, if you're overwhelmed by the diff.

#### What's this PR do?
We've create a distinct app to handle "public" access to the API. In this case, "public" just means a client other than the web app. We've done this to help ensure that breaking changes aren't accidentally made to any endpoints accessed by third-parties (including NRH here. The endpoint GET `/api/v1/public/revenue-programs/` lists all RevenuePrograms. Note that this list is currently paginated. It's very easy to turn that off if we'd prefer, down the line. Each RevenueProgram entry has a `url` key containing a link to the detail for that revenue program. That detail contains all BenefitLevels and all Benefits for that revenue program.

This endpoint is restricted to superusers.

Note that, because this is pretty simple so far, we haven't exposed any documentation for this public API.

Hopefully just use the Postman Collection to POST to /token with superuser credentials, then hit the endpoint mentioned above.

No.


JeremyGibson (approved):
:shipit:

PR #126, opened 2021-08-11
[NRH-267.270] Elements: Update donor info + Donor address + Amount >= 1

michael-caktus:
- [x] Merge NRH-263 in to develop
- [x] Fix tests for this PR

#### What's this PR do?
Implements NRH-267, NRH-270, NRH-271.
1. The "Donor info" block is now just name and email. It is required, can only have one, and all fields must be filled out to donate.
2. New "Donor address" block contains address info, including zip and country. It is required, can only have one, and all fields must be filled out to donate.
3. Donors will see a message "Please enter an amount of at least $1" if they have entered an "Other" amount less than 1.

Edit a donation page and add Donor info and Donor address blocks.
View that page live and try to submit without filling anything out.
Note that error messages appear beneath the required fields.
Also, try to donate $0.50 using the "Other" field. Observe that a message appears instead of the card payment form.

No


apps/organizations/serializers.py
```diff
@@ -46,15 +46,18 @@ class RevenueProgramSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = RevenueProgram
-        fields = ["id", "slug"]
+        fields = ["id", "name", "slug"]
 
     def get_fields(self):  # pragma: no cover
```
JeremyGibson, line 51: Why aren't we covering this?

apps/organizations/tests/test_org_views.py
```diff
@@ -96,74 +96,63 @@ def test_request_id_when_everythings_fine(self):
...
-
-#     def test_cannot_update_revenue_program_slug(self):
-#         rp = RevenueProgram.objects.all().first()
-#         self.authenticate_user_for_resource(rp)
-#         self.login()
-#         new_slug = "a-new-slug"
-#         response = self.client.patch(f"{self.detail_url}/{rp.pk}/", {"name": new_slug})
-#         self.assertEqual(response.status_code, 200)
-#         self.assertNotEqual(response.data["slug"], new_slug)
+class RevenueProgramViewSetTest(AbstractTestCase):
```
JeremyGibson, line 99: And it returns!

apps/organizations/serializers.py
michael-caktus: Though it looks like I added it, I really just replaced it... it was left from a previous commit. I'm not sure why it wasn't covered originally, though I suspect it was a bit of a cheat to get coverage. Right now, there's some question as to whether it's necessary-- we shouldn't be hitting that serializer during patch or post at all right now. In the future, it might be handy, but for now it's superfluous.

michael-caktus: But it would be easy enough to test, if we wanna keep it. I agree it's odd to have the no-coverage statement there.

JeremyGibson: Ahh. I see it was me. Also on my branch NRH-180 This is what was causing the 500s.

JeremyGibson: I just removed the name and org, and left slug.

JeremyGibson (approved):

PR #136, opened 2021-08-17
refs #NRH-306 Fixes the decimal places to 2

JeremyGibson:
#### What's this PR do?
Ensures that calculated fees will always have 2 decimal places.

1. Enable pay fees on a page.
2. Enter an amount of $180 to contribute
3. Observe that the calculated fee is $5.30 rather than $5.3

No.

### DIFF
--- a/spa/src/components/donationPage/pageContent/DPayment.js
+++ b/spa/src/components/donationPage/pageContent/DPayment.js
@@ -62,7 +62,9 @@ export function PayFeesWidget() {
       <S.Checkbox
         label={
           amount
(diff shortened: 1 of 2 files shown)


michael-caktus:
Did this get pretty rigorously tested? Setting toFixed at this stage potentially has some very javascripty consequences. It's probably fine if it's checked pretty thoroughly, but if not it might be better to just do the .toFixed bit rigvht before it's rendered.

michael-caktus (approved):

PR #144, opened 2021-08-18
[NRH-170] "Must be user instance" Sentry error

michael-caktus:
#### What's this PR do?
We've occasionally received a sentry error, `Cannot query "<some email address here>": Must be "User" instance.`. It doesn't seem to break anything, but as it appears to be related to authentication/permissions, it seemed like a good idea to sort it out. It looks like there was a `get_queryset` method being used to limit results to authorized users, but that was failing to check that the user type was "User" and not "Contributor". This error will have occurred after somebody has logged in to the contributor dashboard via magic link, then attempted to access a resource from the either Django admin, or the org dashboard. The does not appear to be a security issue here, just a non-breaking error.  This PR just adds a check to the `get_queryset` method for usertype.

This bug isn't super easy to reproduce. Some magic combination of authenticating with contributor magic link and then trying to view orgs in django admin or some such...? Add to that the fact that this error does surface to the user and I would recommend against trying to test this one. If we get this sentry error again, we'll know something else is going on.

No

### DIFF
--- a/apps/organizations/views.py
+++ b/apps/organizations/views.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
 
 
 class ReadOnly(permissions.BasePermission):
-    def has_object_permission(self, request, view, obj):
+    def has_object_permission(self, request, *args):
         # Read permissions are allowed to any request,
         # so we'll always allow GET, HEAD or OPTIONS requests.
         if request.method in permissions.SAFE_METHODS:
--- a/apps/organizations/views.py
+++ b/apps/organizations/views.py
@@ -30,10 +30,17 @@ class OrganizationLimitedListView:
     model = None
 
     def get_queryset(self):
+        """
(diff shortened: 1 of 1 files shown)


apps/organizations/views.py
```diff
@@ -30,7 +33,7 @@ def get_queryset(self):
         if isinstance(self, Organization):
             return self.model.objects.filter(pk=self.id)
 
-        if self.action == "list" and hasattr(self.model, "organization"):
+        if isinstance(self.request.user, user_model) and self.action == "list" and hasattr(self.model, "organization"):
```
JeremyGibson, line 36: This long one liner obscures what is going on here, can  you add a docstring to explain. FTF! (For The Future)

JeremyGibson (changes requested):

apps/organizations/views.py
JeremyGibson: Even better! :shipit:

JeremyGibson (approved):

PR #153, opened 2021-08-19
[NRH-262-286] Datetime timezones

michael-caktus:
#### What's this PR do?
We were sending datetimes back to the API without timezones. Two bugs were fixed by sending datetimes set on the frontend back with timezones included:
1. 262-- donation page edit, published date time was coming back 4 hours behind the date you saved.
2. 286 -- from and to filters in the donations list were excluding many donations for the selected day.

1. Add a time to a published_date in the page editor. Save it. Check the published time again and it should not be different from what you saved.
2. Set some from and two filters. No donations from the dates set should be removed from the list.

No

### DIFF
--- a/spa/src/utilities/formatDatetimeForAPI.js
+++ b/spa/src/utilities/formatDatetimeForAPI.js
@@ -10,7 +10,6 @@ export function formatDatetimeRoundedDay(datetime, startOfDay = true) {
   let dt = datetime;
   if (startOfDay) dt = setTimeToMidnight(dt);
   else dt = setTimeToJustBeforeMidnight(dt);
-  console.log('that dt', format(dt, "y-MM-dd'T'HH:mm:ss.SSSx"));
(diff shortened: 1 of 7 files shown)


spa/src/utilities/formatDatetimeForAPI.js
```diff
@@ -1,7 +1,30 @@
-import { format } from 'date-fns';
+import { format, setHours, setMinutes, setSeconds, setMilliseconds } from 'date-fns';
 function formatDatetimeForAPI(datetime, withTime = true) {
```
JeremyGibson, line 2: While I can see that this is giving the expected behavior, I don't entirely understand why we need to do this setting of time to right before midnight or to midnight. Could you add a docstring explaining why this is needed?

JeremyGibson (changes requested):

JeremyGibson (approved):
:shipit:

PR #169, opened 2021-08-25
refs #NRH-333 and NRH-334 Adds donor_selected_amount renames fields

JeremyGibson:
#### What's this PR do?
Adds the `donor_selected_amount` to the frontend formData. This also handles the request for NRH-334 and renames the fields that were.

1. Add `donor_selected_amount` to Contribution Metadata as a Payment object.
2. Create a payment either subscription or one-time
3. Navigate to the Payment in Stripe and observe the `donor_selected_amount` in the Metadata.

`donor_selected_amount` needs to be added in the django admin for each deployed environment.

### DIFF
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -233,9 +233,9 @@ class ContributionMetadata(IndexedTimeStampedModel):
     """
 
     lookup_map = {
-        "re_contributor_id": _get_contributor_id,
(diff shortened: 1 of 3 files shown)


michael-caktus (approved):

michael-caktus:
Are we calling this a bugfix and therefore merging it to develop before the other items in PR?

PR #179, opened 2021-08-30
[NRH-330] Footer RevProgram name, email

michael-caktus:
DON’T MERGE ME IN TO DEVELOP UNTIL:
- [x]  All the bugs have been merged in to develop and verified.

Don't REIVEW until:
- [x] Further changes, as discussed in [this basecamp post](https://3.basecamp.com/3120230/buckets/21497549/todos/4083679852), have been added.

#### What's this PR do?
Replaces org name in footer content with rev program name, removes field organization.contact_email, adds revenue_program.contact_email and uses it in footer as well.

Additionally, replaces org address with revprogram address, and the copyright text has the revprogram name, rather than the org name.

The greater effort here, in the end, was generalizing addresses, such that Address is a model with a one-to-one relationship with both Organization and RevenueProgram. There's a data migration included that should bring any old org.addr1 etc over to the related org.address relationship.

Check footer content for above changes.

No.


apps/common/models.py
```diff
@@ -2,10 +2,25 @@
...
+    address1 = models.CharField(max_length=255, blank=True, verbose_name="Address 1")
+    address2 = models.CharField(max_length=255, blank=True, verbose_name="Address 2")
+    city = models.CharField(max_length=64, blank=True, verbose_name="City")
+    state = models.CharField(max_length=2, blank=True, choices=STATE_CHOICES, verbose_name="State")
+    postal_code = models.CharField(max_length=9, blank=True, verbose_name="Postal code")
+
+    def __str__(self):
+        address2 = " " + self.address2 if self.address2 else ""
+        postal_code = " " + self.postal_code if self.postal_code else ""
+        return f"{self.address1}{address2}, {self.city}, {self.state}{postal_code}"
```
JeremyGibson, line 26: __str__ is typically used for humans, and __repr__ is for machine consumption.  Since it looks like you are using this further down the line to send to the frontend. I would put this in the `__repr__` and I would have the string offer a bit more context like `return f"{self.organization}: {self.city} {self.state}"`. It's not a big deal but just a tad more pythonic.

apps/organizations/serializers.py
```diff
@@ -38,15 +38,16 @@ class RevenueProgramListInlineSerializer(serializers.ModelSerializer):
     the user is redirected to `/edit/undefined/page-slug` after page creation.
     """
 
+    address = serializers.SerializerMethodField()
+
+    def get_address(self, obj):
+        if hasattr(obj, "address"):
+            return str(obj.address)
```
JeremyGibson, line 45: if you like the suggestion above `repr(obj.address)`

spa/cypress/integration/donation-page.spec.js
```diff
@@ -206,7 +206,7 @@ describe('Donation page', () => {
     });
   });
 
-  describe('Donation page side effects', () => {
+  describe.only('Donation page side effects', () => {
```
JeremyGibson, line 209: remove the only

JeremyGibson (changes requested):
I'm fine if you want to keep the str representation in address. 

The `only` should be removed. 

And I am curious about the renaming of ModelAdmin to admin.ModelAdmin.

apps/common/admin.py
```diff
@@ -1,10 +1,10 @@
-from django.contrib.admin import ModelAdmin
+from django.contrib import admin
 from django.db import models
 
 from django_json_widget.widgets import JSONEditorWidget
 
 
-class RevEngineBaseAdmin(ModelAdmin):
+class RevEngineBaseAdmin(admin.ModelAdmin):
```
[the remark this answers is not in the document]
michael-caktus: Ah, accident. Leftover from a change that I ended up removing. Switching it back.

apps/common/models.py
michael-caktus: It sounds like by that convention **str** should be okay here, since it's going out to the user. One problem with providing more context is that we don't know if this Address is related to an Org or a RevProgram. Thoughts?

JeremyGibson: If it could be either, perhaps a simple check f"{self.organization if self.organization else self.rev_program}: ..." I mostly just concerned with looking through these objects in the shell later on down the road and all we see is just the address. Not a big deal in the grand scheme.

JeremyGibson (approved):

PR #210, opened 2021-09-14
[NRH-361] OrgAdmin page templates

michael-caktus:
#### What's this PR do?
Implements template creation and template use from the Org Admin.

From the page edit view, notice a new button that you can use to create a template from the existing page. Try it out.
Then, from the page-create modal, notice a new optional dropdown for selecting a template from which to make your new page. Try that out too!

No

### DIFF
--- a/apps/pages/tests/test_viewsets.py
+++ b/apps/pages/tests/test_viewsets.py
@@ -87,15 +87,6 @@ class PageViewSetTest(AbstractTestCase):
         self.assertIn("revenue_program", response.data)
         self.assertIn("slug", response.data["revenue_program"])
 
-    def test_page_create_adds_template_values_when_valid_pk(self):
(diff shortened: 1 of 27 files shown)


apps/pages/tests/test_viewsets.py
```diff
@@ -86,6 +87,16 @@ def test_page_create_returns_revenue_program_slug(self):
...
 
+    def test_page_create_adds_template_values_when_valid_pk(self):
+        pass
+
+    def test_page_create_returns_validation_error_when_invalid_template_pk(self):
+        pass
+
+    def test_page_create_returns_validation_error_when_template_has_invalid_styles_pk(self):
+        pass
+
```
JeremyGibson, line 98: Empty tests.

PR #215, opened 2021-10-05
[NRH-368] Currencies support

michael-caktus:
#### What's this PR do?
Broadly, adds the ability to select a currency per Organization transactions will occur in.

Stripe also expects a country code to be provided in some cases. It made sense to add country to organization.address as well.

The available options are configured with the following two settings (these are also the current values):
```python
COUNTRIES = ["US", "CA"]
# Map currency-code to symbol
CURRENCIES = {"USD": "$", "CAD": "$"}
```

1. In a new or existing Organization, select a currency from django admin and save it. 
2. Then make a donation to that org.
3. Go to stripe dashboard and note the currency (I think? I used the API to test it...)

Yes, sort of. Any additional currencies or countries we want to support here need to be added to the settings mentioned above.


JeremyGibson (approved):
Only a couple of comments no blockers.  I did change the base on this since 376 had been merged in.

michael-caktus:
This got merged in to a multi-feature branch and is on develop

PR #216, opened 2021-10-05
[NRH-373] Donation form validation focus

michael-caktus:
#### What's this PR do?
When a donation form is submitted, the first error is not focused (not just scrolled to).

Submit a donation without filling in name, or address

No


JeremyGibson (approved):
:shipit:

PR #244, opened 2021-10-29
updates reason for giving language

tristanloper:
#### What's this PR do?
updates reason for giving language

view a template such as https://support.revengine-test.org/edit/new-rev-program/caktus-donations and confirm the language changes from "excited to support" to "i support your work because…" and "inspired" to "reason for giving"

I don't think so


x110dc (approved):
LGTM

PR #251, opened 2021-11-03
[NRH-5038] Disable ApplePayDomain.create for MVP

michael-caktus:
#### What's this PR do?
We decided to simply disable the automatic ApplePayDomain creation for MVP. We need to register domains with Apple per revenue-program that wants to use Apple Pay. This will be a separate feature.

Not really a way to do this.

No


JeremyGibson:
I assume leaving the commented code is for future re-enabling. Ship-it.

JeremyGibson (approved):

PR #261, opened 2021-11-11
AXE-CYPRESS

michael-caktus:
This is a Draft. For now, we're testing CI integration.

#### Here's what's done as of this PR:

- [x] Removed all traces of Cypress
- [x] Installed `react-testing-library` and various corollaries
- [x] Tested Login.js as proof-of-concept

#### With this test run we hope to:

- [ ] Successfully integrate with CI

#### What must be done before this effort can be considered successful:

- [ ] Write tests to cover key functionality

#### What would be even better

- [ ] Write tests until a reasonable minimum coverage threshold is met

#### Discussion
What is a reasonable coverage threshold for an app like this, with tests like these, and a coverage report that covers all non-style JS files? It seems infeasible to reach the same coverage threshold as the python test suite (96% ATTOW). The way this app is built, there are just _a lot_ of files. My instinct is to cover as much of the code as is immediately "user-facing", and include some unit tests where it seems useful (maybe utilities?).


spa/src/components/authentication/Login.js
```diff
@@ -36,11 +36,6 @@ function Login({ onSuccess, message }) {
     else history.push(CONTENT_SLUG);
   };
 
-  const handleForgotPassword = (e) => {
-    e.preventDefault();
-    window.location = PASSWORD_RESET_URL;
-  };
-
```
LorienOlive, line 43: Is this cypress related?  Was just curious about why it's included for my own understanding.

LorienOlive (approved):
I need a bit more time in the codebase before I'd be able to provide a thorough review.  I didn't see anything that stood out as wrong, though I did have one comment about a change that seemed unrelated to testing, more for my edification.

spa/src/components/authentication/Login.js
michael-caktus: Hey, sorry, I never saw this comment until now! In case you're still curious-- the <button> which onClick navigates to a new url should really just be an anchor tag, which the updated in the next file in this diff. That's why this is removed.

PR #262, opened 2021-11-11
[NRH-5051] DonationPage header-logo cutoff

michael-caktus:
#### What's this PR do?
Fixes a bug in which the header-logo can potentially but cut off on the x-axis on smaller devices.

Test [this page](https://vosd.revengine-test.org/voice-of-san-diego-donation-page) at smaller viewport widths (assuming nobody has changed the header image)

No

### DIFF
--- a/spa/src/components/donationPage/DonationPageNavbar.styled.js
+++ b/spa/src/components/donationPage/DonationPageNavbar.styled.js
@@ -8,11 +8,15 @@ export const DonationPageNavbar = styled.header`
   display: flex;
   justify-content: center;
   align-items: center;
+  padding: 0 3rem;
 
   box-shadow: ${(props) => props.theme.shadows[0]};
 `;
 
 export const DonationPageNavbarLogo = styled.img`
+  display: block;
   max-height: 50px;
   width: auto;
+  max-width: 100%;
+  margin: 0 auto;
 `;


spa/src/components/donationPage/DonationPageNavbar.styled.js
```diff
@@ -8,11 +8,15 @@ export const DonationPageNavbar = styled.header`
...
   box-shadow: ${(props) => props.theme.shadows[0]};
 `;
 
 export const DonationPageNavbarLogo = styled.img`
+  display: block;
   max-height: 50px;
   width: auto;
+  max-width: 100%;
+  margin: 0 auto;
 `;
```
LorienOlive, line 22: So, I'm not sure if I'm testing this correctly, but I followed the link in the PR, where these styles do not appear to be applied. The "actual" and "expected" layouts in the ticket seem to be the same to me, so I'm also not sure what it should ultimately look like.  However, when I add these styles to this element, it looks like this: 
image

Seems like it might be good to have a bit of padding on either side, that would be my only suggestion.

LorienOlive (commented):
See comment for my concern/suggestion.  I can't see this code live on the staging link provided and I don't think I have a donation page with a logo like this for local testing.  But it did look like it could use some right and left padding after I applied these styles in the browser.

PR #271, opened 2021-11-18
News revenue hub  sc 3591 set hub strip publishable key as environment

LorienOlive:
#### What's this PR do?
This PR adds all of the front-end env vars to the django app and to the Heroku account, to be set on the window via a django request on the index.html of the React app.

Note: I've updated the code to handle json data via django template safely, though it doesn't seem to be working as expected.  Probably some formatting thing I'm unaware of.

Start up python server and navigate to localhost:8000 check window object to see if .ENV object is present.

Please review to ensure I've included all necessary config vars.  I wasn't able to find a Google Analytics ID, for example.


revengine/views.py
```diff
@@ -21,8 +22,12 @@ class ReactAppView(TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         self._add_social_media_context(context)
+        self._serve_spa_env_vars()
         return context
 
+    def _serve_spa_env_vars(self):
+        self.request.spa_env = json.dumps(settings.SPA_ENV_VARS)
+
```
LorienOlive, line 30: Not certain I'm doing this correctly.

spa/public/index.html
```diff
@@ -15,6 +15,13 @@
     </script>
     {% endif %}
 
+    {% if request.spa_env %}
+    {{ request.spa_env|json_script:"spa_env" }}
+    <script>
+      window.ENV = JSON.parse(document.getElementById('spa_env').textContent);
+    </script>
+    {% endif %}
+
```
LorienOlive, line 24: hypothetically the json_script should generate a new script tag, the contents of which can be read by the next script tag, but it's not working properly locally at least.

revengine/views.py
JeremyGibson: No you will need to add the SPA_ENV_VARS to a context variable after the call to super(). 

context['SPA_ENV_VARS'] = self._serve_spa_env_vars()

or without the function call:

context['SPA_ENV_VARS'] = json.dumps(settings.SPA_ENV_VARS)

spa/.gitignore
```diff
@@ -12,6 +12,7 @@
 /build
 
 # misc
+.envrc
```
michael-caktus, line 15: Isn't this redundant? This is already in /.gitignore, right? There shouldn't be an .envrc in /spa, as far as I know.

spa/public/index.html
michael-caktus: Alas, this won't work in a deployed environment. Both of the script tags used here need a "nonce=<that nonce tho>" attribute. Easy enough to add it to the script on line 20, but there doesn't seem to be a way to add arbitrary attributes to the script generated by `json_script`. We _can_, however,  create our own template filter (`json_script` is one of those) that wraps `json_script` and allows us to add the nonce. And maybe even does a little bit more.

spa/src/ajax/axios.js
```diff
@@ -1,9 +1,9 @@
 import axios from 'axios';
 import { LS_CSRF_TOKEN, CSRF_HEADER } from 'constants/authConstants';
 import { TOKEN } from './endpoints';
-import { REACT_APP_API_VERSION } from 'settings';
+import { REACT_API_VERSION } from 'constants/genericConstants';
```
michael-caktus, line 4: I think `REACT_API_VERSION` here should be `REVENGINE_API_VERSION` or something like that. This is the version of the backend API that the frontend should be hitting.

spa/src/constants/analyticsConstants.js
```diff
@@ -1,7 +1,7 @@
 export const HUB_ANALYTICS_APP_NAME = 'rev-engine-analytics';
 export const HUB_GA_V3_PLUGIN_NAME = 'ga-v3-hub';
 // when NRH-76 has been resolved, should use line below instead of two below
-// export const { REACT_APP_HUB_V3_GOOGLE_ANALYTICS_ID: HUB_GA_V3_ID } = process.env;
+// export const { HUB_V3_GOOGLE_ANALYTICS_ID: HUB_GA_V3_ID } = process.env;
```
michael-caktus, line 4: I think the comments in this file apply to this PR-- we should set line 8 (not visible here), which is:
```export const HUB_GA_V3_ID = NONSENSICAL_HUB_GA_V3_ID;```

to 

```export const HUB_GA_V3_ID = window.ENV.HUB_V3_GOOGLE_ANALYTICS_ID;```

File: revengine/settings/base.py [hunk 6]
```diff
@@ -405,5 +405,16 @@
...
+SPA_ENV_VARS = {
+    "HUB_STRIPE_API_PUB_KEY": os.getenv("SPA_ENV_HUB_STRIPE_API_PUB_KEY"),
+    "CAPTURE_PAGE_SCREENSHOT": os.getenv("SPA_ENV_CAPTURE_PAGE_SCREENSHOT"),
+    "APP_SALESFORCE_CAMPAIGN_ID_QUERYPARAM": os.getenv("SPA_ENV_APP_SALESFORCE_CAMPAIGN_ID_QUERYPARAM"),
+    "FREQUENCY_QUERYPARAM": os.getenv("SPA_ENV_FREQUENCY_QUERYPARAM"),
+    "AMOUNT_QUERYPARAM": os.getenv("SPA_ENV_AMOUNT_QUERYPARAM"),
+    "REACT_API_VERSION": os.getenv("SPA_ENV_REACT_API_VERSION"),
+    "STRIPE_API_VERSION": STRIPE_API_VERSION,
+    "HUB_GOOGLE_MAPS_API_KEY": os.getenv("SPA_ENV_HUB_GOOGLE_MAPS_API_KEY"),
+}
```
michael-caktus, line 420: Need to also add `HUB_V3_GOOGLE_ANALYTICS_ID` here.

revengine/settings/base.py
```diff
@@ -405,5 +405,16 @@
...
 STRIPE_API_VERSION = "2020-08-27"
+
+# Front End Environment Variables - Config Vars Heroku
+SPA_ENV_VARS = {
+    "HUB_STRIPE_API_PUB_KEY": os.getenv("SPA_ENV_HUB_STRIPE_API_PUB_KEY"),
+    "CAPTURE_PAGE_SCREENSHOT": os.getenv("SPA_ENV_CAPTURE_PAGE_SCREENSHOT"),
+    "APP_SALESFORCE_CAMPAIGN_ID_QUERYPARAM": os.getenv("SPA_ENV_APP_SALESFORCE_CAMPAIGN_ID_QUERYPARAM"),
+    "FREQUENCY_QUERYPARAM": os.getenv("SPA_ENV_FREQUENCY_QUERYPARAM"),
+    "AMOUNT_QUERYPARAM": os.getenv("SPA_ENV_AMOUNT_QUERYPARAM"),
+    "REACT_API_VERSION": os.getenv("SPA_ENV_REACT_API_VERSION"),
```
michael-caktus, line 417: See `REACT_API_VERSION` = `REVENGINE_API_VERSION` comment below.

michael-caktus (changes requested):

spa/.gitignore
LorienOlive: yeah. that's weird. sorry.

michael-caktus (approved):

michael-caktus:
Approved, aside from the test failure and merge conflict. Then, onward!

michael-caktus:
I think the failure is misleading-- I think the actual failure here is this:
`> Cannot read properties of undefined (reading 'HUB_GOOGLE_MAPS_API_KEY')`

In test, window.ENV is undefined. We might set window.ENV to {{ spa_env }} || {}, or just make all the refs to window.ENV look like this `window.ENV?.MY_ENV_VAR`

PR #278, opened 2021-11-30
[#9717] DonationPage <title> is RevProgram name

michael-caktus:
#### What's this PR do?
Sets the title of the page to "Join | {{ rev program name }}"

Visit a donation page

No

### DIFF
--- a/revengine/settings/dev.py
+++ b/revengine/settings/dev.py
@@ -46,3 +46,7 @@ if os.getenv("TEST_EMAIL", "False") == "True":
 BROKER_URL = "redis://localhost:6379"
 CELERY_RESULT_BACKEND = "redis://localhost:6379"
 CELERY_IMPORTS = ("apps.emails.tasks",)
+
+# Serve SPA via django
+FRONTEND_BUILD_DIR = Path(BASE_DIR) / "spa/public"
+TEMPLATES[0]["DIRS"] = [FRONTEND_BUILD_DIR, os.path.join(PROJECT_DIR, "templates")]
--- a/revengine/views.py
+++ b/revengine/views.py
@@ -20,20 +20,21 @@ class ReactAppView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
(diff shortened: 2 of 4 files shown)


revengine/views.py
```diff
@@ -20,20 +20,21 @@ class ReactAppView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        self._add_social_media_context(context)
+        if revenue_program := self._get_revenue_program_from_subdomain():
+            self._add_social_media_context(revenue_program, context)
+            context["revenue_program_name"] = revenue_program.name
```
LorienOlive, line 25: I gather this change is because social media is only necessary on donation pages. Is that correct?

revengine/settings/dev.py
```diff
@@ -46,3 +46,7 @@
 BROKER_URL = "redis://localhost:6379"
 CELERY_RESULT_BACKEND = "redis://localhost:6379"
 CELERY_IMPORTS = ("apps.emails.tasks",)
+
+# Serve SPA via django
+FRONTEND_BUILD_DIR = Path(BASE_DIR) / "spa/public"
+TEMPLATES[0]["DIRS"] = [FRONTEND_BUILD_DIR, os.path.join(PROJECT_DIR, "templates")]
```
LorienOlive, line 52: I think this is the only part i'm not sure of. This is for running locally correct?  Does it just reflect a restructuring of folders or does it change something about how it works?

LorienOlive (approved):
Looks good. I checked on one of my test donation pages on localhost:8000 and the page title did say "Join".  Left a few comments, but they're really just questions to make sure I understand a couple of the changes.

revengine/settings/dev.py
michael-caktus: This _should_ just be a copy of the code you added in one of your branches (that I hadn't been merged in to dev at the time of authoring this ticket). This is the bit that makes the django app serve the right index.html locally. The thing we added to make your env work for donation pages a little while back.

PR #279, opened 2021-11-30
[#9726] Apple/Google Pay Label is RevProgram

michael-caktus:
#### What's this PR do?
Makes the payment request label the rp name, not the generic "revengine donation"

Start an apple or google pay payment

No


LorienOlive (approved):
Looks correct, based on the code, but I didn't see it come through in the request data.  Maybe I'm not looking in the right place, or perhaps I don't have things set up properly (see console errors below).

image

michael-caktus:
Neither Google Pay nor Apple Pay will be usable locally unless you tunnel through an HTTPS proxy like ngrok (apple and google pay require https). So that bit of it is not easy to test locally. I'm content to test this on staging!

PR #280, opened 2021-11-30
[#4517] Custom Font Support

michael-caktus:
#### What's this PR do?
Support for choosing fonts from a static list of fonts. Current typography types supported are "heading" and "body". Adds a model to the django-admin, `Font` under the "Pages" app. Here you define specify:
- name: how you want this font to be displayed in the org admin
- source: for now either "Google" or "Typekit"
-  Font name: must be the name of the font according to CSS (this is usually the same as "Name" above)
- accessor: For google fonts, just the font_name. For typekit, the kit id for the font.

Each font you define here is available to choose from on your stylesheet.

Customize your fonts

No


LorienOlive (commented):
Screen Shot 2021-12-06 at 7 13 21 PM
Screen Shot 2021-12-06 at 7 13 57 PM


I'm not sure this is related to this ticket, but I'm seeing this in the edit mode of a rev program.

LorienOlive (commented):
I may not understand the full flow, but I tried adding "Roboto" as a google font.  I didn't see it show up in the editor as an option, but maybe it's not supposed to?

michael-caktus:
Sorry, I'm not sure I see the problem in those images.


j/k.

Commit coming up with a fix for that.  The failed prop warnings are the results of some collision between styled-components and other libraries that forward props carte blanche. Those have been around a bit, but there's a ticket to find a way to fix/suppress them. The Google API key error is probably just because you don't have a Maps API key defined locally, which is fine. Doesn't break anything except the address autocomplete on the donation form.

PR #282, opened 2021-12-01
[#9728] OAuth Connect Refactor

michael-caktus:
#### What's this PR do?
After discovering that Stripe clones existing accounts if they're onboarded via Stripe Connect Onboarding, we determined that a better MVP method of onboarding existing stripe accounts would be to utilize Stripe's OAuth flow for Connecting Standard accounts. This PR implements that refactor. The change is all but invisible to the user. We'll want to do a bunch of verification around accepting payments and things after this work is on staging.

Onboard an org with an existing account. This is harder than it sounds, since you'll probably want to use a test account, but I'm not sure you can create a test account from the get-go. Might need to get creative here.

Yes, but it should already be there.
`STRIPE_CLIENT_ID` which is a value found in the Stripe dashboard. I've already added these in both environments.


michael-caktus:
Note to self: Look in to product ID generation here.

michael-caktus:
Confirmed that Production creation is still working as expected. Product.create is called once an org is fully verified. Nothing has changed there, and this should still work.

JeremyGibson (approved):

PR #285, opened 2021-12-03
Create componentLoader util and ChunkErrorBoundary component + styles

LorienOlive:
#### What's this PR do?
Creates a custom component loader to handle chunk loading errors from code splitting as well as an error boundary that shows the error and reloads the page.

You can try to test locally, I can provide the settings for throttling, though I was never able to trigger it.  My guess is that it must be tested in a deployed environment.

### DIFF
--- a/spa/src/components/errors/ChunkErrorBoundary.js
+++ b/spa/src/components/errors/ChunkErrorBoundary.js
@@ -1,21 +1,23 @@
-import React, { Fragment, useState, useEffect } from 'react';
+import React from 'react';
 import * as Sentry from '@sentry/react';
 
-import Button from 'elements/buttons/Button';
 import * as S from '../../elements/buttons/Button.styled';
 import * as ErrorS from './ErrorFallback.styled';
 
-const ChunkErrorFallback = (error, componentStack, handleReset, resetError) => {
+const ChunkErrorFallback = (error, componentStack, resetError) => {
+  console.log(error, componentStack, resetError);
   const chunkFailedMessage = /Loading chunk [\d]+ failed/;
 
   if (error?.message && chunkFailedMessage.test(error.message)) {
+    // Leaving this here for testing purposes, should retry only for chunk loading errors
+    console.error(error);
     resetError();
   } else {
     return (
-      <Fragment>
+      <>
         <ErrorS.ErrorWrapper>
           <ErrorS.ErrorHeading layout>
-            <h2>You have encountered an error</h2>
+            <h2>😰 We've encountered a problem. Click below to reload the page</h2>
           </ErrorS.ErrorHeading>
           <ErrorS.ErrorMessage>
             <div>{error.toString()}</div>
@@ -23,34 +25,24 @@ const ChunkErrorFallback = (error, componentStack, handleReset, resetError) => {
           <ErrorS.ErrorStack>
             <div>{componentStack}</div>
           </ErrorS.ErrorStack>
-          <S.Button>
-            <Button
-              onClick={() => {
-                handleReset('An error has occurred');
-                resetError();
-              }}
-            >
-              Click here to reload this page
-            </Button>
+          <S.Button
+            onClick={() => {
+              resetError();
+            }}
+          >
+            Reload Page
           </S.Button>
         </ErrorS.ErrorWrapper>
-      </Fragment>
+      </>
     );
   }
 };
 
 const ChunkErrorBoundary = (props) => {
-  const [message, setMessage] = useState();
-
-  const handleReset = (newMessage) => {
-    setMessage(newMessage);
-  };
-
+  console.log(props);
   return (
     <Sentry.ErrorBoundary
-      fallback={({ error, componentStack, resetError }) =>
-        ChunkErrorFallback(error, componentStack, handleReset, resetError)
-      }
+      fallback={({ error, componentStack, resetError }) => ChunkErrorFallback(error, componentStack, resetError)}
     >
       {props.children}
     </Sentry.ErrorBoundary>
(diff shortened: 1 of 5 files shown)


spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -0,0 +1,60 @@
...
+            </Button>
+          </S.Button>
+        </ErrorS.ErrorWrapper>
+      </Fragment>
+    );
+  }
+};
+
+const ChunkErrorBoundary = (props) => {
+  const [message, setMessage] = useState();
```
michael-caktus, line 43: This doesn't seem to get used anywhere. Leftovers from a previous iteration?

spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const handleReset = (newMessage) => {
+    setMessage(newMessage);
+  };
```
michael-caktus, line 47: This seems to be called, but to no effect.

spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -0,0 +1,60 @@
...
+          <ErrorS.ErrorMessage>
+            <div>{error.toString()}</div>
+          </ErrorS.ErrorMessage>
+          <ErrorS.ErrorStack>
+            <div>{componentStack}</div>
+          </ErrorS.ErrorStack>
+          <S.Button>
+            <Button
+              onClick={() => {
+                handleReset('An error has occurred');
```
michael-caktus, line 29: This can probably be removed-- it doesn't appear to do anything anymore
```suggestion
```

spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -0,0 +1,60 @@
...
+                resetError();
```
michael-caktus, line 30: From what I can tell, resetError will try to render the children of the ErrorBoundary, but I don't think it will try to re-import them.  When I run this locally it renders a blank page without an error (but that could be because of the way I'm triggering an error)

spa/src/utilities/componentLoader.js
```diff
@@ -0,0 +1,17 @@
+const componentLoader = (lazyComponent, attemptsLeft = 3, interval = 1000) => {
```
michael-caktus, line 1: Love this! Might not be unreasonable to bump the number of attempts up to 5 or even 10. This doesn't use a lot of resources.

spa/src/utilities/componentLoader.js
```diff
@@ -0,0 +1,17 @@
...
+  return new Promise((resolve, reject) => {
+    lazyComponent()
+      .then(resolve)
+      .catch((error) => {
+        setTimeout(() => {
+          if (attemptsLeft === 1) {
+            reject(error);
+            return;
+          }
+          componentLoader(lazyComponent, attemptsLeft - 1, interval).then(resolve, reject);
```
michael-caktus, line 11: What if we do interval * 1.5 or something, to give it that exponential backoff feeling? We'd probably want to cap the attempts at 5 or so if we did that, counter to what I said above

michael-caktus (changes requested):

LorienOlive:
Hey Michael,  So I reviewed all of these.  In summary, my main goal was to just get this into staging and see if I could catch a chunk loading error in the wild. But if you'd rather have the code be as close to complete as possible before merging into staging (understandable).  I can make a local test branch with the extra code for local testing, so it's there in case I need to fix anything further.  I'll work on this first thing Monday.

spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import * as Sentry from '@sentry/react';
+
+import * as S from '../../elements/buttons/Button.styled';
+import * as ErrorS from './ErrorFallback.styled';
+
+const ChunkErrorFallback = (error) => {
```
michael-caktus, line 7: error prop not used here

spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -0,0 +1,33 @@
...
+          Refresh
+        </S.Button>
+      </ErrorS.ErrorWrapper>
+    </>
+  );
+};
+
+const ChunkErrorBoundary = (props) => {
+  return (
+    <Sentry.ErrorBoundary fallback={({ error }) => ChunkErrorFallback(error)}>{props.children}</Sentry.ErrorBoundary>
```
michael-caktus, line 29: I think we can jsut do this:

```suggestion
    <Sentry.ErrorBoundary fallback={<ChunkErrorFallback />}>{props.children}</Sentry.ErrorBoundary>
```

Or something like it.

spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -0,0 +1,33 @@
...
+
+import * as S from '../../elements/buttons/Button.styled';
+import * as ErrorS from './ErrorFallback.styled';
+
+const ChunkErrorFallback = (error) => {
+  return (
+    <>
+      <ErrorS.ErrorWrapper>
+        <ErrorS.ErrorHeading layout>
+          <h2>😰 We've encountered a problem!</h2>
```
michael-caktus, line 12: I personally love emojis on websites, but I'm guessing this isn't the right tone for Hub.

michael-caktus (changes requested):

LorienOlive:
@michael-caktus re:  emoji.  I'm kinda literal brained, so I use the content you gave me, esp since I'm not as familiar with the client as you are.  Just an fyi for the future.

PR #286, opened 2021-12-07
[#9729] call verifyapplepaydomain on revenueprogram creation

michael-caktus:
#### What's this PR do?
Performs the necessary Stripe API call for Apple Pay domain verification whenever a new RevenueProgram is created. Does not handle tear-down.

This must be tested with a live key. Stripe does not allow us to call the ApplePayDomain.create endpoint with a test key. But in prod, just create a new revenue program and verify that it has a "apple_pay_domain_verification_date" that is not blank in Django admin. One should probably also go a step further and verify that apple pay appears on a donation page for that rev program-- but again, only testable live.

We've added a DOMAIN_APEX config variable that should be "revengine.org" for prod and "revengine-test.org" for staging. These should already be set in Heroku.


apps/organizations/signals/handlers.py
```diff
@@ -0,0 +1,10 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+
+@receiver(post_save, sender="organizations.RevenueProgram")
+def my_handler(sender, **kwargs):
```
JeremyGibson, line 6: change def name

JeremyGibson (approved):
Change signal definition name and lgtm

PR #289, opened 2022-01-05
News revenue hub  sc 9593 implement hub gtm for additional third party

LorienOlive:
#### What's this PR do?
This PR corresponds to ticket: https://app.shortcut.com/caktusgroup/story/9593/implement-hub-gtm-for-additional-third-party-tags

It adds two GTM scripts provided by NRH to the index.html.  

I have confirmed that they're working by checking the `dataLayer` in the console, but I assume that someone with access to the google analytics account will have to do more complete testing once this has been deployed to prod.

No, the necessary Google Analytics tag has already been added to the production config vars.  There is no valid GA tag stored in the staging env's config vars.


spa/public/index.html
```diff
@@ -32,9 +41,29 @@
...
+    </script>
+
+    <script async="" src="https://www.googletagmanager.com/gtag/js?id=AW-783491115"></script>
+    <script>
+      window.dataLayer = window.dataLayer || []; 
+      function gtag(){dataLayer.push(arguments);} gtag('js', new Date()); 
+      gtag('config', 'AW-783491115'); 
+    </script>
+
+    <title>RevEngine</title>
```
michael-caktus, line 58: I'm not sure these needs to be included. Having `gtag` and Google Tag Manager is redundant as far as I can tell. My best interpretation of the basecamp post that included these scripts is that the goal is for the Google Tag Manager to accomplish the same thing that they were accomplishing with these gtags.

spa/public/index.html
```diff
@@ -8,6 +8,15 @@
...
     <meta name="description" content="RevEngine" />
 
+    <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+      new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+      j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+      'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+      })(window,document,'script','dataLayer','GTM-N626BSB');
+    </script>
+    <!-- End Google Tag Manager -->
```
michael-caktus, line 18: The basecamp post for this feature describes different scripts for staging and production. The only differences between the scripts, however, are the GTM ids. That means that we need to `'GTM-N626BSB'` here to be {{ request.HUB_GTM_ID }} or some such new config var. That config var needs to be set in Heroku config vars per env.

spa/public/index.html
```diff
@@ -8,6 +8,15 @@
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="RevEngine" />
 
+    <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```
michael-caktus, line 12: This script must be below the "csp-nonce" meta tag on line 20.  Also, we need to add the nonce as an attribute to the script tag, as below:
```suggestion
    <script nonce="{{ request.csp_nonce }}">(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```

spa/public/index.html
```diff
@@ -32,9 +41,29 @@
...
+
+    <title>RevEngine</title>
     <title>{% if revenue_program_name %}Join | {{ revenue_program_name }}{% else %}RevEngine{% endif %}</title>
   </head>
   <body>
+    <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-N626BSB"
+      3height="0" width="0" style="display:none;visibility:hidden"></iframe>
+    </noscript>
+    <!-- End Google Tag Manager (noscript) -->
```
michael-caktus, line 66: See the comment about the the env var above

michael-caktus (changes requested):

spa/public/index.html
LorienOlive: Ok.  they were listed in the scripts that they wanted added in this basecamp link: https://3.basecamp.com/3120230/buckets/21497549/todos/4358543571

LorienOlive: By way of explanation, I added those to confirm that the kinds of scripts they would like to run (as they mention in their basecamp note).  So that was added for testing purposes to make sure that the above tag enabled the other scripts they might want to insert in the future, if that makes sense.  I can remove them, but we might want to reach out to them to get a concrete example of what they might like to use for testing purposes on staging so we can go ahead and add it for QA purposes.  Or they can just insert themselves in the html during their testing.

spa/public/index.html
LorienOlive: Should this be inside the conditional like the other script tag?

LorienOlive:
I made the HUB_GTM_ID settings variable separate from the other SPA_ENV_VARS since the only place it's needed (that I know of is in the index.html file), but let me know if you think it should be part of that object in base.py.

LorienOlive:
I added the HUB_GTM_ID as a separate variable in base.py.  I didn't include it in the SPA_ENV_VARS object since the only place it's needed (as far as I know) is the index.html.  I also added it to local.py so we can test that it's coming through on localhost:8000 in the browser.   

I was also unsure if I should move the GA script tag inside the nonce conditional like the other script tag that relies on it.  Let me know if  that's better.

I removed the additional scripts, see explainer above for why they were there.

michael-caktus (approved):

michael-caktus:
🚀

LorienOlive:
Merging this branch into develop following approval.

PR #290, opened 2022-01-06
#10666 Configure Sentry via Env

michael-caktus:
#### What's this PR do?
Ensures that all uses of the Sentry DSN reference the SENTRY_DSN env var. Also makes the CSP_REPORT_URI a config var so that it's flexible like the SENTRY_DSN now is.

Set the env vars on staging to the desired DSN and-- maybe trigger some error? Not sure about that one.

Add the desired DSN to Heroku config vars as SENTRY_DSN to each environment. Also make sure to set the CSP_REPORT_URI env var in each env.

### DIFF
--- a/revengine/settings/base.py
+++ b/revengine/settings/base.py
@@ -420,7 +420,13 @@ SECURE_REFERRER_POLICY = "strict-origin-when-cross-origin"
 # Stripe API Target Version
 STRIPE_API_VERSION = "2020-08-27"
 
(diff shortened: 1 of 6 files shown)


benjaminEwhite (commented):
Looks good overall. Might make sense to have a test that shows that useSentry hook achieves expected results (perhaps by triggering an error somewhere in component tree???).

michael-caktus:
@benjaminEwhite -- Testing the hook with cypress turns out to be a bit of a nightmare that I'm not sure is worth the effort. It might be worth considering bringing in react-testing-library alongside cypress to test things like this.

PR #292, opened 2022-01-07
#10661 ApplePayDomain.create management command

michael-caktus:
#### What's this PR do?
You can now run `./manage.py appledomainverify --slug my-rp-slug` to verify a RP's subdomain with Apple via Stripe.

This can only be tested in STRIPE_LIVE_MODE, since the ApplePayDomain.create command requires a live key. In Prod, you could create a rev program, then unset the "domain_apple_verified_date" field in admin. This won't "undo" the verification, but if you then run the management command it should reset the value to now.

No.


benjaminEwhite:
LGTM

benjaminEwhite (approved):

PR #293, opened 2022-01-10
#10749 Improve Page Style Color Settings

michael-caktus:
#### What's this PR do?
Implements the changes discussed in [this basecamp post](https://3.basecamp.com/3120230/buckets/21497549/todos/4376261341).

Try out the new values in a new stylesheet. Old styles will not work anymore.

No, other than possibly reworking existing pages to whatever extent you want

### DIFF
--- a/spa/src/components/authentication/Login.styled.js
+++ b/spa/src/components/authentication/Login.styled.js
@@ -8,7 +8,6 @@ export const Login = styled.div`
 `;
 
 export const LoginCard = styled.section`
-  margin-top: 10rem;
   padding: 4rem;
   background: ${(props) => props.theme.colors.white};
   max-width: 450px;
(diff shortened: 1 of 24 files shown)


LorienOlive (commented):
I pulled down and confirmed that the new styles are properly labeled.  Earlier revenue program with old styling threw an error.  However, I had trouble fully creating a new rev program for testing bc there was an error when trying to load a stylesheet. 
image

And I got the same error with the new rev program I created as I did with the old one: 
image

Perhaps these errors are linked and it's just something I need to add to my env?  If it's a simple fix, I'm happy to do that and finish testing. LMK

LorienOlive (changes requested):
Screen Shot 2022-01-18 at 9 23 14 AM
Screen Shot 2022-01-18 at 9 25 59 AM
The previous error has been resolved and the styles are properly renamed and showing up in new rev programs I create.  The only remaining issue I saw can be seen in the above screenshots, taken after creating a new set of styles.  I'm happy to help look into it further, but thought I'd bring it up here in case you already know what might be causing it.

PR #296, opened 2022-01-11
Release 2022-01-11

michael-caktus:
Contains PRs:
PR #290, refs ticket #10666
PR #292, refs ticket #10661
PR #291, refs ticket #9701
PR #287 
PR #288


revengine/settings/base.py
```diff
@@ -417,6 +420,14 @@
 # Stripe API Target Version
 STRIPE_API_VERSION = "2020-08-27"
 
+
+# Sentry Configuration
+SENTRY_ENABLE_FRONTEND = os.environ.get("SENTRY_ENABLE_FRONTEND", "false").lower() == "true"
```
x110dc, line 425: If I set `SENTRY_ENABLE_FRONTEND` to `False` (say for development) and don't set `SENTRY_DSN_FRONTEND` at all won't this throw an exception? (And similarly for `_BACKEND`)

x110dc (changes requested):
One minor question. Otherwise LGTM.

revengine/settings/base.py
michael-caktus: I might be missing something, but what's hopefully happening here is that if `SENTRY_ENABLE_FRONTEND` is "False" (or "false"), then Sentry.init won't be called at all. See `spa/src/hooks/useSentry.js:18`.  Same for `_BACKEND`-- see `revengine/settings/deploy.py:111`

Am I misunderstanding the concern?

Also, we can verify this on staging if we want to clear it up that way. Just set the config vars as you mentioned above in heroku for staging. Main challenge there is then triggering an error. Not totally sure how to do that dependably.

x110dc: No, mea culpa. I was thinking `os.environ['VARIABLE']` where it would throw an exception if undefined. Never mind. I'll approve and merge.

x110dc (approved):
LGTM.

PR #297, opened 2022-01-12
Fix - TypeError in prod - Handle boolean conversion of string spa_env…

LorienOlive:
…_vars in django base settings instead of react settings

Ticket Ref: https://app.shortcut.com/caktusgroup/story/10825/resolve-confusion-around-booleans-strings-from-env-in-js-settings

#### What's this PR do?
This PR resolves a `TypeError` being thrown in the production environment.  The reason for this error is that when config vars containing Booleans are set in Heroku, they are stored as string versions of a python boolean (uppercase "True" or "False").  When the `spa_env` json is served from django to the html and is parsed in a JS script, the value of the `SENTRY_ENABLE_FRONTEND` config var in production (which had already been being lower-cased in base.py) was interpreted as a Boolean.  In `settings.js` there was another attempt to lowercase the string, but it found a Boolean instead, causing the `TypeError`.

This PR ensures that these booleans are being lowercased (and given a default value) in python base settings.  The repeated logic is removed from the `settings.js` file in the `spa`.

In staging, navigate to any nrh page and open the devtools.  In the console, type `window.ENV`, you should find that any variables with the value of `true` or `false` are represented as Booleans, not strings.  The page should render and there should be no `TypeError`.

In order to test this in staging, I set `SENTRY_ENABLE_FRONTEND` to `"True"`.  This isn't a necessary variable for staging and is for testing purposes only.  It could be removed following this test.  No changes to Prod are necessary.

### DIFF
--- a/revengine/settings/base.py
+++ b/revengine/settings/base.py
@@ -431,7 +431,7 @@ SENTRY_DSN_BACKEND = os.environ.get("SENTRY_DSN_BACKEND")
 # Front End Environment Variables - Config Vars Heroku
 SPA_ENV_VARS = {
     "HUB_STRIPE_API_PUB_KEY": os.getenv("SPA_ENV_HUB_STRIPE_API_PUB_KEY"),
-    "CAPTURE_PAGE_SCREENSHOT": os.getenv("SPA_ENV_CAPTURE_PAGE_SCREENSHOT", "false"),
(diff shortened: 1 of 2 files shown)


LorienOlive:
@kmtracey  This is what I see locally, in staging, `SENTRY_ENABLE_FRONTEND` should be `true` when you look at `window.ENV`

kmtracey (approved):
This looks good to me :-)

PR #302, opened 2022-01-24
#9579 New pages start with default content

michael-caktus:
#### What's this PR do?
New pages are created with default content, including an admin configurable default header logo.

Create a new page and observe the default settings.
If you want to test the default header logo as well, go in to django admin and add a "logo" to the DefaultHeaderLogo model and save it. Then create a new page and notice that this is used as the header logo.

No, aside from setting the default header logo


benjaminEwhite (commented):
Overall, looks good. Couple small comments, and a question re: hardcoded UUIDs in defaults.

PR #303, opened 2022-01-26
#10808 Limit DjangoAdmin select options based on other values

michael-caktus:
This is a draft PR to see if anybody can see a less baroque way to handle this.

There are several places where we're limiting dropdown value based on other values in DjangoAdmin. All of these changes are in change/add forms

1. In BenefitLevels, selecting an Organization should change the values available to the Benefits Inline dropdowns.
2. In RevenueProgram, selecting an Organization should limit the values available to the BenefitLevels Inline dropdowns (Add only)
3. In RevenueProgram, Organization should be read-only on the change form. The "default donation page" field should only show donation pages for the org that is previously selected. You should not be able to select a default donation page on initial RevenueProgram creation.
4. In DonationPage, selecting a RevenueProgram should change the values available to the Styles dropdown.

These are handled in two different ways. All of the above except number three are handled via an InclusionTag added to the ModelAdmin changeform template. This inclusion tag adds some javascript to the changeform. The javascript makes an ajax request whenever the parent dropdown is changed that hits a view which in turn accesses a property on the parent model that must be defined and must return a list of options for the front end. The view returns this as json and the javascript updates the child select widget accordingly.

Number three is handled in a no-less convoluted way, but there's no javascript necessary. We override the RevenueProgramAdmin `formfield_for_db` method and set the `default_donation_page` limit_choices_to property to a Q object filtering by revenue program. The particularly convoluted part of this is that in order for the RevenueProgram instance to be available to the `formfield_for_db` method, we have to create a partial out of `formfield_for_dbfield` in another method and pass it the obj from there.  A bit strange, but it works!

I'd love to get some eyes on this to see if:
- There are better ways of handling these problems?
- Is this code even remotely understandable? Maintainable? Any improvements here?


apps/common/tests/test_views.py
```diff
@@ -15,3 +21,75 @@ def test_page_title_includes_rev_program_name_when_subdomain(self):
...
+        response = self.client.patch(reverse("admin-select-options"))
+        self.assertEqual(response.status_code, 405)
+
+        response = self.client.put(reverse("admin-select-options"))
+        self.assertEqual(response.status_code, 405)
+
+        response = self.client.delete(reverse("admin-select-options"))
+        self.assertEqual(response.status_code, 405)
+
+    def test_can_only_be_accessed_by_staff(self):
```
benjaminEwhite, line 49: Test name is possibly slightly misleading because doesn't demonstrate that staff can access. Maybe rename test to `test_cannot_be_accessed_by_non_staff`.

apps/organizations/admin.py
```diff
@@ -31,6 +33,13 @@ class BenefitLevelBenefit(admin.TabularInline):
     verbose_name_plural = "Benefits"
     extra = 0
 
+    def get_formset(self, request, obj=None, **kwargs):
```
benjaminEwhite, line 36: Maybe docstring explaining use case of the override. I gather `can_add_related` and `can_change_related` are parameters on a built-in Django admin object?

apps/organizations/admin.py
```diff
@@ -120,6 +129,14 @@ class BenefitLevelAdmin(RevEngineBaseAdmin):
 
     inlines = [BenefitLevelBenefit]
 
+    # Overriding this template to add the `admin_limited_select` inclusion tag
+    change_form_template = "organizations/benefitlevel_changeform.html"
+
+    def get_readonly_fields(self, request, obj=None):
```
benjaminEwhite, line 135: Docstring here explaining purpose of override would be helpful.

apps/organizations/tests/test_admin.py
```diff
@@ -58,3 +61,19 @@ def test_feature_add_limit_fail(admin_client):
         soup = bs4(response.content)
         assert soup.find(class_="errorlist nonfield")
     assert Feature.objects.all().count() == 0
+
+
+def test_revenue_program_default_donation_page_options_limited(admin_client):
```
benjaminEwhite, line 66: This test looks good, but leaving a parallel comment here: a lot of the logic in this PR is about the dynamic update of available values based on user selecting a parent. Ideally would test that logic, but that's probably to E2E of a test given we're in Python world and don't want Selenium in the mix. Barring that, could at least test the admin options endpoint, for happy path, plus the two unhappy paths it tries to catch.

revengine/views.py
```diff
@@ -48,3 +51,37 @@ def _add_gtm_id(self, context):
...
+# method on every model via the `admin_select_options` view-- like some
+# terrible accidental RPC API.
+SAFE_ADMIN_SELECT_PARENTS = ["organizations.Organization", "organizations.RevenueProgram"]
+SAFE_ADMIN_SELECT_ACCESSOR_METHODS = ["admin_style_options", "admin_benefit_options", "admin_benefitlevel_options"]
+
+
+@require_GET
+@login_required
+@staff_member_required
+def admin_select_options(request):
```
benjaminEwhite, line 66: See comment above about possibly testing this.

PR #304, opened 2022-01-28
[#11590] Move dislcaimer text to payent widget

michael-caktus:
#### What's this PR do?
The initial implementation of the static disclaimer text placed it at the bottom of the page no matter where the credit card form was. That didn't make sense, so we've tied the disclaimer text to the submit button (or payment request button)

Move around your Payment element and notice that the disclaimer text moves with it.

No


benjaminEwhite (approved):
LGTM 👍

PR #305, opened 2022-01-28
#11447 RevenueProgram and DonationPage slug block list

michael-caktus:
#### What's this PR do?
Implements [this requirement](https://3.basecamp.com/3120230/buckets/21497549/todos/4515004087). Adds a "config" app. "Config" app contains a `Deny-list Word` model. Each instance represents a word that will not be allowed as a RevenueProgram or DonationPage slug. This prohibition works for both explicitly set slugs, and slugs that are inferred from the name of the RP or DP. The check for deny-list words is case insensitive.

Create a deny-list word or two. Then try to create an RP with that slug. Same with DP. You can also test that, given deny-list word "testing", a new DonationPage with name "Testing!" with slug left blank (will be inferred from "Testing!") also fails validation.

No, aside from setting up your deny-list however you'd like via django admin.


apps/config/tests/factories.py
```diff
@@ -0,0 +1,15 @@
+from factory.django import DjangoModelFactory
+from faker import Faker
```
benjaminEwhite, line 2: Can also do this via factory directly, without importing faker. https://factoryboy.readthedocs.io/en/stable/reference.html?highlight=faker#faker

benjaminEwhite (commented):
Overall looks good.

Tested it out locally and works as expected. I did notice that the error message that gets displayed is potentially confusing. Since the validation for slug fails, when the form displays the error, the field for "slug" is blank, if you allow it to autogenerate. The message says "This slug is not allowed" which might prompt the user to think that a blank slug is not allowed.

Screen Shot 2022-01-31 at 10 52 31 AM

PR #307, opened 2022-01-28
Fix header fallback

michael-caktus:
Restores the option to use a header background image. If not present, defaults to stylesheet provided color


michael-caktus:
Shouldn't have been based on previous branch

PR #310, opened 2022-02-03
#11350 Small Text In Rich Text Block

dchukhin:
#### What's this PR do?
This pull request makes it possible to choose the size of text in a rich text block:
 - the rich text editor config now uses the `'fontSize'` option, with a few sizes that match other font sizes on the page
 - a `DISCLAIMER_TEXT_SIZE_PX` constant has been defined, so that changing the disclaimer text size in the donation page styles also displays that text size in the rich text editor

 - find a rich text block on the site (for example, on a test donation page)
 - type some text
 - highlight the text, and choose the size of the text from the new font size dropdown in the rich text area buttons
 - observe that the text looks like the disclaimer text. You can also make the text italic by pressing the italic rich text button
 - see that the text looks ok on small screens, medium screens, and big screens

No


michael-caktus (approved):
Looks great!

PR #316, opened 2022-02-08
#11810 bug: 400 error triggered when making payment

michael-caktus:
#### What's this PR do?
Fixes a bug, the outward signs of which were failed donations for a particular set of pages for a particular RevenueProgram. 

1. It was possible for DonationPages to belong to RevenueProgram 1 and Organization 2, but for RevenueProgram 1 not to belong to Organization 2. This should be an impossible situation.
2. Because of the way the DonationPage was related to RP and Org, whenever our code needed to access Org stripe credentials by way of DonationPage relationships, there were two different paths that could get you to those creds. One was directly via the DonationPage --> Org relationship. The other was DonationPage --> RevenueProgram --> Organization. Because of the issue in number 1 above, this could (and did) result in different places using different credentials to try to complete the same payment. This caused the front-end visible errors that revealed this bug.

The solution was to remove the foreign key on DonationPage to Organization. As things have evolved, it makes less sense for that direct relationship to exist. This obviously solves issue number one. It solves issue number two by eliminating the direct path for getting Org credentials. This was a modest refactor. One possibly unintuitive result of this change was that DonationPages in the Pages list view were no longer sorted by Organization "for free". We've written an additional viewset mixin to filter models based on RevenueProgram. This and some of the other work done here should actually help us quite a lot in the roles + permissions work coming up next.

Two places should be checked out:
1. The pages list view in the Org Admin
2. And of course DonationPages should no longer be able to get in to this state.

It's probably worth mentioning that any DonationPages that were broken before this PR will still be broken after. This PR makes it impossible to get in that state again. We can:
1. Ignore them
2. Trash them
3. Fix them by running a few lines of code in the shell


apps/organizations/views.py
```diff
@@ -24,6 +24,24 @@ def has_object_permission(self, request, *args):
...
+
+        model_has_rp_rel = hasattr(self.model, "revenue_program")
+
+        if is_org_user and model_has_rp_rel and not self.request.user.is_superuser:
+            revenue_programs = self.request.user.get_revenue_programs()
+            return self.model.objects.filter(revenue_program__in=revenue_programs)
+        return self.model.objects.all()
+
+
 class OrganizationLimitedListView:
```
benjaminEwhite, line 45: Just curious why this one is specifically for list view, while `RevenueProgramLimitedMixin` is not limited to list view.

apps/api/permissions.py
```diff
@@ -12,6 +12,9 @@ class UserBelongsToOrg(permissions.BasePermission):
     message = "User does not have permission to access this resource."
 
     def has_object_permission(self, request, view, obj):
+        if request.user.is_superuser:
+            return True
```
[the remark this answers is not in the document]
michael-caktus: This will just bail on the method early though. There are other checks below it.

apps/organizations/views.py
```diff
@@ -24,6 +24,24 @@ def has_object_permission(self, request, *args):
...
+        Filter quersets such that only instances belonging to revenue programs this user has access to are returned.
+        """
+        # Ensure the user in question is a User, not a Contributor
+        is_org_user = isinstance(self.request.user, user_model)
+
+        model_has_rp_rel = hasattr(self.model, "revenue_program")
+
+        if is_org_user and model_has_rp_rel and not self.request.user.is_superuser:
+            revenue_programs = self.request.user.get_revenue_programs()
+            return self.model.objects.filter(revenue_program__in=revenue_programs)
```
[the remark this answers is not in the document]
michael-caktus: Yeah, I think that check really just shouldn't be there. We're preventing non-org-users from accessing this endpoint elsewhere.

apps/organizations/views.py
michael-caktus: I noticed that too-- I don't know, and I can't see why it would, but I didn't want to mess with it...

apps/pages/models.py
```diff
@@ -169,11 +171,11 @@ def clean_fields(self, **kwargs):
 
     def save(self, *args, **kwargs):
         limit = self.has_page_limit()
-        if limit and not self.id:
-            if self.total_pages + 1 > int(limit.feature_value):
-                raise ValidationError(
-                    {"non_field_errors": [f"Your organization has reached its limit of {limit.feature_value} pages"]}
-                )
+        # breakpoint()
```
benjaminEwhite, line 174: Remove this

benjaminEwhite (approved):
There's a commented out breakpoint to remove but otherwise LGTM :shipit:

PR #319, opened 2022-02-09
#5577 Audit Trail

dchukhin:
#### What's this PR do?
This pull request creates an audit trail of changes to various models across the project:
 - `django-simple-history` has been added to the project to track changes for each model
 - changes are visible on each object's detail page in the Django admin (for example, at the /nrhadmin/organizations/organization/1/history/ URL)
 - a new `RevEngineHistoricalChange` model keeps track of all revisions, and exposes them in a Django admin view (at the /nrhadmin/common/revenginehistoricalchange/ URL), showing admin users:
   - who made each change
   - when the change occurred
   - what type of change occurred (create, update, delete)
   - which fields changed during an update
 - a Django signal exists to create a `RevEngineHistoricalChange` object when `django-simple-history` tries to create a historical record

 - go to a model’s detail page in the Django admin (for example, at the /nrhadmin/organizations/organization/1/change/ URL)
 - make a change (for example, edit the ‘Salesforce ID’), save the object, and then click the “HISTORY” button in the top right corner of the object’s detail page in the Django admin
 - observe a list of all of the changes to the model, including the change made, the datetime for each change, the type of change made, and the user who made the change
 - go to the list of all tracked changes in the Django admin (at the /nrhadmin/common/revenginehistoricalchange/ URL)
 - observe the change you just made in the list view
 - make any other changes (in the Django admin or in the app, such as to a donation page), and observe them in the list view at the /nrhadmin/common/revenginehistoricalchange/ URL

No


michael-caktus (commented):
This works great, wonderful. I've got just one question-- I noticed when running this locally that I have to make two changes to a model for it to track a change. I assume this is because we need a history record to run a diff against. Is there any way around this?

For instance, can we create a data migration as part of this PR that essentially "touches" each record? Or just creates a history record of the model as-is at time of data migration?

Let me know. If this is terribly difficult, we could possibly just move ahead. It may not be a deal breaker that this "two changes required, aka first change is lost" exists.

dchukhin:
Thanks for the review.
You are correct: a diff can only be shown if a previous edit exists, so the first edit for any model will be blank. I don't think it will be a large amount of work to "touch" each object in relevant models in a migration. I can work on adding that today.

PR #321, opened 2022-02-14
[#11930] Associate styles with RPs not orgs

michael-caktus:
#### What's this PR do?
Removes the direct association between Styles and Organizations. Stylesheets are now associated to RevenuePrograms.
The bulk of the work here was these two things:
1. Writing and testing a data migration for existing Styles.
2. Updating the Org portal AddStylesModal to include a RevenueProgram picker.

The path we chose for managing existing Style -- DonationPage -- RevenueProgram relationships was as follows:
- For Styles used across multiple RevenuePrograms, pick one RevenueProgram (and all its DonationPages) to use that style. For the rest, create a duplicate Style (with "[duplicate]" in the name)
- For Styles used by only one RevenueProgram, associate the Style with the RevenueProgram for the Page(s) that use that Styles.
- For Styles not used by any RevenuePrograms, just pick the "first" RevenueProgram in its Organization.

Lots to test here, but hopefully the above gives a reasonable set of test cases. The big picture is what's important here-- do all your pages still look the way you expect? Are there any Styles in "impossible" situations where they're attached to a page that doesn't belong to same RevenueProgram.  Also test the new RevenueProgram picker in the AddStylesModal

No


kmtracey (commented):
FWIW LGTM -- I don't think when we talked I picked up on the 3rd case of styles not in use at all. It seems to me the migration covers all the bases, assuming an organization must have at least one revenue program.

JeremyGibson:
I verified 3 duplicates against their parents and they all looked good.  Presuming this test passes :shipit:

PR #323, opened 2022-02-17
[#12066] restores org slug

michael-caktus:
#### What's this PR do?
Restores Organization.slug

Creates default slug for orgs based on org name (required due to database schema)

Verify that there's an org slug field


LorienOlive (approved):
Looks good.  Appearing in django admin organization as expected.

PR #328, opened 2022-02-23
Style and HistoricalStyle migration conflict

michael-caktus:
#### What's this PR do?
A bunch of concurrent work in a single area, confounded by the particular implementation of HistoricalRecord, resulted in some migration confusion. A migration that created HistoricalRecord for Style created it based on a schema that was invalidated by a migration with the same dependency. This resulted in the need for a merge migration, but for some reason it never got created and the migration history moved on. This PR rewrites that history in order to correct that initial issue.

If the migrations run and the app doesn't crash when accessing Styles or HistoricalRecords, we're good.

We're going to need to manually migration Staging back to pages.0039 before this is deployed in order for this change to take effect. Some minor staging data will likely be lost. In particular, historical records created in the last few days. No effect on production, since these migrations haven't made it out to prod yet.


apps/organizations/serializers.py
```diff
@@ -75,6 +75,7 @@ class Meta:
             "id",
             "name",
             "slug",
+            "benefit_levels",
```
JeremyGibson, line 78: Want to remove this in favor of what is on my PR? Doesn't really hurt anything though.

JeremyGibson (approved):
Looks good. Just the one comment about the unrelated change from the branch parent.

PR #336, opened 2022-03-11
Dev 1092 pay fee toggle

gurjittoptal:
#### What's this PR do?
The pay toggle section is moved out of stripe-section thats hidden.
ref
https://news-revenue-hub.atlassian.net/browse/DEV-1092

Check on safari(iPhone) and Chrome(Android Phone)

No


x110dc (changes requested):
- I think we'd like the "pay fees" toggle to remain before the submit button. 
- We also need to confirm that it is usable and visible regardless of the type of button. It should be fairly easy to test with Google Pay and Browser-Saved cards, but Apple Pay is tougher because it won't show outside of production.

gurjittoptal:
This was updated. Toggle is duplicated in the two sections. ( single line change now )

benjaminEwhite (approved):
Code looks good to me 👍

x110dc (approved):

PR #337, opened 2022-03-11
DEV-995: use NODE_ENV instead of domain

harsha09:
#### What's this PR do?
As of now React is configured to tag hostname as environment in Sentry, this PR changes the environment tag to actual ENVIRONMENT from env variables like `{production, staging, test, dev, demo}`

Raising an exception / warning from an UI, sentry log will be generated with ENVIRONMENT `{production, staging, test, dev, demo}` as a tag instead of hostname.

Need to add an env variable
ENVIRONMENT `{production, staging, test, dev, demo}`


x110dc (changes requested):

x110dc (approved):
LGTM.

spa/src/settings.js
```diff
@@ -52,6 +52,9 @@ export const STRIPE_API_VERSION = resolveConstantFromEnv('STRIPE_API_VERSION', '
 export const STRIPE_CLIENT_ID = resolveConstantFromEnv('STRIPE_CLIENT_ID', 'test_1234');
 export const STRIPE_OAUTH_SCOPE = resolveConstantFromEnv('STRIPE_OAUTH_SCOPE', 'read_write');
 
+// Environment {production, staging, test, dev, demo}
+export const ENVIRONMENT = resolveConstantFromEnv('ENVIRONMENT');
```
benjaminEwhite, line 56: Should this also pass in a default value for second argument? That's a genuine question -- it's fine if the answer is no, I just saw that `resolveConstantFromEnv` can take a default value.

benjaminEwhite (approved):
This looks good to me.

spa/src/settings.js
harsha09: I've removed the default value here as it get's the default value from Django's `ENVIRONMENT` env variable, which makes sense to me.

PR #341, opened 2022-03-14
DEV-1393 [upper|lower]_limit API fields

njharman:
[DEV-1393]

#### What's this PR do?
Add fields for lower_limit and upper_limit to /api/v1/pages/revenue-programs/<detail> "benefit_levels" 

Separate fields return unformatted numbers. Believe presentation should
be up to / handled by front ends. Argument could be made that "dollars"
is a unit and not presentation. Still would have that as separate field
and keep limits as "raw" numbers.

donation_range also exposed via unused /api/pages, limit fields not
exposed there.

To get pre-commit to pass:

  - Update flake8 pre-commit to 4.0.1 (previous conflict with black)

run automated tests.

poke /api/v1/public/revenue-programs/1 and see the fields (assumes appropriate db records)

No.

[DEV-1393]: https://news-revenue-hub.atlassian.net/browse/DEV-1393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


x110dc (approved):
I created a plan, org, revenue program and level in the review app and checked the output. Looks as expected. LGTM.

harsha09 (approved):
LGTM

PR #345, opened 2022-03-15
[DEV-699] Introduce roles and use them to programmatically manage access to orgs, rev programs, and other data models

benjaminEwhite:

NA

#### What's this PR do?

- Introduces role-based access for the primary resources exposed through the REST API.
- Introduces three new user roles: 
  - Hub admin - prototypically can CRUD everything
  - Org admin - prototypically can CRUD only things belonging to their org
  - Revenue program admin - prototypically can CRUD only things belonging to their RPs
- Allows role assignments to be managed in Django admin. Role assignments connect a user, a role type, an organization, and a set of revenue programs.
- Adds custom testing mixins that provide extensive domain model data setup and convenient test assertions to standardize testing the API in light of permissions
- Extensively tests CRUD operations for each role type, plus superusers (and in some places, authed users and contributor users) for most API resources.

Documented extensively in the related JIRA ticket

In test environment, go to Django admin. Assuming it's already populated with org/rp/page etc. data, create different users with different role types, and interact with the API via postman, authenticating as each distinct user.

In theory, this can also be tested by logging in to the SPA as, say, a revenue program admin and looking at the contributions list. You should only see contributions listed for your rps.  This pattern is true for any of the resources listed in the SPA admin.

- Adds custom testing mixins that provide extensive domain model data setup and convenient test assertions to standardize testing the API in light of permissions
- Extensively tests CRUD operations for each role type, plus superusers (and in some places, authed users and contributor users) for most API resources.
- Adds some tests in otherwise untouched code in order to get coverage > 95%

NA

NA

[DEV-699]


apps/pages/serializers.py
```diff
@@ -57,13 +58,11 @@ def create(self, validated_data):
...
 
     def set_revenue_program(self, validated_data):
-        rp_data = validated_data.get("revenue_program")
-        if not rp_data:
-            raise serializers.ValidationError(
-                {"revenue_program": "revenue_program is required when creating a new Style"}
-            )
+        _, rp_slug = get_org_and_rp_from_request(self.context["request"])
+        if not rp_slug:
+            raise serializers.ValidationError({"rp_slug": "RevenueProgram.slug is required when creating a new page"})
```
njharman, line 63: "rp_slug" -> f"{settings.RP_SLUG_PARAM}" . 
Maybe same with the RP not found below

unless you meant f"{rp_slug}" ???  Which I like even less.

apps/pages/serializers.py
```diff
@@ -82,7 +81,7 @@ class DonationPageFullDetailSerializer(serializers.ModelSerializer):
     styles_pk = serializers.IntegerField(allow_null=True, required=False)
 
     revenue_program = RevenueProgramListInlineSerializer(read_only=True)
-    revenue_program_pk = serializers.IntegerField(allow_null=True, required=False)
+    # revenue_program_pk = serializers.IntegerField(allow_null=True, required=False)
```
njharman, line 84: add comment why commented out or delete.

apps/pages/serializers.py
```diff
@@ -158,13 +157,11 @@ def _create_from_template(self, validated_data):
...
+        _, rp_slug = get_org_and_rp_from_request(self.context["request"])
+        if not rp_slug:
+            raise serializers.ValidationError({"rp_slug": "RevenueProgram.slug is required when creating a new page"})
         try:
-            rev_program_pk = validated_data.pop("revenue_program_pk")
-            rev_program = RevenueProgram.objects.get(pk=rev_program_pk)
+            rev_program = RevenueProgram.objects.get(slug=rp_slug)
             validated_data["revenue_program"] = rev_program
         except RevenueProgram.DoesNotExist:
             raise serializers.ValidationError({"revenue_program_pk": "Could not find revenue program with provided pk"})
```
njharman, line 167: same as above re: rp_slug, revenue_program_pk

apps/public/permissions.py
```diff
@@ -3,4 +3,4 @@
 
 class IsSuperUser(permissions.BasePermission):
     def has_permission(self, request, *args):
-        return request.user.is_active and request.user.is_superuser
+        return hasattr(request.user, "is_active") and request.user.is_active and request.user.is_superuser
```
njharman, line 6: minor, the test if obj has attr and that attr is truthy can be simplifed to getattr with default

`getattr(obj, attar, False)`

apps/users/forms.py
```diff
@@ -0,0 +1,62 @@
...
+
+            # Filter for any revenue programs that do not belong to this Org
+            mismatched_rps = self.cleaned_data["revenue_programs"].filter(
+                ~Q(organization=self.cleaned_data["organization"])
+            )
+
+            if mismatched_rps.exists():
+                error_dict[
+                    "revenue_programs"
+                ] = f"The following RevenuePrograms do not belong to your chosen Org: {', '.join([rp.name for rp in mismatched_rps])}"
```
njharman, line 49: Is there security/privacy concern showing non-owned RP names?

apps/users/forms.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if error_dict:
+            raise ValidationError(error_dict)
+
+    class Meta:
+        model = RoleAssignment
+        fields = "__all__"
+
+    # class Media:
```
njharman, line 58: Add comment why commented out or delete.

apps/users/management/commands/bootstrap-review-app.py
```diff
@@ -1,8 +1,16 @@
...
 
 class Command(BaseCommand):  # pragma: no cover
     help = "Bootstrap Heroku review app"
 
     def handle(self, *args, **options):
-        self.stdout.write(self.style.SUCCESS("Bootstrapped review app"))
+        User = get_user_model()
+        qatester_exists = User.objects.filter(email="qatester@example.com").exists()
+        if settings.ENVIRONMENT != "review" or qatester_exists:
+            self.stdout.write("**Not running bootstrap tasks**")
```
njharman, line 13: Helpful to user running to explain why,  either not "review env" or "qatester exits.

apps/users/models.py
```diff
@@ -53,3 +45,25 @@ class OrganizationUser(models.Model):
...
+
+class Roles(models.TextChoices):
+    HUB_ADMIN = "hub_admin", "Hub Admin"
+    ORG_ADMIN = "org_admin", "Org Admin"
+    RP_ADMIN = "rp_admin", "RP Admin"
+
+
+class RoleAssignment(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    role_type = models.CharField(max_length=max(len(x) for x in Roles.values), choices=Roles.choices)
```
njharman, line 58: Rusty on django, help me out. I'm queasy about dynamically generated model field parameter

Will the db migration generator tool (forget name, South?) detect and create migration to change size if we add a longer role?

[Also with Postgres having tight max_length doesn't get you anything (re storage) and I think it silently truncates.  you do get validation on the django side. I guess I'm asking why not just max_length 20 or 40]

apps/users/serializers.py
```diff
@@ -2,20 +2,54 @@
...
+from apps.users.models import Roles
 
 
 class UserSerializer(serializers.ModelSerializer):
-    organization = serializers.SerializerMethodField()
+    """
+    This is the serializer that is used to return user data back after successful login.
+    It returns a complete list of (pared-down) available Organzitions and RevenuePrograms based on the user's
+    super_user status and RoleAssignment.
+    """
```
njharman, line 18: Thank you!

revengine/settings/base.py
```diff
@@ -224,12 +228,44 @@
...
+            "class": "django.utils.log.AdminEmailHandler",
+        },
+        "warn_admins": {
+            "level": "WARNING",
+            "filters": ["require_debug_false"],
+            "class": "django.utils.log.AdminEmailHandler",
+        },
         "console": {
-            "level": "INFO",
+            "level": "DEBUG",
```
njharman, line 242: Was this intentential commit to change debug level, or forgotten local edit?

njharman (changes requested):
Various in line comments and 

Coverage
FAIL Required test coverage of 95% not reached. Total coverage: 93.87%

CI failures

-   build_test_python The hosted runner: GitHub Actions 10 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
 -  js_test  Same error

Big commit/unfamiliar/etc... 
I did not see user with no roles is denied all these things (which I also didn't see as acceptance criteria). But seems that case handled by this, eh? But also see some endpoints set classes to [] which overrides this default, right?

  "DEFAULT_PERMISSION_CLASSES": [
          "apps.api.permissions.HasRoleAssignment",

apps/contributions/views.py
```diff
@@ -103,9 +116,16 @@ def stripe_oauth(request):
 
 
 @api_view(["POST"])
+@permission_classes([IsAuthenticated, HasRoleAssignment | IsSuperUser])
 def stripe_confirmation(request):
+    organization_slug = request.GET.get(settings.ORG_SLUG_PARAM, None)
```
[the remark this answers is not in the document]
benjaminEwhite: I don't disagree, but I'm reluctant to start pulling threads like that on this PR because it's already big, and that's additional scope creep.

apps/users/forms.py
benjaminEwhite: I believe these would be submitted by the user (they come from the form submission), and not generated by the server, so there shouldn't be a security concern.

apps/users/models.py
benjaminEwhite: I believe the Django's migration capabilities would catch the change, but I agree, @njharman I'm not a fan of the dynamic generation of the role type. I'm going to update.

revengine/settings/base.py
benjaminEwhite: Good question. I'm guessing it's forgotten local edit (from previous dev), and at any rate it's not clearly tied to narrow scope of this PR, so will revert this change.

apps/organizations/views.py
```diff
@@ -3,78 +3,47 @@
...
-            return self.model.objects.filter(organization__users=self.request.user)
-        return self.model.objects.all()
-
-
-class OrganizationViewSet(OrganizationLimitedListView, viewsets.ReadOnlyModelViewSet):
+class OrganizationViewSet(viewsets.ReadOnlyModelViewSet):
     model = Organization
-    permission_classes = [IsAuthenticated, UserBelongsToOrg, ReadOnly]
+    queryset = Organization.objects.all()
+    permission_classes = [ReadOnly]
```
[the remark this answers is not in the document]
benjaminEwhite: Good to have pause on this. I'm going to ask about the business logic for this, because I'm not entirely certain.

apps/pages/models.py
```diff
@@ -99,7 +99,7 @@ def make_page_from_template(self, page_data={}):
         unwanted_keys = ["_state", "id", "modified", "created", "published_date", "_history_user"]
         template = cleanup_keys(template_data, unwanted_keys)
         page = cleanup_keys(page_data, unwanted_keys)
-        merged_page = template | page
+        merged_page = template if template else page
```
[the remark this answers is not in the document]
benjaminEwhite: Ohhhhh, yeah that makes a lot more sense. It also explains the failing test i was getting which prompted me to make this change. I was getting a test failing where creating a page from template was causing the name field to not be set correctly, IIRC. 

Personally, I find the starkwargs syntax less ambiguous, and it also resembles object-destructuring syntax in Javascript, so for full stack folks, it might make it a bit more intuitive sense. 

At the same time, I simply wasn't aware of this usage of the `|` operator which appears to have come with Python 3.9. I'm open to using it. As you can see, my intuitive read of `|` here as "or" was wrong, so that's an interesting small fact.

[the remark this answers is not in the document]
benjaminEwhite: Oh wait. According to pyproject.toml, we're targeting 3.8 currently, so the `|` won't work here. Indeed, I get failing tests locally because of this.

There's another example of this in make_template_from_page that I need to take a look at.

benjaminEwhite:
@njharman I responded/made commits based on your inline comments.

Re, coverage: I mentioned in the opening comments for this PR where there were some gaps in testing (and that was without realizing we weren't hitting the coverage min threshold we have set). Running `coverage -m`, it looks like `apps/api/permissions` only has 76% coverage. I think the best way to up that coverage is through API testing of resources that use the permissioning system -- which would mean additional API integration tests for some of the viewsets with using different RoleAssignments.


Re: CI failures, I'll go have a look. I know that there were some issues with the GitHub API yesterday, so that may account for what you were seeing.

RE: user with no roles. There are default permissions set in revengine/settings/base.py for IsAuthenticated and HasRoleAssignment.  You rightly point out that some views override/reset default permissions. I've tried to go through and look at each view and ask "Do the permissions on this view make sense based on my understanding of the app?". In a few places I tightened permissions based on this.

All of that said, it's entirely possible I missed something. If any specific resources need to be more tightly secured, we can do that.

apps/users/management/commands/bootstrap-review-app.py
```diff
@@ -1,8 +1,20 @@
...
+from django.contrib.auth import get_user_model  # pragma: no cover
 from django.core.management.base import BaseCommand  # pragma: no cover
 
 
 class Command(BaseCommand):  # pragma: no cover
     help = "Bootstrap Heroku review app"
 
     def handle(self, *args, **options):
-        self.stdout.write(self.style.SUCCESS("Bootstrapped review app"))
+        User = get_user_model()
```
x110dc, line 10: Something's gone wrong here as I removed (most of) this code in this PR: https://github.com/newsrevenuehub/rev-engine/pull/343

revengine/settings/base.py
```diff
@@ -224,12 +228,44 @@
     "filters": {"require_debug_false": {"()": "django.utils.log.RequireDebugFalse"}},
     "formatters": {"basic": {"format": "%(asctime)s %(name)-20s %(levelname)-8s %(message)s"}},
     "handlers": {
+        "mail_admins": {
+            "level": "ERROR",
+            "filters": ["require_debug_false"],
+            "class": "django.utils.log.AdminEmailHandler",
+        },
```
x110dc, line 235: And again here, I intentionally removed all of this in: https://github.com/newsrevenuehub/rev-engine/pull/342

It shouldn't be added back.

revengine/settings/base.py
```diff
@@ -286,7 +322,7 @@
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 EMAIL_SUBJECT_PREFIX = "[RevEngine] "
 
-ADMINS = [("dc", "daniel@fundjournalism.org")]
+ADMINS = [("nrh-team", "nrh-team@caktusgroup.com")]
```
x110dc, line 325: And here: https://github.com/newsrevenuehub/rev-engine/pull/342

x110dc (changes requested):
The PR section titled "How should this be manually tested?" is blank. Can you fill that out please?

apps/users/choices.py
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+
+
+class Roles(models.TextChoices):
```
benjaminEwhite, line 4: FWIW, this was moved here and out of models.py because of a circular import.

apps/pages/validators.py
```diff
@@ -1,13 +1,47 @@
...
+        target_page = self.page_model.objects.get(pk=value["page_pk"])
+        if (request := serializer.context.get("request", None)) is not None:
+            user = request.user
+            ra = getattr(user, "roleassignment", None)
+            if user.is_superuser or (ra is not None and ra.role_type == Roles.HUB_ADMIN):
+                return
+            elif ra.role_type == Roles.ORG_ADMIN and target_page.revenue_program.organization != ra.organization:
+                raise serializers.ValidationError(UNOWNED_TEMPLATE_FROM_PAGE_PAGE_PK_MESSAGE)
+            elif ra.role_type == Roles.RP_ADMIN and target_page.revenue_program not in ra.revenue_programs.all():
+                raise serializers.ValidationError(UNOWNED_TEMPLATE_FROM_PAGE_PAGE_PK_MESSAGE)
```
njharman, line 47: Believe ra could be None and elif ra.role_type clauses reached. Maybe change to 

```
If user.is_super_user:
  return
if ra is None
  raise?
if ra.role_type == HUB
   return
elif ...
```

apps/pages/views.py
```diff
@@ -120,16 +153,27 @@ def get_serializer_class(self):
...
+
+    def get_queryset(self):
+        return self.filter_queryset_for_user(self.request.user, self.model.objects.all())
 
 
 class FontViewSet(viewsets.ReadOnlyModelViewSet):
+    model = Font
     queryset = Font.objects.all()
+    # anyone who is authenticated read
     permission_classes = [IsAuthenticated]
```
njharman, line 177: black don't force trailing comma in list :(

apps/public/permissions.py
```diff
@@ -3,4 +3,4 @@
 
 class IsSuperUser(permissions.BasePermission):
     def has_permission(self, request, *args):
-        return request.user.is_active and request.user.is_superuser
+        return getattr(request.user, "is_active", False) and request.user.is_superuser
```
njharman, line 6: What am I missing. how can user not have is_active?  Is it because there's a fake user for contributors now? If so that fake user should implement is_active instead of relying on remembering to do this everywhere.

If it's because request.user can be None then `request.user and request.user.is_active` is the way. Short circut boolean logic ftw!

apps/users/models.py
```diff
@@ -53,3 +47,63 @@ class OrganizationUser(models.Model):
...
+
+
+    API resources use role assignnments to do things like limit queryset results to only objects owned by
+    a user's organization (for instance).
+    """
+
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    role_type = models.CharField(max_length=50, choices=Roles.choices)
+    organization = models.ForeignKey("organizations.Organization", null=True, blank=True, on_delete=models.SET_NULL)
+    revenue_programs = models.ManyToManyField("organizations.RevenueProgram", blank=True)
```
njharman, line 63: Can't remember how/if django supports constraint to limit to only rps that belong to organization. It would be good.  Even if only as a todo note to implement in future.

apps/users/serializers.py
```diff
@@ -2,20 +2,58 @@
...
+            return ("superuser", "Superuser")
+        role_assignment = obj.get_role_assignment()
+        return (role_assignment.role_type, role_assignment.get_role_type_display()) if role_assignment else None
+
+    def get_permitted_organizations(self, obj):
+        qs = Organization.objects.all()
+        role_assignment = obj.get_role_assignment()
+        if not role_assignment and not obj.is_superuser:
+            qs = qs.none()
+        elif role_assignment and role_assignment.role_type != Roles.HUB_ADMIN:
```
njharman, line 35: Other places you raise on unrecognized Role.  should this also not be "future proofed" by inclusivity rather than != ?

ra.rt in (Roles.ORG_ADM, Roles.RP_ADM)

apps/users/serializers.py
```diff
@@ -2,20 +2,58 @@
...
+    def get_role_type(self, obj):
+        if obj.is_superuser:
+            return ("superuser", "Superuser")
+        role_assignment = obj.get_role_assignment()
+        return (role_assignment.role_type, role_assignment.get_role_type_display()) if role_assignment else None
+
+    def get_permitted_organizations(self, obj):
+        qs = Organization.objects.all()
+        role_assignment = obj.get_role_assignment()
+        if not role_assignment and not obj.is_superuser:
```
njharman, line 33: Do super users never have role_assignment? If they can have role, they'll get limited by it here.

apps/users/serializers.py
```diff
@@ -2,20 +2,58 @@
...
+                qs = qs.none()
+            else:
+                qs = qs.filter(pk=org.pk)
+        serializer = OrganizationInlineSerializer(qs, many=True)
+        return serializer.data
+
+    def get_permitted_revenue_programs(self, obj):
+        qs = RevenueProgram.objects.all()
+        role_assignment = obj.get_role_assignment()
+        if not role_assignment and not obj.is_superuser:
```
njharman, line 47: Do super users never have role_assignment? If they can have role, they'll get limited by it here.

apps/users/serializers.py
```diff
@@ -2,20 +2,58 @@
...
+            qs = qs.none()
+        elif not obj.is_superuser and role_assignment.role_type != Roles.HUB_ADMIN:
+            if role_assignment.role_type == Roles.ORG_ADMIN:
+                qs = role_assignment.organization.revenueprogram_set.all()
+            elif role_assignment.role_type == Roles.RP_ADMIN:
+                qs = role_assignment.revenue_programs
```
njharman, line 53:
```
else:
   qs = qs.none()
```
to guard against new roles added in future seeing all.

apps/users/views.py
```diff
@@ -62,3 +76,45 @@ def retrieve_user(request):
...
+    @classmethod
+    def filter_queryset_for_user(cls, user, queryset):
+        try:
+            if is_a_contributor(user):
+                return queryset.model.filter_queryset_for_contributor(user, queryset)
+            if user.is_anonymous:
+                return queryset.none()
+            if user.is_superuser:
+                return queryset.all()
+            elif role_assignment := getattr(user, "roleassignment"):
```
njharman, line 97: weird that this is only elif

apps/users/views.py
```diff
@@ -62,3 +76,45 @@ def retrieve_user(request):
...
+                return queryset.model.filter_queryset_by_role_assignment(role_assignment, queryset)
```
njharman, line 98: final "if none of the above" return queryset.none() or raise some error?

revengine/settings/base.py
```diff
@@ -316,10 +319,10 @@
 
 
 # Application subdomains (that are NOT revenue program slugs)
-NON_DONATION_PAGE_SUBDOMAINS = os.getenv("NON_DONATION_PAGE_SUBDOMAINS", "support:www").split(":")
-ORG_PORTAL_SUBDOMAINS = os.getenv("ORG_PORTAL_SUBDOMAINS", "support:").split(":")
+NON_DONATION_PAGE_SUBDOMAINS = os.getenv("NON_DONATION_PAGE_SUBDOMAINS", ["support", "www"])
```
njharman, line 322: Yes! That split weirds me out. "Too lazy to type list punctuation, get Python to do it for me"?

njharman (changes requested):
This commit is too big

In line comments on python code. Didn't look at tests or js.

PR #348, opened 2022-03-21
Dev 1411 linelength

njharman:
#### What's this PR do?
Set black (and isort) line length to 140.  Did not reformat any of the .py files.  ~Reformatting will be required done as other commits make changes to files. Or, can be done in one fell swoop between sprints?~ Except the repo's workflow runs pre-commit with --all-files so it fails. 

Consistancy across repos.

N/A

Already exists pre-commit and CI

N/A

N/A

should be https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2047311873/Python+Style+Guide

[DEV-1411]

No

Part of larger (and existing tickets) to unify and improve code quality stuffs. 

[DEV-1411]: https://news-revenue-hub.atlassian.net/browse/DEV-1411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


njharman:
I was hoping this would not require reformatting all the code. But the way pre-commit is run in github workflow action means it will always fail until we do.

Not sure but maybe this should be reviewed, tentatively approved and held until good time (between sprints?) to do all the black reformatting and merge.

pyproject.toml
```diff
@@ -75,23 +72,19 @@ django-simple-history = "^3.0.0"
 [tool.poetry.dev-dependencies]
 appnope = "0.1.2"
 beautifulsoup4 = "4.9.3"
-black = "20.8b1"
+black = "^20.8b1"
```
x110dc, line 75: This is lower than the version here: https://github.com/newsrevenuehub/rev-engine/blob/develop/.pre-commit-config.yaml#L3

Does that matter? I guess the `^` is designating a minimum version right? So it's fine that these are different?

x110dc (changes requested):
Is there a way to make it not do `--all-files` for now and we could add that back later? Otherwise seems like yeah, need to do at beginning/end of sprint when no other outstanding PRs in flight.

njharman:
removing --all-files, It won't process any files then. pre-commit looks at what is about to be committed. After commit, in CI, it's all committed, so it sees no files that need checking. Hence action uses --all-files.

I googled a bit and there was ways with other actions to get list of files that had been changed. It all looked more trouble than worth.

pyproject.toml
njharman: yes the circumflex(pointy hat) is minimum.  I was creeping scope here. These dev qa things should be "*" like in webhook-receiver.

PR #401, opened 2022-04-28
DEV-1578 support distinct role types in the spa

benjaminEwhite:

#### What's this PR do?

Changes to SPA (and recursive changes to API that emerged) to enable role based permissions.

One consequence of this PR is that page, template, style, and contribution data will be filtered via user roles.

Some noteworthy code changes:

*Backend*

- Simplifies API endpoint configuration for creation, such that there are not query parameters being passed to endpoints that otherwise expect a body of data (aka, no longer faceting based on RP and ORG slugs)
- Adds `apps.common.fields.SerializedOnReadElsePk` field which can be used to enable an API resource where a field is PK on creation, otherwise serialized
- Cleans up some use of custom mixins that were not rquired in views
- Adds some fields to OrganizationInlineSerializer to make front end work
- Adds a validator (`RpPkIsForOwnedRp`) which can be used to validate if the requesting user should be able to reference a related revenue program

*SPA*

- Adds a `forceLogin` command that can be used to directly force login without going through sign on page flow in tests, and replaces a bunch of current use of cy.login with this new method.
- Removes unused revenue program route
- Renames `OrganizationContext` to `PaymentProviderContext`
- In `AddPageModal`, the list of revenue programs that appear in the dropdown now comes from the list of available RPs in a user's role information instead of pulled from API resource.
- Removes unused src/components/organization/Organization.js

will allow us to ship role-based permissions

Run the dev server locally (preferably after retrieving a recent back up from staging and restoring it, then running migrations). Log in to the SPA as a hub admin or super user and you should see all content. Log in as an org or rp user, and you should only see content that you own.

Added and edited.

[DEV-1578]

No

Revisit DEV-699 once this is merged into it, and do clean up before submitting that PR for final review.


benjaminEwhite:
FYI, the Python tests are failing at present because of a migration issue. I'm unable to reproduce this locally and will run down separately. The server-side tests all pass locally, so this need not block reviewing.

x110dc (approved):
LGTM.

PR #407, opened 2022-05-05
Dev 1578 and dev 699 (for merge to main)

x110dc:

#### What's this PR do?

Changes to SPA (and recursive changes to API that emerged) to enable role based permissions.

One consequence of this PR is that page, template, style, and contribution data will be filtered via user roles.

Some noteworthy code changes:

*Backend*

- Simplifies API endpoint configuration for creation, such that there are not query parameters being passed to endpoints that otherwise expect a body of data (aka, no longer faceting based on RP and ORG slugs)
- Adds `apps.common.fields.SerializedOnReadElsePk` field which can be used to enable an API resource where a field is PK on creation, otherwise serialized
- Cleans up some use of custom mixins that were not rquired in views
- Adds some fields to OrganizationInlineSerializer to make front end work
- Adds a validator (`RpPkIsForOwnedRp`) which can be used to validate if the requesting user should be able to reference a related revenue program

*SPA*

- Adds a `forceLogin` command that can be used to directly force login without going through sign on page flow in tests, and replaces a bunch of current use of cy.login with this new method.
- Removes unused revenue program route
- Renames `OrganizationContext` to `PaymentProviderContext`
- In `AddPageModal`, the list of revenue programs that appear in the dropdown now comes from the list of available RPs in a user's role information instead of pulled from API resource.
- Removes unused src/components/organization/Organization.js

will allow us to ship role-based permissions

Run the dev server locally (preferably after retrieving a recent back up from staging and restoring it, then running migrations). Log in to the SPA as a hub admin or super user and you should see all content. Log in as an org or rp user, and you should only see content that you own.

Added and edited.

[DEV-1578]

No

Revisit DEV-699 once this is merged into it, and do clean up before submitting that PR for final review.


njharman (changes requested):
Doing review, but right now there's
1. Test coverage under 95% (CI run fail)
2. Merge conflicts.

apps/pages/validators.py
```diff
@@ -1,13 +1,89 @@
...
+            if user.is_superuser or (ra is not None and ra.role_type == Roles.HUB_ADMIN):
+                return
+            elif ra.role_type == Roles.ORG_ADMIN and target_page.revenue_program.organization != ra.organization:
+                raise serializers.ValidationError(UNOWNED_TEMPLATE_FROM_PAGE_PAGE_PK_MESSAGE)
+            elif ra.role_type == Roles.RP_ADMIN and target_page.revenue_program not in ra.revenue_programs.all():
+                raise serializers.ValidationError(UNOWNED_TEMPLATE_FROM_PAGE_PAGE_PK_MESSAGE)
+
+
+UNOWNED_REVENUE_PROGRAM_PK_MESSAGE = "You aren't permitted to reference this revenue program"
+MISSING_REFRENCE_TO_REV_PROGRAM_MESSAGE = "Can't determine relationship with revenue program"
```
njharman, line 51: Why do we do this, instead of subclassing ValidationError, with the text.  I really dislike GLOBALS. I'm mostly ranting. At least these are marked _MESSAGE and so can be ignored.

apps/pages/validators.py
```diff
@@ -1,13 +1,89 @@
...
+            if (request := serializer.context.get("request", None)) is not None:
+                user = request.user
+                ra = getattr(user, "roleassignment", None)
+                if user.is_superuser or (ra is not None and ra.role_type == Roles.HUB_ADMIN):
+                    return
+                elif ra.role_type == Roles.ORG_ADMIN and target_rp.organization != ra.organization:
+                    raise serializers.ValidationError(UNOWNED_REVENUE_PROGRAM_PK_MESSAGE)
+                elif ra.role_type == Roles.RP_ADMIN and target_rp not in ra.revenue_programs.all():
+                    raise serializers.ValidationError(UNOWNED_REVENUE_PROGRAM_PK_MESSAGE)
+        return
```
njharman, line 89: Bare return

apps/public/permissions.py
```diff
@@ -3,4 +3,4 @@
 
 class IsSuperUser(permissions.BasePermission):
     def has_permission(self, request, *args):
-        return request.user.is_active and request.user.is_superuser
+        return getattr(request.user, "is_active", False) and request.user.is_superuser
```
njharman, line 6: getattr, isn't suppose to look for existence of is_superuser?

Currently, if request.user lacks "is_active" they are not a superuser regardless of what user.is_superuser says.
and will raise AttributeError, if is_active exists and is truish and is_superuser does not exist.

apps/users/models.py
```diff
@@ -53,3 +47,58 @@ class OrganizationUser(models.Model):
...
+    organization = models.ForeignKey("organizations.Organization", null=True, blank=True, on_delete=models.SET_NULL)
+    revenue_programs = models.ManyToManyField("organizations.RevenueProgram", blank=True)
+
+    def __str__(self):
+        if self.role_type == Roles.HUB_ADMIN:
+            return Roles.HUB_ADMIN.label
+        if self.role_type == Roles.ORG_ADMIN:
+            return f"{Roles.ORG_ADMIN.label} for {self.organization.name}"
+        if self.role_type == Roles.RP_ADMIN:
+            return f"{Roles.RP_ADMIN.label} for {self.organization.name}"
```
njharman, line 71: Not listing RPs they're admin for?

apps/users/models.py
```diff
@@ -53,3 +47,58 @@ class OrganizationUser(models.Model):
...
+        return f"Unspecified RoleAssignment ({self.pk})"
```
njharman, line 72: Unknown role_type, eh?  unspecified would be org null and/or rp null, depending, for ORG or RP admins.

apps/users/models.py
```diff
@@ -53,3 +47,58 @@ class OrganizationUser(models.Model):
...
+
+
+    API resources use role assignnments to do things like limit queryset results to only objects owned by
+    a user's organization (for instance).
+    """
+
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    role_type = models.CharField(max_length=50, choices=Roles.choices)
+    organization = models.ForeignKey("organizations.Organization", null=True, blank=True, on_delete=models.SET_NULL)
+    revenue_programs = models.ManyToManyField("organizations.RevenueProgram", blank=True)
```
njharman, line 63: Could add constraints to guarantee ORG/RP hav org defined and RP have rps defined. and NULL otherwise.

njharman (changes requested):
Did not review javascript/SPA as I'm unqualified.
Did not manually test.
Only glanced at tests (very large commit to also pour over these and test coverage is below 95%)

A few questions/clarifications, a few minor changes.

apps/users/models.py
benjaminEwhite: That business logic appears to be enforced in `RoleAssignmentAdminForm.clean`, which is the only interface for creating role assignments, so I think we're okay.

njharman (approved):
assumming cypress test will finish and pass eventually.
lgtm

x110dc:
I pushed a couple commits to fix what looks like a bad merge (probably from when I created the dual-ticket branch). Approving.

PR #409, opened 2022-05-06
Dev 1056 deliver initial confirmation emails for recurring donations mvp

benjaminEwhite:

This is complete in terms of being ready for review, but do need to upload the new email template (/email_templates/nrh-default-contribution-confirmation-email.html) to Mailgun and confirm everything works as expected.

#### What's this PR do?

Causes contribution confirmation emails to be sent to contributors making recurring donations (vs. status quo where only contributors making one time donations get a confirmation email).

Along the way, a number of notable changes were made (some driven by the need to boost test coverage, which required touching otherwise untouched code).

- Retires apps.emails.models.PageTemplateEmail and related admin, signals, and management command files
- Enables eager celery task execution in tests
- Adds `ESP_TEMPLATE_ID_FOR_CONTRIBUTION_CONFIRMATION` to settings/base.py. This value is pulled from environment, so will need to be set up in Heroku when deploy. We'll get this value when we upload the template to Mailgun.
- Renames `send_donor_email` task to `send_email` task, since nothing about implementation was tied to contributors.
- Adds a new task `send_contribution_confirmation_email` which is used to trigger contribution confirmation emails. This allows us in short term to use a single email template to handle both recurring and one time contributions.
- Refactors the view tests around one-time and recurring payments to assert that send_contribution_confirmation_email gets called.
- Refactors validation around whether a user is permitted to reference a given foreign keyed object when creating or updating an instance of a resource.
  - This validator is `apps.common.validators.ValidateFkReferenceOwnership`.
  - This validator allowed `apps.api.permissions.HasCreatePrivilegesRole` and the `user_has_create_permission_by_virtue_of_role` method on some models to be retired.
- Adds `drf_extra_fields` as a dependency so we can use `PresentablePrimaryKeyRelatedField`, which provides a standardized way to refer to a model by primary key in update/create, while serializing in read operations (which was being handled inconsistently across different resources before this). This allowed retiring `apps.common.fileds.SerializedOnReadElsePk`
- Adds an .eslintrc to turn off some overly aggressive rules from testing-library which were errantly flagging the custom `cy.getByTestId` command as problematic.
- Updates some fixtures in line with recent api changes
- **Pins django-simple-history to 3.0.0**: This became necessary late in the game. There were strange problems happening in CI when `python manage.py makemigrations --dry-run --check` ran. Locally, that would not show any missing migration, but in CI it did. I discovered that reverting django-simple-history stopped that problem (without understanding underlying problem).  One failed run can be seen here: https://github.com/newsrevenuehub/rev-engine/runs/6560719461?check_suite_focus=true#step:9:50

Business reasons are discussed in parent ticket.

More specifically, regarding retiring the PageTemplateEmail model, there was no clear use case for the abstraction provided by the model, and it was built in such a way that using a single template to handle both one-time and recurring contributions would go against the semantics of the code.  There's a fair amount of overhead to having this data model without any clear business value.

The advantage of the new approach is that it simplifies the connection with the ESP. The `send_contribution_confirmation_email` (as well as `send_email` which it calls internally) take a template identifier, a to email address, a subject, and template variables (a dict) that gets sent to the ESP, and the task gets called (with `.delay`) in the stripe payment function. 

- After uploading the template to MailGun, in the environment running the app, set an environment variable with the name `ESP_TEMPLATE_ID_FOR_CONTRIBUTION_CONFIRMATION` to the value of the uploaded MailGun email template identifier.
- Start the server (or restart after ensuring updated env is sourced)
- Visit a donation page and make a recurring contribution using an email address that you have access to.
- Check your inbox and confirm that a confirmation email came.

Not added, but updated to ensure task for sending confirmation emails gets called in both one time and recurring cases.

There were two old migrations that were importing directly from apps.emails.models in order to dynamically set default values. In general, it's best to avoid importing models in this way into migrations because it can make it tricky to remove models. 

After removing the PageTemplateEmail and its models file, I was unable to make or run migrations because refs to the PageTemplateEmail model were broken. To get around this, I removed these references from the old migrations. This doesn't seem to create any problems locally, but changing old migrations is possibly tricky (in this case, I think it's fine).

Additionally, it became necessary to pin django-simple-history to 3.0.0 (see note above). At some point it may be worthwhile to run down that issue.

[DEV-1056]

- Upload the new email template to MailGun and grab the template identifier
- Set `ESP_TEMPLATE_ID_FOR_CONTRIBUTION_CONFIRMATION` env var on Heroku to the value of the identifier


njharman (changes requested):
Conflicts from CI need to be resolved.

It's hard to review this as it includes all the changes from other branch mentioned in PR that I just reviewed (I think).

Can that, DEV-1578_and_DEV-699, be merged first.

benjaminEwhite:
@njharman merged in main, so no more conflicts.

Tests pass, but it looks like in retiring some of the apps.email code, the overall test coverage has fallen below 95%. Should be good for a review, and we'll have to decide what to do re: test coverage.

revengine/settings/base.py
```diff
@@ -296,6 +296,8 @@
 # Transactional Email
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 EMAIL_SUBJECT_PREFIX = "[RevEngine] "
+ESP_TEMPLATE_ID_FOR_CONTRIBUTION_CONFIRMATION = os.environ.get("ESP_TEMPLATE_ID_FOR_CONTRIBUTION_CONFIRMATION")
```
njharman, line 299: Please add comment explaining what this setting/envvar is or reference to explanation (if on confluence for example).

njharman (changes requested):
Your PR notes are excellent. Answered several questions.

Do I understand correctly; the "nrh-default..." email template is example of what would be uploaded to ESP. And the specific template to be used is set by that environment var. (also envar to set magic link email tempalte). This is in place of / why we nuked template from our database?

But don't customers need to customize those templates? and there is only one template (set by envvar)?  We run rev-engine as service / not customer self-hosted right?

Few other questions / thoughts done in line.

apps/contributions/views.py
```diff
@@ -50,29 +51,51 @@ def stripe_payment(request):
...
-            response_body = {"detail": "Success"}
+        else:
+            logger.warning(f"stripe_payment view recieved unexpetected interval value: {interval}")
+            raise PaymentBadParamsError()
+        contributor_email = stripe_payment.validated_data["email"]
+        donation_amount_display = f"${(stripe_payment.validated_data['amount'] / 100):.2f}"
+        contribution_date = timezone.now()
+        interval_to_display = {ContributionInterval.MONTHLY.value: "month", ContributionInterval.YEARLY.value: "year"}
+
+        template_data = {
```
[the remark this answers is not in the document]
benjaminEwhite: I copied the template from an existing one, and at time, had not run down whether or not the ESP somehow provided values for those template vars (they're in status quo template too, but nothing gets passed in). 

Just pushed up a commit that removes intro and blurb and formats the template file so it's more readable.

apps/organizations/models.py
```diff
@@ -105,7 +105,6 @@ class Organization(IndexedTimeStampedModel, RoleAssignmentResourceModelMixin):
     )
     stripe_product_id = models.CharField(max_length=255, blank=True)
     domain_apple_verified_date = models.DateTimeField(blank=True, null=True)
-    uses_email_templates = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
benjaminEwhite: Yeah, I think that's more a description of the status quo. Mariclare and I Slacked about removing this criterion. I'll confirm with her.

benjaminEwhite:
@njharman re: high level questions:

> Do I understand correctly; the "nrh-default..." email template is example of what would be uploaded to ESP. And the specific template to be used is set by that environment var. (also envar to set magic link email tempalte). This is in place of / why we nuked template from our database?

All correct, with a bit of nuance on the last point. The app already had that page email template model, which collects a donation page and a reference to a template on an ESP. We only have one use case for transactional emails right now, and I concluded that the maintenance and ease of adding new email functionality would be better if we removed the data model entirely, and instead send emails more directly.  One specific reason I reached this conclusion is that it made sense to have a single email template for both one-time and recurring donations. As implemented, the PageTemplateEmail model wanted me to choose between distinct template types, where one-time and recurring were distinct options.

Overall, I think if you're using an ESP, you kinda just wanna throw the right data over the fence, and lean on the template layer. Sometimes it can be nice to model that an email attempt was made so you can track a denominator of attempts vs, say, bounce rate, but that's not what the PageTemplateEmail model was getting at. 

> But don't customers need to customize those templates? and there is only one template (set by envvar)? We run rev-engine as service / not customer self-hosted right?

There's no business requirement around customizing templates at the moment. And if that need does arise, I think we'll want to model things a bit differently.

Yes, rev-engine as a service. 

And yes re: only one template (set by env var). That's how things work in status quo as well (but only for one time contributions).

apps/contributions/views.py
```diff
@@ -50,29 +51,51 @@ def stripe_payment(request):
...
+            ContributionInterval.YEARLY.value,
+        ):
             stripe_payment.create_subscription()
-            response_body = {"detail": "Success"}
+        else:
+            logger.warning(f"stripe_payment view recieved unexpetected interval value: {interval}")
+            raise PaymentBadParamsError()
+        contributor_email = stripe_payment.validated_data["email"]
+        donation_amount_display = f"${(stripe_payment.validated_data['amount'] / 100):.2f}"
+        contribution_date = timezone.now()
```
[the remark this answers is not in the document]
benjaminEwhite: That's actually a good question. I'm not sure what we want here.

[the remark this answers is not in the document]
benjaminEwhite: Asked, and answer is that this is desired behavior for status quo. We're tracking a separate ticket for moving to what's stored in Stripe.

apps/contributions/views.py
```diff
@@ -50,29 +51,58 @@ def stripe_payment(request):
...
-            # Create subscription with Stripe, associated local models
+            response_body["clientSecret"] = stripe_payment_intent["client_secret"]
+        elif interval in (
+            ContributionInterval.MONTHLY.value,
+            ContributionInterval.YEARLY.value,
+        ):
             stripe_payment.create_subscription()
-            response_body = {"detail": "Success"}
+        else:
+            logger.warning(f"stripe_payment view recieved unexpetected interval value: {interval}")
```
njharman, line 68: logging should use % formatting.

apps/contributions/views.py
```diff
@@ -50,29 +51,58 @@ def stripe_payment(request):
...
+                "contribution_interval_display_value": interval_to_display.get(
+                    stripe_payment.validated_data["interval"]
+                ),
+                "copyright_year": contribution_date.year,
+                "org_name": rp.organization.name,
+            }
+            send_contribution_confirmation_email.delay(contributor_email, **template_data)
+
+    except RevenueProgram.DoesNotExist:
+        logger.warning(
```
njharman, line 94: logging should use % formatting.

benjaminEwhite:
@njharman I made requested changes re: string formatting in log messages.

@x110dc your ship for final(?) review.

Also, fwiw, I don't see any merge conflicts locally (main is merged into this branch). Think the conflicting files above are a relic of the PR (or may in fact have to be resolved when ready to merge).

x110dc (approved):
Excellent PR description. Thanks. LGTM.

revengine/celery.py
```diff
@@ -11,7 +11,7 @@
 
 # Using a string here means the worker doesn't have to serialize
 # the configuration object to child processes.
-app.config_from_object("django.conf:settings")
+app.config_from_object("django.conf:settings", namespace="CELERY")
```
x110dc, line 14: Why was the namespace added?

x110dc (changes requested):

revengine/celery.py
benjaminEwhite: I had to do that in order to get Celery to run locally, which I needed to do in order to confirm that the feature was working. Was an oversight letting that go through PR. Will remove.

benjaminEwhite:
@x110dc made requested changes. As noted in PR notes at top (after I edited them), the merge issues stemmed from an upgrade of django-simple-history. In order to move this ticket across the board, I pinned django-simple-history to 3.0.0, which solves the problem, but at some point in future we may want to look into this issue.

x110dc (approved):

x110dc (approved):

PR #415, opened 2022-05-09
Dev 1517 add query params to thank you page internal and Dev 1518 for external

gurjittoptal:

Dev 1517/Dev 1518, both included in single PR. Should be merged after Dev 1516

#### What's this PR do?
Adds query string parmams uid, frequency and amount to internal and external redirects

analytics

by making payments on donation pages with
- default internal redirect
- external redirect with and without existing params

#### Have automated unittests been added?
na

na

na

na

[DEV-1517],[DEV-1518]

[DEV-1517]: https://news-revenue-hub.atlassian.net/browse/DEV-1517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-1518]: https://news-revenue-hub.atlassian.net/browse/DEV-1518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spa/src/components/paymentProviders/stripe/StripePaymentForm.js
```diff
@@ -172,8 +182,16 @@ function StripePaymentForm({ loading, setLoading, offerPayFees }) {
 
   const handleCardSubmit = async (e) => {
     e.preventDefault();
+    const email = extractEmailFromFormRef(formRef.current);
+    getSHA256Hash(email)
+      .then((res) => {
+        setEmailHash(res);
+      })
+      .catch((err) => console.log(err));
```
benjaminEwhite, line 190: This line is unreachable because `getSHA256Hash` catches all errors and doesn't re-raise.

spa/src/utilities/getSHA256Hash.js
```diff
@@ -0,0 +1,11 @@
+export async function getSHA256Hash(message) {
+  try {
+    const msgUint8 = new TextEncoder().encode(message); // encode as (utf-8) Uint8Array
+    const hashBuffer = await crypto.subtle.digest('SHA-256', msgUint8); // hash the message
+    const hashArray = Array.from(new Uint8Array(hashBuffer)); // convert buffer to byte array
+    const hashHex = hashArray.map((b) => b.toString(16).padStart(2, '0')).join(''); // convert bytes to hex string
+    return hashHex;
+  } catch (err) {
+    console.log(err.message);
```
benjaminEwhite, line 9: Seems to me the try/catch should be removed here. As is, any exceptions will be caught, and the function will successfully complete, which if i'm not mistaken means that emailHash will be set to undefined but the query string will still be generated.

benjaminEwhite (changes requested):
Overall looks good. Think should consider getting rid of the try + bare catch statement.

benjaminEwhite:
One more high-level comment: there should be a test demonstrating that `handlePaymentSuccess` causes the right query params to be pushed onto history. Otherwise we're depending entirely on maintainers remembering the business logic around these changes and preserving their implementation.

spa/src/components/paymentProviders/stripe/StripePaymentForm.js
gurjittoptal: hmm. right. this has been removed now.

spa/src/utilities/getSHA256Hash.js
gurjittoptal: So, the reason to keep the try/catch is that Crypto.subtle is available only in secure contexts or localhost in browsers that implement specs ( most modern browsers ). In others a '' is passes as uid. 

@x110dc this is an FYI for you too. Let me know is some change is needed here.

benjaminEwhite (approved):

x110dc (approved):

PR #426, opened 2022-05-17
DEV-1021 enable feature flagging

benjaminEwhite:

This PR is the first in several that will bring feature flagging to the front- and back-end for rev-engine.

This PR adds django-waffle and integrates feature flagging with the permissions layer (initially for a single endpoint: contributions). 

Later PRs will expose a custom flag model created in this PR as an API resource which can be used in the SPA to gate access to features there. 

#### What's this PR do?

- Adds a custom `Flag` model (which initially is functional equivalent of the default waffle flag) that can later be extended to integrate with roleassignment (so flags can be rolled out for entire programs, etc.).
- Adds an admin model for `Flag`
- Adds a `HasFlaggedAccessToContributionsApiResource` permission that is used to gate the contributions api endpoint to users for whom the related flag is not active. 
- Uses the new feature flag approach to gate access to the contributions API resource (not inclusive of stripe specific API endpoints).
- Adds two tests to `apps.contributions.tests.test_views.TestContributionsViewSet` around feature flagging. The first is the "happy path", and the second shows behavior if `HasFlaggedAccessToContributionsApiResource` can't find its required flag on initialization.

Described in detail on JIRA.

(first three steps are only if you're running locally)

1. Check out this branch.
2. Run migrations
3. Run dev server 
4. In the Django admin, go to the Flag admin, and create a flag whose name is the same value as `apps.flags.constants.CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME` is set to (in source code right now, that's "contributions-api-endpoint-access"). Set `.everyone` to False, `superusers` to True, and add the user ID for a non superuser to `.users`. Save.
5. Try logging in as a superuser, then go to `/api/v1/contributions/` in a browser tab. You should be able to access. Alternatively, you can log in to the SPA as a superuser, and the contributions tab should populate with any contributions that are in the db.
6. Try logging in as a non super user who hasn't been explicitly added. Follow the same steps as 5. This time the API should return a 403 error.
7. Try logging in as the explicitly added user from step 4. Follow same steps as 5. This time the API should return any results (and 200).
8. In the Django admin, update the feature flag's `everyone` value to True. 
9. Repeat step 6. This time the API should 200 and return any results that user is allowed to see.

Yes

I don't believe so. Once this and initial, related SPA-side feature flag tickets are complete, would make sense to consolidate knowledge around feature flagging in Confluence.

[DEV-1021]

After this is merged, we'll need to create a feature flag with name set to the value of `apps.flags.constants.CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME`. That flag should have `superusers` set to True, `everyone` to "Unknown", and any individual users should have their IDs added to the `users` field in the admin.


apps/contributions/views.py
```diff
@@ -205,7 +210,13 @@ class ContributionsViewSet(viewsets.ReadOnlyModelViewSet, FilterQuerySetByUserMi
     # are permitted
     permission_classes = [
         IsAuthenticated,
-        IsActiveSuperUser | HasRoleAssignment | IsContributorOwningContribution,
+        (
+            # `IsContributorOwningContribution` roleassignment based permissions because contributors
+            # will not be granted permission by those.
```
njharman, line 215: I think this is saying flag not checked for contributors because flags aren't implemented for roles? contributors yet?

apps/contributions/views.py
```diff
@@ -205,7 +210,13 @@ class ContributionsViewSet(viewsets.ReadOnlyModelViewSet, FilterQuerySetByUserMi
...
+            IsContributorOwningContribution
+            | (HasRoleAssignment & HasFlaggedAccessToContributionsApiResource)
+            | (IsActiveSuperUser & HasFlaggedAccessToContributionsApiResource)
```
njharman, line 218: Minor but it conveys they logic better IMO (assuming I understood logic correctly)

change to be IsCon... | (Hasflag & (Hasrole.. | isacivie...| N...))

apps/contributions/tests/test_views.py
```diff
@@ -510,6 +511,42 @@ def test_unexpected_role_type(self):
...
+        """Show feature users are gated by feature flag for this resource"""
+        flag_model = get_waffle_flag_model()
+        contributions_access_flag = flag_model.objects.get(name=CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME)
+        contributions_access_flag.everyone = False
+        contributions_access_flag.superusers = True
+        contributions_access_flag.users.add(self.hub_user)
+        contributions_access_flag.save()
+        expected_users_having_access = (
+            self.superuser,
+            self.contributor_user,
```
njharman, line 526: I think Inderstand why contibutor_user has access despite no flag set for them.  A comment here along lines of "contributor bypasses flag check because ???, not implemented yet?"

apps/contributions/tests/test_views.py
```diff
@@ -510,6 +511,42 @@ def test_unexpected_role_type(self):
             reverse("contribution-list"), novel, expected_status_code=status.HTTP_500_INTERNAL_SERVER_ERROR
         )
 
+    #######################
+    # Test feature flagging
+    def test_feature_flagging(self):
```
njharman, line 516: This test is susceptible to false positives because it sets up several conditions and and checks them together.

It is better to construct tests setting one thing, all other things "false", and check that one thing. Also check the inverse.

So, set only flag.superusers = True, verify superuser allowed, allothers denied.  Then set flag.superusers = false verify superuser denied. 

pytest.parameterize is good for this.

apps/flags/models.py
```diff
@@ -0,0 +1,12 @@
+from waffle.models import AbstractUserFlag
+
+
+class Flag(AbstractUserFlag):
+    """Custom flag model
+
+    Initially, we are not overriding anything from `AbstractUserFlag`,
```
njharman, line 7: You argue reasonable "justification". Still feels icky, https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

apps/contributions/tests/test_views.py
```diff
@@ -510,6 +511,42 @@ def test_unexpected_role_type(self):
...
+            self.org_user,
+            self.rp_user,
+            self.generic_user,
+        )
+        for user in expected_users_not_having_access:
+            self.assert_user_cannot_get(self.list_url, user)
+        for user in expected_users_having_access:
+            self.assert_user_can_get(self.list_url, user)
+
+    def test_feature_flagging_when_flag_not_found(self):
```
njharman, line 539: Why did the CREATE_FLAGIFMISSING waffle setting not create this flag?

njharman (changes requested):
Just code review, no testing.

Questions, suggested comment/docs, and philosophical ramblings.

apps/contributions/views.py
benjaminEwhite: Some how I deleted part of the code comment! Thanks for catching this!

apps/contributions/views.py
benjaminEwhite: Agree

apps/contributions/tests/test_views.py
benjaminEwhite: Left a comment. Reason here is actually because I did not apply the flag to contributors since that would prohibit them from getting any meaningful data out of the contributor dashboard.  Believe this is desired behavior, but will confirm.

apps/flags/models.py
benjaminEwhite: Confidence is relatively high that this will be needed in short order, but I agree with the general principle and will remove.

benjaminEwhite: Am removing models and admin, but I'm keeping the app, since the next ticket I'm working on will require exposing flags as an API resource, and that will happen in the flags app.

apps/contributions/tests/test_views.py
benjaminEwhite: I'm removing `WAFFLE_CREATE_MISSING_FLAGS` because there's not an immediate use case. 

To answer your question: it looks like missing flags only get created when calling the `is_active` method on a flag, whereas `HasFlaggedAccessToContributionsApiResource` calls `.is_active_for_user` on the flag.

apps/contributions/tests/test_views.py
benjaminEwhite: Good reason for me to finally get up to speed on `pytest.parameterize`.

njharman (approved):
LGTM

apps/api/permissions.py
```diff
@@ -3,8 +3,10 @@
 from django.conf import settings
 
 from rest_framework import permissions
+from waffle import get_waffle_flag_model
 
 from apps.contributions.models import Contributor
+from apps.flags.constants import CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME
```
x110dc, line 9: What's the value of having this as a variable? In what cases would we want to change the value?

x110dc (approved):

apps/api/permissions.py
benjaminEwhite: When there are flags being managed across several files, IMO it's helpful to have a convention around storing flag names in a single place. That makes it easy to see at-a-glance what all the flags are.

Not a strongly held opinion though.

PR #427, opened 2022-05-19
DEV-1684 - protect contributions section behind a feature flag Part 1 (API changes)

benjaminEwhite:

In working on [DEV-1684], I've decided to break up the work into two separate PRs, an initial one (this one) that is limited to the backend changes required to add active flags to a `.flags` property on serialized users.  In a separate PR, I'll tackle the SPA changes that rely on this data.  This will allow a smaller diff initially that I can pass to Norm for backend review, and then for part 2, can have a dev with React expertise review.

NB: The base for this PR is [DEV-1021] since this branches from that branch, which itself is in an unmerged PR. Once that has been merged, I will change the base for this PR to main. Until then, this approach will allow a review that focuses on narrow changes of this PR, rather than broader set of changes vis-a-vis main that would be in diff from unmerged DEV-1021.

#### What's this PR do?

- Gets rid of the flags app entirely, as I opted to serialize flag data in users (@njharman you were right about not needing those app files 😉 ).
- Moves `CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME` to `apps.common.constants`.
- Adds a `flags` property to the user serializer and a related `get_active_flags_for_user` serializer method. This method retrieves all flags that are active for a user (taking into account whether or not the user is a superuser).
- Adds `apps.users.tests.test_serializers.test_user_serializer_flags` demonstrating expected behavior.

First of two PRs to ship [DEV-1684], the business value of which is articulated in JIRA.

1. Create a flag. In Django admin, set "everyone" to "Yes". Save.
2. Log in as the user (either through Django admin or through SPA).
3. In a browser tab, go to `/ap1/v1/users/`. You should see a `flags` property on the returned user object, and it should be a length-1 list whose only item is a dict whose values for id and name are same as the flag you created.
4. Go through above flow, but try different combinations of flag settings and logged in users (for instance, set `everyone` to False, and `superusers` to True on the flag, and then log in as a super user and a non superuser -- only the superuser should have flags).

Yes

[DEV-1684]: https://news-revenue-hub.atlassian.net/browse/DEV-1684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DEV-1021]: https://news-revenue-hub.atlassian.net/browse/DEV-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/common/constants.py
+++ b/apps/common/constants.py
@@ -65,5 +65,6 @@ STATE_CHOICES = [
     ("YT", "Yukon"),
 ]
 
-
+# HasFlaggedAccessToContributionsApiResource references a flag with this name in order
+# to determine access permission.
 CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME = "contributions-api-endpoint-access"
--- a/apps/common/constants.py
+++ b/apps/common/constants.py
@@ -64,3 +64,6 @@ STATE_CHOICES = [
     ("SK", "Saskatchewan"),
     ("YT", "Yukon"),
 ]
+
+
+CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME = "contributions-api-endpoint-access"
--- a/apps/users/tests/test_serializers.py
+++ b/apps/users/tests/test_serializers.py
@@ -1,7 +1,10 @@
 from django.contrib.auth import get_user_model
 
-from rest_framework.test import APITestCase
+import pytest
+from rest_framework.test import APIRequestFactory, APITestCase
+from waffle import get_waffle_flag_model
 
+from apps.contributions.tests.factories import ContributorFactory
 from apps.organizations.models import Organization, RevenueProgram
 from apps.organizations.tests.factories import OrganizationFactory, RevenueProgramFactory
 from apps.users import serializers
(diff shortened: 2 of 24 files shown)


apps/common/constants.py
```diff
@@ -64,3 +64,6 @@
     ("SK", "Saskatchewan"),
     ("YT", "Yukon"),
 ]
+
+
+CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME = "contributions-api-endpoint-access"
```
njharman, line 69: Why here and not in contributions app? 

Also isn't contributions-api-endpoint-access a Django url config name? (I forget the exact Django terminology) what you can use in code and templates to refer to the URL of an endpoint?  I don't see "contributions-api-endpoint-access" anywhere in code other than this global.

apps/users/tests/test_serializers.py
```diff
@@ -127,3 +130,66 @@ def test_listed_revenue_programs_include_org_id(self):
...
+
+    flag2 = Flag.objects.create(name="flag2", everyone=flag2_everyone, superusers=flag2_superusers)
+    if flag2_add_user:
+        flag2.users.add(user)
+        flag2.save()
+
+    request = APIRequestFactory().get("/")
+    request.user = user
+    data = serializers.UserSerializer(user, context={"request": request}).data
+    expected_flag_count = len([x for x in [expect_flag1, expect_flag2] if x])
```
njharman, line 186: eh? sum(bool(x) for x in [expect_flag1, expect_flag2])

apps/users/tests/test_serializers.py
```diff
@@ -127,3 +130,66 @@ def test_listed_revenue_programs_include_org_id(self):
...
+    assert len(data["flags"]) == expected_flag_count
+    if expect_flag1:
+        assert any([flag["name"] == flag1.name and flag["id"] == flag1.id for flag in data["flags"]])
+    if not expect_flag1:
+        assert not any([flag["name"] == flag1.name and flag["id"] == flag1.id for flag in data["flags"]])
+    if expect_flag2:
+        assert any([flag["name"] == flag2.name and flag["id"] == flag2.id for flag in data["flags"]])
+    if not expect_flag2:
+        assert not any([flag["name"] == flag2.name and flag["id"] == flag2.id for flag in data["flags"]])
```
njharman, line 195: I think the outer square brackets (forming list) are redundant within function call parens (it will be generator, which any can handle, bug for instance len() cannot) [I'm surprised linter did not mention/complain (mine did complain about ContributorFactor imported and not used though...)]

> any(flag["name"] == flag1.name and flag["id"] == flag1.id for flag in data["flags"])

Maybe something is required to bind and more tightly?

Minor, but already lots of punctuation in that line.

apps/users/tests/test_serializers.py
```diff
@@ -127,3 +130,66 @@ def test_listed_revenue_programs_include_org_id(self):
...
+        flag2.users.add(user)
+        flag2.save()
+
+    request = APIRequestFactory().get("/")
+    request.user = user
+    data = serializers.UserSerializer(user, context={"request": request}).data
+    expected_flag_count = len([x for x in [expect_flag1, expect_flag2] if x])
+    assert len(data["flags"]) == expected_flag_count
+    if expect_flag1:
+        assert any([flag["name"] == flag1.name and flag["id"] == flag1.id for flag in data["flags"]])
```
njharman, line 189: eh?
flags = [(x["name"], x["id"]) for x in data["flags"]
# then in if statments
   assert (flag1.name, flag1.id) in flags:

[to be clear "eh?" here and elsewhere, doesn't mean change it.  Just offering possibly clearer (for you to decide) method for same result.]

njharman (approved):
Approved with "things to consider"

apps/common/constants.py
benjaminEwhite: Back when I had a flags app, this lived there in a constants file. I'm not particularly wedded to this location.

apps/users/tests/test_serializers.py
benjaminEwhite: Much more readable.

apps/common/constants.py
benjaminEwhite: `CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME` can have an arbitrary value and if this collides with something Django's looking for, can be changed. The `HasFlaggedAccessToContributionsApiResource` references a flag with this name in order to determine access permission.

benjaminEwhite: I'll add a note.

gurjittoptal:
@benjaminEwhite Won't this PR need a migrations file for django waffle model?

benjaminEwhite:
@gurjittoptal - Django Waffle adds its own migrations, and you don't have to create them with `python manage.py makemigrations`.   Below is the output when running `python manage.py migrate` (without having created any additional migrations).

Screen Shot 2022-05-23 at 9 02 53 AM

x110dc (approved):

PR #431, opened 2022-05-23
DEV-956: Restrict style dropdown in page editor based on active parent element

benjaminEwhite:

#### What's this PR do?

This PR filters styles available in the dropdown in the page editor. It does this by including a query parameter for the current RP when making a request to retrieve styles from the server.

This required adding a filter to the styles API resource so that can be filtered by revenue program ID.

Makes the page editor easier to use for  users who have access to many styles.

Log in to the SPA as, say, an RP admin user for an org that has many styles. Either create or edit an existing page, and in the page editor, go to the styles tab. The dropdown should only include styles for the RP of the page.

No.

[DEV-956]

No

No

### DIFF
--- a/apps/pages/filters.py
+++ b/apps/pages/filters.py
@@ -1,0 +1,11 @@
+import django_filters
+
+from apps.pages.models import Style
+
+
+class StyleFilter(django_filters.FilterSet):
+    revenue_program = django_filters.NumberFilter(field_name="revenue_program__id")
+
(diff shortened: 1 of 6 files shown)


gurjittoptal (commented):
LGTM

gurjittoptal (approved):
LGTM

x110dc (approved):

PR #433, opened 2022-05-24
DEV-905 — Protect content section behind a feature flag

benjaminEwhite:

Initially have set base as `DEV-1021-enable-feature-flagging` as it has not been merged, and this branch requires those changes. If that branch is merged before this PR, then base should be changed to `main`.

#### What's this PR do?

Gates the content section and sidebar item behind a feature flag. As per AC, when a user tries to access the content section and has not been granted access, they receive a 404. 

The 404 requirement necessitated touching more code than I had expected, but all to good effect. 

Here are some noteworthy changes in this PR:

- Makes displaying "Content" in sidebar and Content section contingent on a flag with the name defined at spa.src.constants.featureFlagConstants.CONTENT_SECTION_ACCESS_FLAG_NAME appearing in user.flags (where the user data is returned by the /api/v1/users/ endpoint).
- Updates existing Cypress tests that broke with introduction of content section flag.
- Upgrades `@cypress/webpack-preprocessor` in order to get past errors that started happening when running Cypress locally (as per [this suggestion](https://github.com/cypress-io/cypress/issues/8948#issuecomment-1021681739))
- Moves two dev dependencies into dev dependencies (had been saved as non-dev dependencies before).
- **Retires** the `useFeatureFlags` hook in favor of using context/provider. This became necessary in order to allow a "higher up the tree" component to initially retrieve user flags before child components are allowed to load.
- Non-trivial refactoring of routing. The donation page editor is now a child of the org dashboard (whereas before it was a sibling). Donation page editor is now at /dashboard/edit/my/page whereas before, it was at /edit/my/page.
- Moves, `LivePage404` to spa/src/components/common (after checking it out from DEV-1678) which is used for displaying 404s within the dashboard app

See JIRA.

1. Log in to the Django admin
2. Create a flag whose name is the value defined in spa.src.constants.featureFlagConstants.CONTENT_SECTION_ACCESS_FLAG_NAME. You can choose which settings (everyone or for a specific user).
3. In a separate, incognito browser tab, log in to the SPA as a user for whom you have granted the flag described above. You should see the content section and a link to it in the sidebar.
5. Back in Django admin, update the flag so that it no longer applies to the user you logged in to the SPA with.
6. Refresh the SPA. This time, you should not be able to see any reference to content and instead should see a 404 page.

Yes

After this is merged, it will be necessary to create a feature flag in the Django admin with the name defined in spa.src.constants.featureFlagConstants.CONTENT_SECTION_ACCESS_FLAG_NAME, and then grant access as required.


benjaminEwhite:
@gurjittoptal i just realized there’s an unmerged or that adds 404 that you did (and I even approved!), so I’ll look into using that component instead if possible.

gurjittoptal (approved):
Looks good.

benjaminEwhite:
Tests passing again. Your ship @x110dc

x110dc (approved):

PR #437, opened 2022-05-27
remove stripe account from contributions api

harsha09:
#### What's this PR do?
- Removes `org_stripe_id` from response object of `/api/v1/contributions`.
- Removes the usage of `org_stripe_id` from frontend where it's being used, which is not necessary. https://github.com/newsrevenuehub/rev-engine/blob/c4caabee6b158be95a1d3c3fe076bad856eeb558/spa/src/components/contributor/contributorDashboard/EditRecurringPaymentModal.js#L37

Endpoint `/api/v1/contributions` is exposing `org_stripe_id`, which is confidential information and can potentially be abused by bad actors.

Send a Get request to `/api/v1/contributions` and `org_stripe_id` shouldn't come in the response.
Update payment method should work as it is. (I was unable to test this end-to-end in my local)

No

NA

NA

NA

[DEV-1752]

NA

NA


x110dc (approved):
LGTM.

PR #459, opened 2022-06-13
Dev-1691 new data source for contributor portal

harsha09:

Modify existing Endpoint to pull contributions data from stripe instead of application database for contributors portal.

#### What's this PR do?
1. Celery tasks will pull contributions data from stripe and loads it to redis cache when ever a magic link is requested.
2. As contributor access the contributor portal, the stripe data from cache is served.
3. Revenue Program is send as query params from contributor portal.
4. All contributions are faceted by revenue program.

Contributor will be able to view contributions done before RevEngine went live.

Request for magic link.
Open the contributor portal with the magic link you got.
You should be able to view the contributions faceted by the revenueprogram.

Yes.

NA

Celery worker runs when the user requests a magic link, data gets saved in redis for about 30 mins.
If data is not available in cache / expired the celery worker gets triggered again on contributor portal access.
This causes delay to view contributions and user has to keep on refreshing the page for contributions. This problem needs to be addressed in future.

Not yet

[DEV-1691]
https://news-revenue-hub.atlassian.net/browse/DEV-283
https://news-revenue-hub.atlassian.net/browse/DEV-1567

NA

NA

[DEV-1691]: https://news-revenue-hub.atlassian.net/browse/DEV-1691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/serializers.py
```diff
@@ -463,3 +463,22 @@ class StripeRecurringPaymentSerializer(AbstractPaymentSerializer):
...
 
     payment_method_id = serializers.CharField(max_length=255)
+
+
+class PaymentProviderContributionSerializer(serializers.Serializer):
+    """
+    Payments provider serializer, payment provider Eg: Stripe.
+    """
+
+    status = serializers.CharField(max_length=10)
```
x110dc, line 473: I'm not sure exactly how these fields work, but should some be more specific, like a `ChoiceField` since their values can only be a limited set of choices? Like status, payment_type, interval, and maybe card_brand?

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,168 @@
...
+
+import stripe
+
+from apps.contributions.models import ContributionInterval, ContributionStatus
+from apps.contributions.serializers import PaymentProviderContributionSerializer
+
+
+MAX_STRIPE_RESPONSE_LIMIT = 100
+MAX_STRIPE_CUSTOMERS_LIMIT = 10
+REDIS_TTL = 60 * 60 * 2
```
x110dc, line 16: I'd a comment about the unit, i.e. seconds.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,168 @@
...
+CACHE_DB = "default"
+
+logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
+
+
+class StripeEventIgnoreError(Exception):
+    pass
+
+
+class InvalidIntervalError(StripeEventIgnoreError):
```
x110dc, line 26: I'd recommend calling them something besides events here, because that could be confused with actual Stripe events, which isn't what these are, right?

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,168 @@
...
+        self.charge = charge
+
+    @property
+    def interval(self):
+        interval = self.charge.invoice.lines.data[0].plan.interval
+        interval_count = self.charge.invoice.lines.data[0].plan.interval_count
+        if interval == "year" and interval_count == 1:
+            return ContributionInterval.YEARLY
+        if interval == "month" and interval_count == 1:
+            return ContributionInterval.MONTHLY
```
x110dc, line 57: While these are the only three intervals that NRE supports directly, there could be other intervals in Stripe. I don't think we should assume it's "one time" if it's not monthly or yearly. If one comes across with interval=month and interval_count=2 (bi-monthly) this would call it one-time. I think it's okay for now to throw an exception for an unexpected interval.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,168 @@
...
+        next_attempt = self.charge.invoice.next_payment_attempt
+        if next_attempt:
+            return datetime.utcfromtimestamp(int(next_attempt))
+
+    @property
+    def refunded(self):
+        return self.charge.refunded or self.charge.amount_refunded > 0
+
+    @property
+    def id(self):
```
x110dc, line 126: This overrides the built-in `id` function. I usually append an underscore to avoid this. I.e. `def id_(self)`

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,168 @@
...
+
+def pull_stripe_contributions_to_cache(email_id):
+    stripe_provider = StripeContributionsProvider(email_id)
+    data = {}
+    for charge in stripe_provider.get_charges():
+        try:
+            serializer = PaymentProviderContributionSerializer(instance=StripeCharge(charge))
+            data[charge.id] = serializer.data
+        except StripeEventIgnoreError as ex:
+            logger.info(ex)
```
x110dc, line 159: Maybe it can/should be done in a subsequent ticket, but I think we should segregate by RP by including the RP slug in the cache key.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,168 @@
...
+
+    caches[CACHE_DB].set(email_id, json.dumps(data), timeout=REDIS_TTL)
+
+
+def load_stripe_data_from_cache(email_id):
+    data = caches[CACHE_DB].get(email_id)
```
x110dc, line 165: Are we distinguishing between a cache result that's empty (because Stripe had no charges for the email) and an entry that's expired/non-existent? 

I'd think for the former we'd want to return the cached results and for the latter we'd want to fetch new results. I'm not sure at what abstraction layer that would be best. Also, given that the cache TTL is longer than the magic link TTL maybe this is an edge case that doesn't need to be solved (yet).

revengine/settings/dev.py
```diff
@@ -55,3 +55,25 @@
...
 
 ESP_TEMPLATE_ID_FOR_CONTRIBUTION_CONFIRMATION = "someFakeValue"
+
+REDIS_URL = os.getenv("REDIS_TLS_URL", os.getenv("REDIS_URL", "redis://redis:6379"))
+CACHE_HOST = REDIS_URL
+CONNECTION_POOL_KWARGS = {}
+if CACHE_HOST.startswith("rediss"):
+    import ssl
+
+    # See: https://github.com/mirumee/saleor/issues/6926
```
x110dc, line 65: Did you actually run into an issue here or just coding defensively?

revengine/settings/dev.py
```diff
@@ -55,3 +55,25 @@
 TEMPLATES[0]["DIRS"] = [FRONTEND_BUILD_DIR, os.path.join(PROJECT_DIR, "templates")]
```
x110dc, line 56: Shouldn't most/all of this go into `settings/base.py`?

x110dc (changes requested):
This looks great. I'd like @benjaminEwhite or @njharman to review as well.

apps/contributions/views.py
```diff
@@ -277,13 +278,16 @@ class ContributionsViewSet(viewsets.ReadOnlyModelViewSet, FilterQuerySetByUserMi
...
+            return load_stripe_data_from_cache(self.request.user)
+
         # this is supplied by FilterQuerySetByUserMixin
         return self.filter_queryset_for_user(self.request.user, self.model.objects.all())
 
     def get_serializer_class(self):
         if isinstance(self.request.user, UserModel):
             return serializers.ContributionSerializer
-        return serializers.ContributorContributionSerializer
+        return serializers.PaymentProviderContributionSerializer
```
benjaminEwhite, line 290: Should the check here be for `isinstance(self.request.user, Contributor)` (in which case `PaymentProviderContributionSerializer`), instead?  Seems like should be same check as `get_queryset`.

apps/api/views.py
```diff
@@ -154,6 +155,9 @@ def post(self, request, *args, **kwargs):
 
             if not domain:
                 return Response({"detail": "Missing Revenue Program subdomain"}, status=status.HTTP_404_NOT_FOUND)
+
+            pull_stripe_contributions_to_cache(email)
```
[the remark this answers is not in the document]
benjaminEwhite: Would it make sense for this to be an async Celery task instead of blocking (which might also handle re-pulling from Stripe if cache is empty?)

apps/contributions/stripe_contributions_provider.py
benjaminEwhite: I was also wondering about this edge case too. If nothing else, there should be a big code comment next to the cache TTL var explaining that the contributions API resource depends on this relationship with magic link TTL in order for there to be results.

Also wondering what happens if a past contributor 1.) gets a magic link to review past contributions; 2) soon after, makes a new contribution; 3.) then follows their magic link to look at their contributions again.  For the duration of the magic link TTL, they'll only be able to get cached results, which might lead someone to believe they haven't been charged again (in which case they might make an additional payment, or otherwise think the the system is flakey).

apps/contributions/views.py
```diff
@@ -277,13 +278,16 @@ class ContributionsViewSet(viewsets.ReadOnlyModelViewSet, FilterQuerySetByUserMi
     filterset_class = ContributionFilter
 
     def get_queryset(self):
+        if isinstance(self.request.user, Contributor):
+            return load_stripe_data_from_cache(self.request.user)
+
```
benjaminEwhite, line 283: Give that contributor users get both a different queryset and a different serializer, and that they will be consuming the results of this resource in a different UI (the contributor dashboard), it makes me wonder if it would make more sense to create a distinct viewset (something like `ContributorContributionsViewSet`) — one that is limited to contributor users for now.

apps/api/views.py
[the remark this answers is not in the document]
benjaminEwhite: Or, would it make sense to move the stripe pull/cache to the view layer? That would decouple the stripe caching layer TTL from magic link TTL, and also possibly make re-fetching more straightforward.

apps/contributions/stripe_contributions_provider.py
x110dc: After thinking about it more, the cache key should probably include something to identify the Stripe account (like Stripe account ID or the PK for it in Django).  Only using the email for the key means that someone could get cache results for an entirely different account.

apps/contributions/stripe_contributions_provider.py
x110dc: There's already a disclaimer at the top of the portal results addressing this ("Changes made may not be reflected immediately."). But yes, we should think through these scenarios. And thinking about it more, I guess it's not only the TTL for the magic link itself but the TTL for the session created by the magic link. And that is longer than the TTL for the cache. So they could expire while someone is still logged in unless we adjust that TTL to match.

harsha09: @x110dc: I agree with you, we can make cache TTL to match magic links TTL.
Based on @benjaminEwhite's concern, we can store the `max_contribution_created_date` and query StripeAPI for each request where `created_date > max_contribution_created_date` and store in the cache and return the cached results. This will make things complicate a bit, I made it simple as this will not be a long-term solution.

@x110dc : What would you suggest?

revengine/settings/dev.py
harsha09: @x110dc : I just copied it from `deploy.py` file. I think we can move it to `base.py` if redis cache is available for all deployments in Heroku.

apps/api/views.py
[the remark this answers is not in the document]
harsha09: @benjaminEwhite : Added it before sending magic link to improve the response time for user whenever he opens contribution portal.
[DEV-1691] is for enabling stripe contributions, will consider it in [DEV-283].

@x110dc : Your opinion?

apps/contributions/views.py
harsha09: In this case there will be a new endpoint right @benjaminEwhite?

apps/contributions/stripe_contributions_provider.py
harsha09: This doesn't override the `id`built-in function
image

apps/contributions/serializers.py
harsha09: Added `ChoiceField` for status, payment_type, interval and card_brand.

apps/contributions/stripe_contributions_provider.py
harsha09: changed it to `ContributionIgnorableError, LMK if this doesn't sounds good

revengine/settings/dev.py
harsha09: It's there in `deploy.py`, not sure why it's so.

apps/contributions/stripe_contributions_provider.py
harsha09: I've moved it to `base.py`.
`# Set cache TTL to magic link token lifetime so that the stipe contributions in cache will expire with it and contributions will again pulled from stripe and stored in cache with the magic link request.
CONTRIBUTION_CACHE_TTL = CONTRIBUTOR_LONG_TOKEN_LIFETIME`

apps/contributions/stripe_contributions_provider.py
harsha09: We can get revenue program slug from https://github.com/newsrevenuehub/rev-engine/blob/main/apps/api/views.py#L153 but how do we get revenue program slug in https://github.com/newsrevenuehub/rev-engine/blob/main/apps/contributions/views.py#L258

apps/contributions/views.py
benjaminEwhite: Yeah, that's what I'm imagining. Contributor portal would be the consumer of this endpoint.

apps/contributions/stripe_contributions_provider.py
x110dc: Like I said, we shouldn't use an identifier for the RP, we should use something to identify the Stripe account itself (which can be inferred from the RP). 

As for how you get that, can't you retrieve it from the `request` object? @benjaminEwhite would know.

apps/contributions/stripe_contributions_provider.py
x110dc: To be clear, I think there are 3 TTLs here:
- the cache
- the magic link (currently hardcoded to 15m)
- the session created by the magic link (currently hardcoded to 3h)

So, as long as the cache is at least as long as the magic link then their results should be there when they visit the link after requesting it. 

For now, I think we could store empty results and return those to the FE. But if the user makes a request and there's nothing in the cache we should fetch new results (and cache them).  I don't think we need to worry at this point about the edge case given we state a disclaimer on the page. 

Let's create a variable for this TTL that we can tune. We can log when there are no cache hits and tune later. A good default could be 30m for now.

revengine/settings/dev.py
x110dc: Maybe we just expect `redis://` URLs in `dev.py`.

revengine/settings/dev.py
x110dc: I'd think this should go in `base.py` unless @benjaminEwhite knows otherwise.

apps/api/views.py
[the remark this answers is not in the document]
x110dc: I agree it should be a celery background task. Otherwise it will slow down the the sending of the magic link.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,179 @@
...
+        return self.charge.id
+
+
+class StripeContributionsProvider:
+    def __init__(self, email_id):
+        self.customers = self.fetch_all_customers(email_id)
+
+    @staticmethod
+    def fetch_all_customers(email_id):
+        results = stripe.Customer.search(query=f"email:'{email_id}'", limit=MAX_STRIPE_RESPONSE_LIMIT)
```
x110dc, line 144: I'm confused about which Stripe account this is fetching data from. Where is that being set?

apps/api/views.py
```diff
@@ -154,6 +157,11 @@ def post(self, request, *args, **kwargs):
 
             if not domain:
                 return Response({"detail": "Missing Revenue Program subdomain"}, status=status.HTTP_404_NOT_FOUND)
+
+            # Pull contributions from Stripe and store the serialized data in cache whose TTL is configured with the
+            # magic_link life time
+            pull_serialized_stripe_contributions_to_cache(email)
```
x110dc, line 163: The RP or Stripe account should be a parameter on this so you know which Stripe account to fetch data from.

revengine/settings/dev.py
harsha09: moved it to base.py

apps/contributions/stripe_contributions_provider.py
harsha09: I made it 30m

apps/contributions/stripe_contributions_provider.py
harsha09: @x110dc : I'm not able to get that from request object. I've added I've added https://github.com/newsrevenuehub/rev-engine/pull/459/files#diff-58df16d1548a5310067158c74972c58a99aed8e6b29d88808ca3e9bd6e07e18aR65 to get revenue program from query params

benjaminEwhite:
Closing and reopening to trigger review app creation.

benjaminEwhite (changes requested):
Left a bunch of inline comments. 

As a general comment, doc strings would be helpful on a lot of the methods and functions, because their purpose and context is not always obvious, and this will only become more so the more time passes.

Running locally, I was able to see confirmation and magic link emails sending (printing to terminal), and contributions showed up in db/django admin. Magic link emails were expired on arrival, so wasn't able to check out that side of things.

In review app environment, was able to make contribution, but didn't appear in db -- probably a problem with webhooks?  I also wasn't able to get confirmation emails or magic link emails to appear, but maybe that's because emails don't send for review apps???

revengine/settings/dev.py
harsha09: @benjaminEwhite 
It's useful to switch the broker between Redis (For Dev environment) or a DummyCache (For UnitTests).
Otherwise UTs are failing due unavailability of Redis.
LMK if there is a different way to approach this problem.

apps/contributions/tests/test_views.py
```diff
@@ -628,6 +598,54 @@ def test_unexpected_role_type(self):
...
 
 
+class TestContributorContributionsViewSet(RevEngineApiAbstractTestCase):
+    def setUp(self):
+        super().setUp()
+        self.org = OrganizationFactory()
+        self.payment_provider1 = PaymentProviderFactory()
+        self.payment_provider2 = PaymentProviderFactory()
+        self.rp1 = RevenueProgramFactory(organization=self.org, payment_provider=self.payment_provider1)
+        self.rp2 = RevenueProgramFactory(organization=self.org, payment_provider=self.payment_provider2)
```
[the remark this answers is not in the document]
harsha09: @benjaminEwhite 
I was using `.set_up_domain_model` but raised duplicate email exception and few other issues.
And moreover `.set_up_domain_model` will setup lot of things which is not useful for my code, but I forgot to remove inheriting from `AbstractTestCase`.
Now I've changed it to `APITestCase`.

apps/contributions/views.py
```diff
@@ -284,14 +286,38 @@ class ContributionsViewSet(viewsets.ReadOnlyModelViewSet, FilterQuerySetByUserMi
     filterset_class = ContributionFilter
     filter_backends = [DjangoFilterBackend, filters.OrderingFilter]
 
+    def get_object(self):
+        if isinstance(self.request.user, Contributor):
```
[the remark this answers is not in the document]
harsha09: Got an error due to one PR merged with main and it was returning Contributor object, so in order to resolve the issue I've added this code to return dict if it's a contributor.
I don't exactly remember which PR was that.
I've re-visited it now and there is no error if I remove this code.
Hence removed.

apps/contributions/serializers.py
```diff
@@ -463,3 +465,23 @@ class StripeRecurringPaymentSerializer(AbstractPaymentSerializer):
...
 
     payment_method_id = serializers.CharField(max_length=255)
+
+
+class PaymentProviderContributionSerializer(serializers.Serializer):
+    """
+    Payments provider serializer, payment provider Eg: Stripe.
+    """
+
+    id = serializers.CharField(max_length=255)
```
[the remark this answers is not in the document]
harsha09: It's a stripe object id, in this case it's a charge object and it will start with `ch_*`.
Here is the doc I found out to be helpful https://stripe.com/docs/upgrades#what-changes-does-stripe-consider-to-be-backwards-compatible.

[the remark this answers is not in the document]
harsha09: I've added the comment as well.

apps/api/views.py
```diff
@@ -154,6 +155,12 @@ def post(self, request, *args, **kwargs):
 
             if not domain:
                 return Response({"detail": "Missing Revenue Program subdomain"}, status=status.HTTP_404_NOT_FOUND)
+
+            revenue_program = RevenueProgram.objects.get(slug=data.get("subdomain"))
```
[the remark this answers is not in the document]
harsha09: @benjaminEwhite 
We are taking revenue program slug from subdomain, I assume subdomain will not form for an rp which doesn't exist.
Still considering a scenario where url is malformed, this might happen.
I've added the code which will raise 404 error if not found.

apps/contributions/migrations/0003_alter_contribution_status.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.13 on 2022-07-04 21:06
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("contributions", "0002_contribution_donation_page"),
```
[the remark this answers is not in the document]
harsha09: Done

apps/contributions/models.py
```diff
@@ -58,6 +58,39 @@ class ContributionStatus(models.TextChoices):
...
+    EPS = "eps", "eps"
+    GIROPAY = "giropay", "giropay"
+    IDEAL = "ideal", "ideal"
+    KLARNA = "klarna", "klarna"
+    MULTIBANCO = "multibanco", "multibanco"
+    P24 = "p24", "p24"
+    SEPA_DEBIT = "sepa_debit", "sepa_debit"
+    SOFORT = "sofort", "sofort"
+    STRIPE_ACCOUNT = "stripe_account", "stripe_account"
+    WECHAT = "wechat", "wechat"
```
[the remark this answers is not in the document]
harsha09: I just followed `ContributionStatus`. Modified `CardBrand` and `PaymentType`, LMK if `ContributionStatus` also can be modified.

apps/contributions/serializers.py
```diff
@@ -463,3 +465,23 @@ class StripeRecurringPaymentSerializer(AbstractPaymentSerializer):
...
+    status = serializers.ChoiceField(choices=ContributionStatus.choices)
+    card_brand = serializers.ChoiceField(choices=CardBrand.choices, required=False, allow_null=True)
+    last4 = serializers.IntegerField()
+    payment_type = serializers.ChoiceField(choices=PaymentType.choices, required=False, allow_null=True)
+    next_payment_date = serializers.DateTimeField()
+    interval = serializers.ChoiceField(choices=ContributionInterval.choices)
+    revenue_program = serializers.CharField(max_length=255)
```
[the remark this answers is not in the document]
harsha09: Thank you for reminding this. RevenueProgram.name is of 255 char length in DB schema. We are considering only RevenueProgram.slug here and it's 63 char length.

apps/contributions/serializers.py
```diff
@@ -463,3 +465,23 @@ class StripeRecurringPaymentSerializer(AbstractPaymentSerializer):
...
+    amount = serializers.IntegerField()
+    provider_customer_id = serializers.CharField(max_length=255)
```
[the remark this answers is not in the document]
harsha09: same applies here like charge object id

apps/contributions/serializers.py
```diff
@@ -463,3 +465,23 @@ class StripeRecurringPaymentSerializer(AbstractPaymentSerializer):
...
+    credit_card_expiration_date = serializers.CharField(max_length=30)
```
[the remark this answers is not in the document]
harsha09: mm/yyyy - keeping it 7

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,210 @@
...
+class StripeCharge:
+    """
+    Wrapper on stripe charge object to extract the required details in
+    apps.contributions.serializers.PaymentProviderContributionSerializer and serializable.
+    """
+
+    def __init__(self, charge):
+        if not hasattr(charge, "invoice") or not charge.invoice:
+            raise NoInvoiceGeneratedError(f"No invoice object for charge : {charge.id}")
+        self.charge = charge
```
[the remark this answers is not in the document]
harsha09: Yeah it violates DRY, made it a common prop

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,210 @@
...
+
+    @property
+    def interval(self):
+        interval = self.charge.invoice.lines.data[0].plan.interval
```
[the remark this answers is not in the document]
harsha09: To my knowledge I've not seen an invoice without line item. I added the check to make it more robust.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,210 @@
...
+                logger.exception("This can be ignored")
+        return data
+
+    def upsert(self, contributions):
+        data = self.serialize(contributions)
+        cached_data = self.cache.get(self.key)
+
+        if cached_data:
+            cached_data = json.loads(cached_data)
+            data.update(cached_data)
```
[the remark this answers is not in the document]
harsha09: It will not because we are updating the data here.
If you visualise the data will be like
{
    "<email-id>-<stripe_account_id>": {
         "ch_000000": {<data>},
         "ch_000001": {<data>},
    }
}

apps/contributions/tests/test_stripe_contributions_provider.py
```diff
@@ -0,0 +1,188 @@
...
+        self.assertEqual(stripe_charge.refunded, False)
+        self.assertEqual(stripe_charge.id, "charge_1")
+
+        self.charge["invoice"]["lines"]["data"][0]["plan"]["interval"] = "month"
+        stripe_charge = StripeCharge(self.charge)
+        self.assertEqual(stripe_charge.interval, ContributionInterval.MONTHLY)
+
+        self.charge["invoice"]["lines"]["data"][0]["plan"]["interval"] = "day"
+        with self.assertRaises(InvalidIntervalError):
+            StripeCharge(self.charge).interval
```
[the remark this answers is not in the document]
harsha09: Most primarily to increase UT coverage, otherwise it's failing.
Considering positive and negative cases and testing the interval as well.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        interval = self.charge.invoice.lines.data[0].plan.interval
+        interval_count = self.charge.invoice.lines.data[0].plan.interval_count
+        if interval == "year" and interval_count == 1:
+            return ContributionInterval.YEARLY
+        if interval == "month" and interval_count == 1:
+            return ContributionInterval.MONTHLY
+        raise InvalidIntervalError(f"Invalid interval {interval} for charge : {self.charge.id}")
+
+    @property
+    def revenue_program(self):
```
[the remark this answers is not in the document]
harsha09: In the context of contributions revenue_program is slug and we are not using revenue program name anywhere.

benjaminEwhite (approved):

benjaminEwhite (approved):

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,236 @@
...
+            query=f"email:'{self.email_id}'",
+            limit=MAX_STRIPE_RESPONSE_LIMIT,
+            stripe_account=self.stripe_account_id,
+        )
+        return [customer.id for customer in customers_response.auto_paging_iter()]
+
+    @staticmethod
+    def generate_chunks(data, size):
+        for i in range(0, len(data), size):
+            yield data[i : i + size]
```
njharman, line 176: Only used in generate_chunked_customers_query I believe.  I would have put it inside of that method as a plain function. Or really, just as outer for loop.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,236 @@
...
+
+    @property
+    def payment_type(self):
+        return self.charge.payment_method_details.type
+
+    @property
+    def next_payment_date(self):
+        next_attempt = self.charge.invoice.next_payment_attempt
+        if next_attempt:
+            return datetime.utcfromtimestamp(int(next_attempt))
```
njharman, line 134: I'd expect a date, not datetime from something called next_payment_date.  But, maybe the use case is to return datetime, despite name?

apps/contributions/views.py
```diff
@@ -285,13 +288,37 @@ class ContributionsViewSet(viewsets.ReadOnlyModelViewSet, FilterQuerySetByUserMi
...
+                )
+            revenue_program = get_object_or_404(RevenueProgram, slug=rp_slug)
+            cache_provider = ContributionsCacheProvider(self.request.user.email, revenue_program.stripe_account_id)
+
+            contributions = cache_provider.load()
+            # trigger celery task to pull contributions and load to cache if the cache is empty
+            if not contributions:
+                task_pull_serialized_stripe_contributions_to_cache.delay(
+                    self.request.user.email, revenue_program.stripe_account_id
+                )
```
njharman, line 306: Should we not return some sort of message to user (maybe SPA does?). 
Or, populate cache synchronically here only on cache miss (delay HTTP response until data retrived)? 

Instead of returning empty contributions. which just seems it would confuse user.

revengine/settings/base.py
```diff
@@ -165,6 +169,29 @@
     DATABASES["default"].update(db_from_env)
 
 
+REDIS_URL = os.getenv("REDIS_TLS_URL", os.getenv("REDIS_URL", "redis://redis:6379"))
+CACHE_HOST = REDIS_URL
+CONNECTION_POOL_KWARGS = {}
+if CACHE_HOST.startswith("rediss"):
```
njharman, line 175: is the misspelling "rediss" two esses, correct?

revengine/settings/base.py
```diff
@@ -126,6 +127,9 @@
 
 USER_TTL = timedelta(hours=24)
 
+# Contributions cache set to 30 minutes < CONTRIBUTOR_LONG_TOKEN_LIFETIME
+CONTRIBUTION_CACHE_TTL = timedelta(minutes=30)
+CONTRIBUTION_CACHE_DB = "default"
```
njharman, line 132: Believe we should use specific name, "contribution" or "stripe" rather than "default".

njharman (changes requested):
tests pass, but under 95% will have to address that. looking at html coverage report 

- org.models.stripe_account_id lacks coverage
- strijp_contributions_provider has some easy to fill coverage gaps
- api/view missing coverage if no RP subdomain

Made few comments / suggestions. Nothing that would "block" approval.

revengine/settings/base.py
harsha09: It's right, redis url will start with `rediss://<host>:<port>` to communicate over ssl.
This was already there in `deploy.py`, I just moved it to `base.py`.

revengine/settings/base.py
harsha09: We are using Redis as default cache provider and only one instance, this confusion might be due to the naming.
I'm going to rename it to `DEFAULT_CACHE` which is more accurate instead of `CONTRIBUTION_CACHE_DB`.

apps/contributions/stripe_contributions_provider.py
harsha09: Yeah it returns a datetime field just like `last_payment_date = models.DateTimeField(null=True)` in contributions but this is future payment date.

apps/contributions/stripe_contributions_provider.py
harsha09: You are right, not using this anywhere else.

apps/contributions/views.py
x110dc: Can't we just do this call synchronously as Norm suggests if there's a cache miss? It won't be fast but at least they'll get the data. Can just remove the `.delay`, no?

harsha09: @x110dc 
`task_pull_serialized_stripe_contributions_to_cache` will call `task_pull_charges` asynchronously for a set of 10 (Stripe's MAX customer query values) customers at a time and doesn't return anything.

It'll need some changes to the tasks to make it happen, will it not beat the purpose of a celery task?

harsha09: @x110dc 
This will also make the response very slow because all calls will be asynchronous as we need to fetch customers first and then corresponding charges.
Instead how about moving this strategy to UI by show loading icon until there is some data in cache with a hard limit? - This doesn't take much time because the cache will be populated incrementally.

njharman: The syncrhously was only on the edge case of cache miss. Idea is, a slow response is better than no/empty response. But, I'm also worrying about a what if that may never happen enough to justify effort.

We have ticket DEV-2092 to look at this with metrics. If we notice problem before then we can fix then.

x110dc (approved):

PR #460, opened 2022-06-13
[WIP] DEV-1776 - add pytest-xdist and parallelize in make-tests

benjaminEwhite:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ run-redis:
 run-tests:
 	@echo 'Checking for migrations'
 	python manage.py makemigrations --dry-run --check
-	pytest
+	pytest -n auto
 
 check-dc:
 	docker-compose -f docker-compose.yml -f docker-compose-dev.yml ps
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -39,7 +39,7 @@ run-redis:
 run-tests:
 	@echo 'Checking for migrations'
 	python manage.py makemigrations --dry-run --check
-	pytest
+	pytest -n auto
```
njharman, line 42: In code (vs command line) I prefer the long versions of arguments (for self documentation) 

> pytest --numprocesses auto

njharman, line 42: xdist config should also be added to pytest.ini

njharman (changes requested):
Comparing other build **6m** https://github.com/newsrevenuehub/rev-engine/runs/6910837384?check_suite_focus=true
With build time for this branch **4m41s** is something.  I wonder how many "cores" CI actually has/reports.

But running locally, with all the cores!, it's **1m46s** vs **4m 45s** Booyah!

x110dc (approved):
I made one minor change by unpinning `pytest-xdist` in the `pyproject.toml`. I prefer that dev dependencies not be pinned here so Dependabot and poetry will suggest upgrades and since `poetry.lock` pins them.

PR #466, opened 2022-06-14
DEV-1186 - Only record object changes when triggered by a user action

benjaminEwhite:

##### The problem:

Presently, rev-engine relies on [django-simple-history](https://django-simple-history.readthedocs.io/en/latest/) in order to provide an audit log of changes to model instances.  The primary use case for this functionality is to enable admin users to monitor and at times revert user triggered actions to a model instance. For instance, if an org admin changes the layout for a live page through the org dashboard, or if an admin deletes a contribution in the Django admin, those changes are recorded.

However, `django-simple-history` tracks all changes to model instances, including those generated by system level events like webhooks or signals. In practice, these audit logs end up drowning out the signal (which is user-generated changes), making our object-level audit logs less useful than they might otherwise be. There is not an "off the shelf" way to make `django-simple-history` only record user generated changes.

##### The solution:

In order to clean up object level audit log to make it useful, we should retire [django-simple-history](https://django-simple-history.readthedocs.io/en/latest/) in favor of [django-reversion](https://django-reversion.readthedocs.io/en/stable/).

`django-reversion` can be registered on particular model admin instances and at the view level, which will allow us to only add revision history objects for the limited cases we care about (I've already got a proof of concept of this running locally, and have confirmed that changes through org dashboard and django admin cause new revisions, while updating model through `python manage.py shell` does not). 

`django-reversion` will also nicely support [DEV-86 - Hub admin can revert versions of a donation page](https://news-revenue-hub.atlassian.net/browse/DEV-86). It does so out of the box, and there is also a related package ([django-reversion-compare](https://github.com/jedie/django-reversion-compare)) that provides an interface that allows a user to more easily compare the differences between any given two revisions.

*Additionally*, I propose that we retire [django-safedelete](https://github.com/makinacorpus/django-safedelete), as django-reversion will give allow us to revert both revisions and deletions.  This will allow us to consolidate audit logs and reversion of deletions into a single package and admin utility.

##### Implications for migrations

I've discovered that there is some complexity to removing django-simple-history in that there are several data migrations that explicitly import `simple-history`, and those migrations will fail if that package is not available.  I believe that there has been discussion about zeroing out our migrations and freshly creating them all from current state of models. This would be a good moment to do that to avoid some busy work removing the simple-history related migrations (which would involve removing those migrations then updating any migrations that point to them as predecessors). 

**UPDATE**: It looks like if you retain django-simple-history as a dependency, those migrations don't cause a problem. So we could also choose to simply retain that dependency even though our app no longer really needs it. All things equal, I'd prefer not to keep a dependency around just because it's referenced in some migrations, but my inner pragmatist definitely wins out vs. whatever religious objection I have here.

Outside of data migrations related to django-simple-history, I only see one additional data migration (apps/contributions/migrations/0007_auto_20210517_1852.py).  I haven't wrapped my head around why that one was needed, but it would be easy enough to retain it when wiping out all other migrations, if need be. 

#### What's this PR do?

- Adds a section to README.md documenting django-reversion usage. Additionally does a fair amount of cleanup of README.md and adds table of contents at top of file.
- Removes all references to `django-simple-history`, but for now retains as dependency because removing causes old data migrations to fail.
- Removes `django-safedelete` and all references to it in favor of `django-reversion`'s deletion recovery capabilities.
- Retires `RevEngineSimpleHistoryAdmin`, `RevEngineHistoricalChangeAdmin`, and `RevEngineHistoricalChange`
- Removes the `apps.common.signals.py` file, as it's only function was a callback pertained to `django-simple-history`.
- Removes: apps/common/tests/factories.py and apps/common/tests/test_admin.py as they are no longer needed.
- Adds `reversion.admin.VersionAdmin` as base class for model admin classes that formerly used `RevEngineSimpleHistoryAdmin`.  NB: registering a model admin with django-reversion causes its respective model to be registered (but only changes undertaken through admin will be recorded).
- Removes `SafeDeleteModel` from `DonationPage` and `Style`, as deletion recovery comes for free with `django-reversion`.
- Adds the `RevsionMixin` from django-reversion to `PageViewSet`, `TemplateViewSet`, and `StyleViewSet` so that API-generated write operations to their respective models will be captured by django-reversion.
- Adds `revengine.tests.test_django_reversion_setup.py`, which guarantees that expected model admins and views are hooked up with `django-reversion`, while demonstrating that the same respective models *do not* yield reversion instances when changes are made directly through `.save()` outside of API/admin context.
- Changes `bootstrap-review-app.py` to update all config vars at once, instead of piecemeal, as the old approach triggered 3-4 separate rebuilds (one for each config var individually set).
- Updates `scripts/heroku-postdeploy.sh` to drop tables before dumping/restoring in order to prevent vexing migration errors in review apps (explained in detail in comment in that file).
- Removes test parallelization in CI. `PageViewSetTest` and `TemplateViewSetTest` started having some failures in CI. Across successive runs, different tests would fail. This appears to stem from move to test parallelization. I tried to run down these problems now, but determined it was a bigger effort, and shouldn't block us here. I've created [DEV-1983 | Refactor Python tests to allow for parallelization again](https://news-revenue-hub.atlassian.net/browse/DEV-1983) to track this.

Described at length in implementation plan at top.

- Edit or delete a model instance in the Django admin, save changes, then for that same instance, click on the 'History` button in the top right hand corner of the admin interface. You should see a record for each save you have made.
- Edit a style, page, or template from the dashboard app. Then go look at the respective model instance in the Django admin. Click on its history, and you should see the revisions reflected there.

Yes. See `revengine.tests.test_django_reversion_setup.py`

N/A

No.

Yes. Added a section to `README.md` about use of `django-reversion`, inclusive of need to run `python manage.py createinitialrevisions` after registering a new model with `django-reversion`.

[DEV-1186](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-1186)

[DEV-86](https://news-revenue-hub.atlassian.net/browse/DEV-86)

[DEV-1223](https://news-revenue-hub.atlassian.net/browse/DEV-1223)

On initial merge of this PR, and on any subsequent registration of a model with django-reversion, there is a management command that needs to be run. [See here](https://django-reversion.readthedocs.io/en/stable/commands.html#createinitialrevisions).

I've already done that in the current review app, fyi.

No.


pyproject.toml
```diff
@@ -62,16 +62,16 @@ django-solo = "^1.1.5"
...
 django-health-check = "^3.16.4"
 pycountry = "^20.7.3"
 django-csp = "^3.7"
 cloudflare = "^2"
 heroku3 = "^5"
 drf-extra-fields = "^3"
-django-simple-history = "3.0.0"
 django-waffle = "^2.4.1"
+django-reversion = "^5.0.0"
+django-simple-history = "^3.1.1"
```
njharman, line 74: Deserves comment here why simple-history kept and when it can be removed.

njharman (changes requested):
Would like the [createinitialrevisions ](https://django-reversion.readthedocs.io/en/stable/commands.html#createinitialrevisions) command mentioned in README. That's something we will forget I bet.

Could not get migrations to apply after 
> git co main
> poetry install
> make nuclear
> docker-compose up -d
> ./manage migrate
> ./manage createsuperuser
> git co DEV-1186...
> poetry install
> ./manage migrate
>
>  django.db.utils.ProgrammingError: column "organization_id" of relation "contributions_historicalcontribution" does not exist

But wiping db, and doing migrations from DEV-1186 works.

benjaminEwhite:
@njharman I did a cleanup of `README.md` and then added a section giving all the details re: django-reversion.

njharman (approved):
lgtm

benjaminEwhite:
Closing in order to destroy current review so I can get a fresh one to investigate the migration error happening in CI on build/deploy for review app.

benjaminEwhite:
Closing and reopening again because want to confirm that change i'm making to bootstrap-review-app.py works as expected.

PR #470, opened 2022-06-19
dev-1920 dashboard side-bar and top-bar

gurjittoptal:

Updates side-nav, top-nav as new dashboard designs

#### What's this PR do?
Changes side-nav, adds top-nav and hides both for edit/create page

to update dashboard ui

by logging into dashboard and checking all links ( pages/ styles/ contributions/ logout ) work and the site is responsive

no

n/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-1920


spa/src/components/dashboard/topbar/DashboardTopbar.js
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
+    <S.DashboardTopbar>
+      <S.TopLogo>
+        <S.Logo src={logo} />
+      </S.TopLogo>
+      <S.TopLogoMobile>
+        <S.Logo src={mobileLogo} />
+      </S.TopLogoMobile>
+      <S.TopMenu onClick={logout} whileHover={{ scale: 1.05, x: -3 }} whileTap={{ scale: 1, x: 0 }}>
```
benjaminEwhite, line 21: This element needs a margin-right (or else parent needs padding). The sign out text gets cut off.

Screen Shot 2022-06-21 at 9 24 16 AM

benjaminEwhite (changes requested):
Left a few inline comments.

Some additional notes:

- The comps call for Roboto font for nav elements, but this is using Montserrat
- Clicking on the signout takes to root and 404. I'm pretty sure that's pre-existing, but might be worth changing behavior so logout redirects to login as long as some of the log out code is being touched anyways.
- Probably should have follow up tickets for descoped features. Namely, the top bar elements that were mentioned in the JIRA ticket as being descoped for now.
- I suspect this is a temporary issue, but when clicking on a style or  page, when you get taken to the edit view, there's a flash of white content, and it's pretty abrupt. If there's an easy way to animate that transition so it feels smoother, that would be nice.
- There should be some unit tests, at a minimum describing what elements we expect to find in the side and top bar, and demonstrating their expected configuration (for instance, that the href for contributions is the contributions URL, etc).

spa/src/components/dashboard/Dashboard.js
```diff
@@ -52,45 +53,50 @@ function Dashboard() {
...
-              ) : null}
-              {hasContentSectionAccess ? (
-                <Route path={CUSTOMIZE_SLUG}>
-                  <Customize />
-                </Route>
-              ) : null}
-              {hasContentSectionAccess ? (
-                <Route path={EDITOR_ROUTE_PAGE}>
-                  <PageEditor />
+    <div>
```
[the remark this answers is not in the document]
gurjittoptal: can't use fragment here so instead used an outer element for consistency, to handle styling later.

gurjittoptal:
> Left a few inline comments.
> 
> Some additional notes:
> 
>     * The comps call for Roboto font for nav elements, but this is using Montserrat
yes, i did not change as it will be a follow up ticket. Likely we change the site-wide default font and this fixes
> 
>     * Clicking on the signout takes to root and 404. I'm pretty sure that's pre-existing, but might be worth changing behavior so logout redirects to login as long as some of the log out code is being touched anyways.
PR for this was already merged to main. 
> 
>     * Probably should have follow up tickets for descoped features. Namely, the top bar elements that were mentioned in the JIRA ticket as being descoped for now.
https://news-revenue-hub.atlassian.net/browse/DEV-1950 handles the next steps
> 
>     * I suspect this is a temporary issue, but when clicking on a style or  page, when you get taken to the edit view, there's a flash of white content, and it's pretty abrupt. If there's an easy way to animate that transition so it feels smoother, that would be nice.
hmm. i see that and shall fix it in a follow up of DEV-1950. issue is updated
> 
>     * There should be some unit tests, at a minimum describing what elements we expect to find in the side and top bar, and demonstrating their expected configuration (for instance, that the href for contributions is the contributions URL, etc).
so, right now all elements are exactly like they were earlier. but, yes we need to refactor a bit for more unit tests.

x110dc (dismissed):
Hi @gurjittoptal can you add the unit tests that @benjaminEwhite requested. Thanks!

spa/src/components/dashboard/sidebar/DashboardSidebar.js
```diff
@@ -17,50 +18,23 @@ function DashboardSidebar({ shouldAllowDashboard }) {
...
-        ) : null}
+      <S.NavList role="list" data-testid="nav-list" aria-labelledby="sidebar-label-id">
+        <S.NavItemLabel>
+          <S.NavItemIcon icon={ICONS.DASHBOARD} />
+          <S.SideBarText id="sidebar-label-id">Dashboard</S.SideBarText>
+        </S.NavItemLabel>
+
+        <ContentSectionNav
+          hasContentSectionAccess={hasContentSectionAccess}
+          shouldAllowDashboard={shouldAllowDashboard}
```
benjaminEwhite, line 31: Also here, I don't think we want `ContentSectionNav` having a notion of dashboard access. Instead it should only care about whether or not it's in a disabled state. Indirectly, it may be in a disabled state because of dashboard access state, but in the future, it could be some other condition that might cause it to be in disabled state.

spa/src/components/dashboard/sidebar/DashboardSidebar.styled.js
```diff
@@ -2,60 +2,149 @@ import styled from 'styled-components';
 import { NavLink } from 'react-router-dom';
 import SvgIcon from 'assets/icons/SvgIcon';
 import { motion } from 'framer-motion';
+import { ICONS } from 'assets/icons/SvgIcon';
```
benjaminEwhite, line 5: `ICONS` is not used anywhere and should be removed.

spa/src/components/dashboard/sidebar/navs/ContentSectionNav.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import * as S from './../DashboardSidebar.styled';
+import { CONTENT_SLUG, CUSTOMIZE_SLUG } from 'routes';
+import { ICONS } from 'assets/icons/SvgIcon';
+
+function ContentSectionNav({ hasContentSectionAccess, shouldAllowDashboard }) {
```
benjaminEwhite, line 6: Same comment here about `shouldAllowDasbhoard` — think it should be `disabled`.

spa/src/components/dashboard/sidebar/navs/ContributionSectionNav.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import * as S from './../DashboardSidebar.styled';
+import { DONATIONS_SLUG } from 'routes';
+import { ICONS } from 'assets/icons/SvgIcon';
+
+function ContributionSectionNav({ hasContributionsSectionAccess, shouldAllowDashboard }) {
```
benjaminEwhite, line 6: Same comments re: `hasContributionSectionAccess` and `shouldAllowDashboard`. The former logic should live in higher order component, and `shouldAllowDashboard` should be renamed to something more semantic like `disabled`.

spa/src/components/dashboard/sidebar/navs/ContributionSectionNav.test.js
```diff
@@ -0,0 +1,17 @@
...
+  render(<ContributionSectionNav hasContributionsSectionAccess={true} shouldAllowDashboard={true} />);
+  const contributionsLink = screen.queryByTestId(CONTRIBUTIONS_TEST_ID);
+  expect(contributionsLink).toBeInTheDocument();
+});
+
+it('should not show contributions link if hasContributionsSectionAccess is false', () => {
+  render(<ContributionSectionNav hasContributionsSectionAccess={false} shouldAllowDashboard={true} />);
+  const contributionsLink = screen.queryByTestId(CONTRIBUTIONS_TEST_ID);
+  expect(contributionsLink).not.toBeInTheDocument();
+});
```
benjaminEwhite, line 17: Same comment here -- the relevant test would be enabled vs. disabled state, and logic for whether or not its displayed should live higher up.

benjaminEwhite (commented):
This is moving in good direction. Left some comments, but also wanted to reply to comment you left in JIRA here.

> However i did keep the hasAccess in the smaller functions for a few reasons. First, we can unit-test the featureFlag tests much more easier, and secondly the code is more readable.
>
> in the future there will be at least 3 more components inserted and this way we can isolate and test while keeping the sidebar code compact and clean.
>
> ```
><S.DashboardSidebar>
>	<S.NavList role="list" data-testid="nav-list" aria-labelledby="sidebar-label-id">
>		...
>		<DashboardHomeNav ...  />
>		<ContentSectionNav ... />
>		<ContributionSectionNav ... />
>		<AnalyticsNav ...  />
>		<ColorThemeNav ...  />
>		...
>	</S.NavList>
></S.DashboardSidebar>
> ```

In general, unit tests only get more as components further down the tree get smarts in them. For the case of unit testing around feature flags in particular, as currently implemented, an isolated test of `ContentSectionNav` when `hasContentSectionAccess` is set to false actually doesn't prove the feature flag is correctly implemented / hasn't reverted. All it demonstrates is that the component respects an attribute called `hasContentSectionAccess`, which really should be called `shouldDisplay`.  The parent context for `ContentSectionNav` may or may not set that attribute with respect to feature flag state.  Which means, if you want to build confidence about the business logic around content section access insofar as it effects the sidebar, that test should be on the sidebar element itself, which is what makes the call to `flagIsActiveForUser`.

Regarding readability, nothing requires putting a bunch of conditionals in the JSX. In the example you provided, you could instead do something like this (in this example I'm assuming that each sidebar components:

```
const sidebarElements = [
    [<DashboardHomeNav foo={bar} />, true],
    [<ContentSectionNav disabled={disabled} />, flagIsActiveForUser(CONTENT_SECTION_ACCESS_FLAG_NAME, featureFlags)],
    [<ContributionSectionNav disabled={disabled} />, flagIsActiveForUser(CONTRIBUTIONS_SECTION_ACCESS_FLAG_NAME, featureFlags)],
    // ...etc.
  ].filter(([section, shouldDisplay]) => shouldDisplay);

  return (
    <S.DashboardSidebar>
      <S.NavList role="list" data-testid="nav-list" aria-labelledby="sidebar-label-id">
        {/* ... */}
        {sidebarElements}
```

spa/src/components/dashboard/Dashboard.js
```diff
@@ -52,45 +53,50 @@ function Dashboard() {
...
-              {hasContentSectionAccess ? (
-                <Route path={CUSTOMIZE_SLUG}>
-                  <Customize />
-                </Route>
-              ) : null}
-              {hasContentSectionAccess ? (
-                <Route path={EDITOR_ROUTE_PAGE}>
-                  <PageEditor />
+    <S.Outer>
+      {isEditPage ? null : <DashboardTopbar shouldAllowDashboard={getShouldAllowDashboard()} />}
```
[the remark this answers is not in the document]
gurjittoptal: Fair Point.

See this
https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/dashboard/Dashboard.js#L61

As we have moved logout to topbar, sidebar does not need to know of getShouldAllowDashboard(), because none of pages/styles/contribution routes are allowed if !getShouldAllowDashboard()

So all navs become jsx functions with no inputs/ conditional logic.

spa/src/components/dashboard/sidebar/DashboardSidebar.styled.js
gurjittoptal: removed

spa/src/components/dashboard/sidebar/navs/ContentSectionNav.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+it('should not show pages and styles link if user does not haveContentSectionAccess', () => {
+  render(<ContentSectionNav hasContentSectionAccess={false} shouldAllowDashboard={true} />);
+
+  const pagesLink = screen.queryByTestId(PAGES_TEST_ID);
+  expect(pagesLink).not.toBeInTheDocument();
+
+  const stylesLink = screen.queryByTestId(STYLES_TEST_ID);
+  expect(stylesLink).not.toBeInTheDocument();
+});
```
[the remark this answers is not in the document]
gurjittoptal: agreed. its made into a totally dumb one with just jsx

spa/src/components/dashboard/sidebar/navs/ContentSectionNav.js
gurjittoptal: not relevant anymore as these navs are just jsx blocks

spa/src/components/dashboard/Dashboard.js
```diff
@@ -52,45 +53,50 @@ function Dashboard() {
...
+      <S.Dashboard data-testid="dashboard">
+        {isEditPage ? null : <DashboardSidebar shouldAllowDashboard={getShouldAllowDashboard()} />}
```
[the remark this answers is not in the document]
gurjittoptal: See above comment

spa/src/components/dashboard/sidebar/navs/ContentSectionNav.js
```diff
@@ -0,0 +1,35 @@
...
+        aria-labelledby="customize-nav-item-id"
+        role="listitem"
+        data-testid="nav-styles-item"
+        to={CUSTOMIZE_SLUG}
+        disabled={!shouldAllowDashboard}
+      >
+        <S.NavItemIcon icon={ICONS.CUSTOMIZE} />
+        <S.SideBarText id="customize-nav-item-id">Customize</S.SideBarText>
+      </S.NavItem>
+    </S.NavSection>
```
benjaminEwhite, line 31: Given recent commit, `shouldAllowDashboard` references should be removed here.

spa/src/components/dashboard/sidebar/navs/ContributionSectionNav.js
```diff
@@ -0,0 +1,25 @@
...
+        aria-labelledby="contributions-nav-item-id"
+        role="listitem"
+        data-testid="nav-contributions-item"
+        to={DONATIONS_SLUG}
+        disabled={!shouldAllowDashboard}
+      >
+        <S.NavItemIcon icon={ICONS.CONTRIBUTIONS} />
+        <S.SideBarText id="contributions-nav-item-id">Contributions</S.SideBarText>
+      </S.NavItem>
+    </S.NavSection>
```
benjaminEwhite, line 21: Also should remove `shouldAllowDashboard` here.

spa/src/components/dashboard/sidebar/DashboardSidebar.test.js
```diff
@@ -0,0 +1,33 @@
...
+  useFeatureFlagsProviderContext: () => ({ featureFlags: [] })
+}));
+
+jest.mock('utilities/flagIsActiveForUser', () => ({
+  __esModule: true,
+  default: () => true
+}));
+
+it('should have expected appearance and links', () => {
+  render(<DashboardSidebar shouldAllowDashboard={true} />);
```
benjaminEwhite, line 18: Need to update this to remove `shouldAllowDashboard={true}`.

benjaminEwhite (commented):
@gurjittoptal changes look great. There are a couple references to `shouldAllowDashboard` to be removed, but then I'm 👍 (and after you do those, no need for me to re-review — you can pass on to Daniel).

spa/src/components/dashboard/sidebar/DashboardSidebar.test.js
gurjittoptal: done

benjaminEwhite (approved):

benjaminEwhite (approved):
LGTM.

PR #472, opened 2022-06-20
DEV-86 | Enable admins to revert to a previous version of a page

benjaminEwhite:

This PR is initially into DEV-1186, which should be merged before this is merged. This PR branches from and presupposes DEV-1186, so provisionally setting that as base so diff will be unique to this PR.

#### What's this PR do?

Adds django-reversion-compare as dependency and configures `DonationPageAdmin` to use it, in order to make it a bit easier for admin users to compare any two page versions when considering whether or not to revert.

Make it easy for admins to understand differences between donation page versions when considering making a change.

In Django admin, make and save one or more changes to a donation page. Then from the detail view for that donation page, click the `History` button in the top right hand corner. There, you'll see the compare interface which will allow you to see differences between any two given versions.

No. We just add the package and use -- no need to test.

N/A

No

No

[DEV-86 | Hub admin can revert to previous version of page](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-86)

No.

No.


njharman (approved):
lgtm

Was able to do donation page history revision shenanigans. Also delete page and recover it kept all the history (the delete from recovery was treated as new revision)

x110dc (approved):
LGTM.

PR #477, opened 2022-06-24
DEV-1223 Reset Django migrations before launch

benjaminEwhite:

#### What's this PR do?

- Resets all app migrations, deleting the old ones. This results in a large diff in terms of file numbers, but most of the diff is deleted files.
- Creates a new post deploy script (`./scripts/heroku-post-migration-reset-postdeploy`) and points `app.json` (which configures review apps) to call that script in postdeploy step. This script is an interim step, and after this PR has been merged into staging, the new script can be deleted and `app.json` can be reverted (a ticket has been created to track this, see below).
- Adds `./scripts/reconcile_migrations.sh` which (as noted in code comment and in deploy instructions below) should be run per deploy environment after merging this PR. That script will ensure that the respective database's `django_migrations` table reflects the reset migrations having been applied.

Note that there were two apps where I did not fully reset the migrations, instead retaining the first migration: `apps.config` and `apps.users`. The former was retained because it configures the postgres citext extension. When I tried resetting all the way to zero for `apps.config`, I was unable to get migration commands to work, because of a dependency on citext. Similarly, when I tried zeroing out `apps.users`, I would get similar issues.

Also, worth noting that in some cases, resetting ended up creating 2 migrations instead of the expected 1 (for instance `apps.contributions` and `apps.organizations`). This is how Django defaulted to creating these migrations from a zeroed out state, and going from dozens and dozens of migrations to two seemed like a big enough win, so I did not pursue getting them to condense to a single migration per app.

Should make migrations run more quickly, and help test speed. Ease of maintenance, and opportunistically, pre-launch is the time to do this.

_In review app_

In a review app, confirm that old functionality continues to work.

In a shell for the review app, run `python manage.py makemigrations` and `python manage.py migrate`. The former should not create new migrations while the later should report there are no migrations needed.

_Local dev environment_

1. `git pull`
2. `make nuclear`
3. `docker compose up -d`
4. `git checkout DEV-1223-consider-resetting-django-migrations-before-launch`
5. `heroku pg:backups:download --app rev-engine-staging b008`
6. `pg_restore --clean --no-acl --no-owner -d revengine latest.dump`
7. `./scripts/reconcile_migrations.sh`
8. `python manage.py migrate` (Should report no migrations required)

After above, all tests should pass locally, and if you run app, all functionality should work. 

N/A

N/A

No

N/A

- [DEV-1223 | Reset Django migrations before launch](https://news-revenue-hub.atlassian.net/browse/DEV-1223)
- [DEV-1186 | Only record object changes when triggered by user actions](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-1186)
- [Delete heroku-post-migration-reset-postdeploy.sh and revert app.json to point at to `./scripts/heroku-postdeploy.sh`](https://news-revenue-hub.atlassian.net/browse/DEV-1985)
 

After merging into the deploy branch, run `./scripts/reconcile_migrations.sh`. This script will get the `django_migrations` table to reflect that the reset migrations have been run.

[Delete heroku-post-migration-reset-postdeploy.sh and revert app.json to point at to `./scripts/heroku-postdeploy.sh`](https://news-revenue-hub.atlassian.net/browse/DEV-1985)


njharman (approved):
Reviewed code. Migrates from empty db, tests pass, coverage is good. Accessed django admin.

lgtm, comment in script is more "has this been considered" than change request or merge blocker

x110dc (approved):

PR #478, opened 2022-06-24
DEV-1893 - DEV-1893_force_https

x110dc:
#### What's this PR do?

Asks Django to redirect any non-TLS requests (http) to HTTPS.

Security.

```
python manage.py runserver
curl -I http://localhost:8000/
```
Confirm that you see something like:
```
HTTP/1.1 301 Moved Permanently
...
Location: https://localhost:8000/
```
```
SECURE_SSL_REDIRECT=False python manage.py runserver
curl -I http://localhost:8000/
```
Confirm that you get a 200:
```
HTTP/1.1 200 OK
...
```

No.

N/A

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-1893

No, but for local development you'll probably want to set `SECURE_SSL_REDIRECT=False` unless you want to SSL locally.

No.


benjaminEwhite (changes requested):
This works but does break the development server when running locally. 

We could solve that in one of two ways:

1. Update `README.md` to advise developers to add `export SECURE_SSL_REDIRECT=false` to their `.envrc` file.
2. Update `revengine.settings.local.example.py` to have the line `SECURE_SSL_REDIRECT = os.environ.get("SECURE_SSL_REDIRECT", "false").lower() == "true"`. Then, devs would need to know that they should copy the updated `local.example.py` to `local.py` in their local dev environment.

There's probably not much of a use case for having SSL redirect in development, so option 2 might make more sense.

x110dc:
I'd added a task to ask engineers to add `SECURE_SSL_REDIRECT=False` to their environments, but your suggestions are better, so I'll do the 2nd, as you suggested.

PR #479, opened 2022-06-24
DEV-1856 | logs should use %s formatting

benjaminEwhite:

N/A

#### What's this PR do?

- Adds flake8-logging-format as a dev dependency and to `.pre-commit-config.yaml`. 
- Cleans up log string formatting in 15 Python files
- Updates a handful of tests that inspect what is logged

Will enable log grouping in Sentry

N/A

N/A

N/A

N/A

N/A

[DEV-1856 | logs should use %s formatting](https://news-revenue-hub.atlassian.net/browse/DEV-1856)

[DEV-1977 | Reduce Sentry Noise for initial public release](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-1977)

No

No


.flake8
```diff
@@ -6,9 +6,13 @@ max-line-length=100
...
 exclude=migrations,docs,node_modules,revengine/settings
+enable-extensions=G
 ignore =
     E2,    # missing whitespace after ','
     W503,   # line break before binary operator
     E501,
     F841,   # Line too long
     E741,   # Ambiguous variable
+    G002,   # Logging statements should not use % formatting for their first argument
+    G003,   # Logging statements should not use + concatenation for their first argument
```
njharman, line 17: -1 on ignoring this. Where do we use concatenation? Or more to point, why shouldn't it be rewritten to not use concatenation. Curious I checked, doesn't see there are any concatenations???

Removing G003 and running flake8 didn't find any violations of G003. removing all the GXXX didn't show any violations.

Oddly, did find a bunch of other flake8 issues... see team-eng post and ticket

apps/common/utils.py
```diff
@@ -51,8 +51,8 @@ def delete_cloudflare_cnames(ticket_id):
             if ticket_id.lower() in domain.lower():
                 logger.info("Deleting DNS entry for: %s", domain)
                 cloudflare_conn.zones.dns_records.delete(zone_id, record_id)
-        except CloudFlare.exceptions.CloudFlareAPIError as error:
-            logger.warning(error)
+        except CloudFlare.exceptions.CloudFlareAPIError:
+            logger.exception()
```
njharman, line 55: This changes log level from warning to error. The "template" I've used for logging exception data is

> logger.[level]("msg", exc_info=error)  

See https://docs.python.org/3/library/logging.html#logging.debug

apps/organizations/models.py
```diff
@@ -277,9 +277,9 @@ def stripe_create_apple_pay_domain(self):
                 )
                 self.domain_apple_verified_date = timezone.now()
                 self.save()
-            except stripe.error.StripeError as stripe_error:
-                logger.warning(
-                    f"Failed to register ApplePayDomain for RevenueProgram {self.name}. StripeError: {str(stripe_error)}"
+            except stripe.error.StripeError:
+                logger.exception(
```
njharman, line 281: Changes log level from warning to error.

[I don't like logger.exception() it violates zen of python ought to be one way to do things, it obscures what log level this will be logged at, and denies the ease of changing log level that the warn/error/etc methods provide.

Bah humbug!]

apps/organizations/tests/test_org_models.py
```diff
@@ -117,7 +117,7 @@ def test_apple_pay_domain_verification_when_stripe_error(self, mock_logger, appl
         apple_pay_domain_create.side_effect = StripeError
         factories.RevenueProgramFactory()
         apple_pay_domain_create.assert_called_once()
-        mock_logger.warning.assert_called_once()
+        mock_logger.exception.assert_called_once()
```
njharman, line 120: Changes log level

apps/slack/slack_manager.py
```diff
@@ -118,15 +120,11 @@ def send_hub_message(self, channel, text, blocks):
         except SlackApiError as slack_error:
             error_type = slack_error.response["error"]
             if error_type == "invalid_auth":
-                logger.error(
-                    f"SlackApiError. HubSlackIntegration has invalid token. SlackError: {slack_error.response}"
-                )
+                logger.exception("SlackApiError. HubSlackIntegration has invalid token")
```
njharman, line 123: Was it intentional to change a few of these here to log full exception/traceback vs just the slack_error.response. But further on, in send_org_message, still logging just the response and not full traceback?

njharman (changes requested):
re flake8-logging, the "extra=" logging argument is new to me. Seems best method (more consistent with f-string). Why was % formatting selected instead? [I originally suggested it, as at the time it was the way I knew of sentry grouping]

apps/slack/slack_manager.py
benjaminEwhite: Yes, intentional. I figured there's no reason to log only part of the exception, given that the new flake8 settings prohibit error + exception reference in the message.

benjaminEwhite:
> re flake8-logging, the "extra=" logging argument is new to me. Seems best method (more consistent with f-string). Why was % formatting selected instead? [I originally suggested it, as at the time it was the way I knew of sentry grouping]

@njharman - Yes, this is why we are making the move to enforce %s. From what [I've read online ](https://discuss.python.org/t/safer-logging-methods-for-f-strings-and-new-style-formatting/13802/18)(though can't find this directly mentioned in Sentry docs), the only way to get consistent grouping is via %s formatting.  It seems there are also some security concerns around using f-strings in logs ([see here](https://discuss.python.org/t/safer-logging-methods-for-f-strings-and-new-style-formatting/13802/18)) .

I found [where `extra=""` is mentioned in the Sentry docs](https://docs.sentry.io/platforms/python/guides/logging/), but I didn't get the impression that extra data plays into grouping at all, just that the extra data will available in the event data, and that you can search by it. 

Unless you know differently, @njharman , IIUC the only way to get grouping is by using %s formatting.

apps/common/utils.py
benjaminEwhite: Good catch. Changed back to warning, and added a msg that has domain as %s arg.

apps/organizations/tests/test_org_models.py
benjaminEwhite: reverted to warning.

njharman (approved):
lgtm, ship it!

re "% format" vs "extra="; I don't know, and didn't read the Sentry docs on this (recently).  "% formatting is what I've always used and know it solves this issue. Just wondering/hoping extra= was "better" alternative.

x110dc (approved):
Added a tiny commit to unpin the flake8 dep in `pyproject.toml`

PR #482, opened 2022-06-27
Dev 1920 new dashboard navs unit tests

benjaminEwhite:

Demo of how to test sidebar, with minimal refactoring to make testable with React Testing Library (which in turn encourages creating accessible UI components).

#### What's this PR do?

- Updates `package.json` and adds two Jest mocks in `spa/__mocks__` in order to allow components to load in Jest tests when they have static assets (in this case CSS, as dependencies), as [described in the Jest docs](https://jestjs.io/docs/webpack#handling-static-assets). Note that this became necessary because the component under test (`DashboardSideBar`) has dependencies in `Dashboard`, which itself imports css (`import 'react-datepicker/dist/react-datepicker.css'`).
- Adds a Jest test for the DashboardSidebar element, and does minimal refactoring of the subject under test so that we can use `getByRole` queries in our Jest tests. This is in line with the [priority of query methods](https://testing-library.com/docs/queries/about/#priority), which encourage developers to write tests that "resemble how users interact with your code").
- Renames `DonationTopBar.test.js` to `DashboardTopBar.test.js`, and lightly refactors to get sign out element by its text (instead of by id) and to test the click-behavior for that element.


gurjittoptal (approved):
We need this, but by refactor i meant keeping markup and state/logic as seperate as possible.
https://news-revenue-hub.atlassian.net/browse/DEV-1920?focusedCommentId=48681

We should merge this to PR for DEV-1920 and take it from there.

PR #501, opened 2022-07-02
DEV-1577_spa_payment_provider_not_provided_handler

gurjittoptal:

fixes the white screen that shows up if paymentprovider is not connected

#### What's this PR do?
does the following
1. fixes the white screen
2. adds a logout link to the generic error page

better user experience

login as a  hub admin for org org-admin or rp-admin to check various scenarios

no

na

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-1577


spa/cypress/integration/10 - connect.spec.js
```diff
@@ -20,6 +20,7 @@ const orgAdminWithContentFlag = {
 
 const redirectPath = '/dashboard/content';
 
+/*
```
benjaminEwhite, line 23: This whole file should be deleted.

spa/src/components/Main.js
```diff
@@ -17,6 +17,8 @@ const FeatureFlagsProviderContext = createContext(null);
 function Main() {
   // Organization Context management
   const [checkingProvider, setCheckingProvider] = useState(false);
+
+  // Given we are tying payment_provider at RP level, all this can be cleaned.
```
benjaminEwhite, line 21: Should this be a todo (with associated JIRA ticket)?

spa/src/utilities/getGlobalPaymentProviderStatus.js
```diff
@@ -15,6 +15,9 @@ function getGlobalPaymentProviderStatus(user) {
   const organization = Boolean(organizations) ? organizations[0] : null;
   const dontNeedProviderTypes = [USER_ROLE_HUB_ADMIN_TYPE, USER_SUPERUSER_TYPE];
   const roleType = user.role_type[0];
+
+  return PP_STATES.CONNECTED;
```
benjaminEwhite, line 19: This seems strange to have a function that invariably returns the connected state. If connected state is always going to be true, then the conditional logic around this should be entirely removed.

spa/src/utilities/getGlobalPaymentProviderStatus.js
```diff
@@ -24,7 +27,7 @@ function getGlobalPaymentProviderStatus(user) {
     return getStripePaymentManagerProviderStatus(accountIdPresent, accountVerified);
   } else {
     return undefined;
-  }
+  }*/
 }
 
 function getStripePaymentManagerProviderStatus(accountIdPresent, accountVerified) {
```
benjaminEwhite, line 33: Realize this is not touched by this PR, but this is not used anywhere and can be removed.

spa/src/utilities/getGlobalPaymentProviderStatus.js
```diff
@@ -24,7 +27,7 @@ function getGlobalPaymentProviderStatus(user) {
     return getStripePaymentManagerProviderStatus(accountIdPresent, accountVerified);
   } else {
     return undefined;
-  }
+  }*/
```
benjaminEwhite, line 30: Realize this commented code is not touched by this PR, but should be removed.

benjaminEwhite (changes requested):
I think we may want a slightly different approach here. If I understand correctly, this PR makes it so that the only state the app can return is connected, for payment provider, regardless of whether or not a provider is in fact connected on the backend. In that case, it would make more sense to remove the payment provider set up logic altogether.

I also noted in the review app, that when you remove the payment provider from an RP, it is no longer possible to load existing pages in the editor.

Screen Shot 2022-07-04 at 11 23 19 AM

gurjittoptal:
@benjaminEwhite 
re: I also noted in the review app, that when you remove the payment provider from an RP, it is no longer possible to load existing pages in the editor.

This has to be handled in the backend. i.e. If an RP cannot be created unless a payment provider is provided, a payment-provider SHOULD-NOT be allowed to be physically deleted from the backend IF it is attached to an RP.

I think we should keep the front-end code clean of this logic.

spa/src/components/Main.js
gurjittoptal: yes, connected_state code has to be cleaned as payment_provider is moved to rp level via a new jira.

spa/src/utilities/getGlobalPaymentProviderStatus.js
gurjittoptal: yes. via new jira ticket. see comment above.

spa/src/utilities/getGlobalPaymentProviderStatus.js
gurjittoptal: comment above.

spa/src/utilities/getGlobalPaymentProviderStatus.js
gurjittoptal: comment above.

benjaminEwhite (changes requested):
Given that we'll be revisiting the flows around payment provider set up soon anyways, I still would lean towards removing the payment provider logic entirely in this PR, rather than commenting out most of the existing getGlobalPaymentProviderStatus function body and having it return a static value.

There's not much value in keeping this old code around, and if for some reason we ever want to go back and look at it, it will be in the repo history anyways.

I've also suggested another possibility which is removing the call to getGlobalPaymentProviderStatus in Main.js and just hard coding the PP_STATES.CONNECTED value into the useState instantiation, along with a comment explaining why.  

Between the two options, I'd lean more toward the first.

@gurjittoptal , I agree that this PR doesn't necessitate new unit tests, as it's removing functionality, and we'll want to add unit tests to new business logic around payment provider set up, which will be landing soon.

spa/cypress/integration/10 - connect.spec.js
benjaminEwhite: I don't understand why this diff shows up as only adding a line at bottom of this file since in `main` this file is not commented out.

🤷 🤷‍♀️

spa/src/components/Main.js
```diff
@@ -17,6 +17,8 @@ const FeatureFlagsProviderContext = createContext(null);
...
   const [paymentProviderConnectState, setPaymentProviderConnectState] = useState(
     getGlobalPaymentProviderStatus(JSON.parse(localStorage.getItem(LS_USER)))
```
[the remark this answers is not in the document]
gurjittoptal: @benjaminEwhite 
_on "I still would lean towards removing the payment provider logic entirely in this PR"_

i agree. I did cleanup all the code in commit [1cd3b7e9](https://github.com/newsrevenuehub/rev-engine/pull/501/commits/1cd3b7e95f80d0b9aa7fbe882425c326079da134)

But we should do this after new-nav  is merged.

It also does make code cleaner and more readable which is a good thing

[dashboard.js - dev-1577](https://github.com/newsrevenuehub/rev-engine/blob/DEV-1577_spa_payment_provider_not_provided_handler/spa/src/components/dashboard/Dashboard.js)
vs
[dashboard.js - main](https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/dashboard/Dashboard.js)

PR #507, opened 2022-07-05
DEV-2005 | unable to request magic link for contributor portal when domains are gt 20 characters long

benjaminEwhite:

#### What's this PR do?

- Renames `SLUG_MAX_LENGTH` to `RP_SLUG_MAX_LENGTH` so meaning aligns with use
- Removes max length limit on  `ContributorObtainTokenSerializer.subdomain`.  This was originally set at 20 characters and was the cause of the bug with magic link email requests. 
- Removes max length limit  on `ContributionMetadataSerializer.revenue_program_slug`. This code was already touched by renaming `SLUG_MAX_LENGTH` TO `RP_SLUG_MAX_LENGTH`, and I concluded that there shouldn't be a limit here in the first place.

The reason for not limiting subdomain in `ContributorObtainTokenSerializer.subdomain` has to do with how that serializer is used. It is narrowly used by the SPA to request magic email links. The SPA derives the current subdomain from `window.location.host`, and passes that value along in the body of the request to `RequestContributorTokenEmailView`. Thus if a request gets made where the value for subdomain is beyond the max length enforced at ORM layer (which is set to `RP_SLUG_MAX_LENGTH`, either a malicious actor is interacting directly with the API layer, or the SPA is somehow being served from a subdomain that violates the constraints around subdomain/rp-slug length. In neither case does it make sense to send validation errors about max length of subdomain request body param.

The reason for not limiting `ContributionMetadataSerializer.revenue_program_slug` is similar. That serializer gets used in the view that the SPA posts to when a user submits a stripe payment. The value that gets sent for `revenue_program_slug` is derived in the same way from `window.location.host`.

Makes form for requesting magic link emails work when the subdomain is greater than 20 characters.

Go to https://newsrevenuehub-dev-2005.revengine-review.org/contributor and submit the form with a valid email address. It should submit, and you get the normal message about an email being sent.

No. Changes in this PR remove business logic in serializer layer around enforcing max length on two fields. That business logic arguably should have been tested in past, but since it's being removed, there's no need for additional tests.

N/A

N/A

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-2005

No

Sort of. Decoupled from this ticket, but the SPA should be refactored to display meaningful feedback to user when a request for magic email link fails. That's tracked in [DEV-2020](https://news-revenue-hub.atlassian.net/browse/DEV-2020)


apps/organizations/models.py
```diff
@@ -251,8 +251,8 @@ def __str__(self):
 
     def clean_fields(self, **kwargs):
         if not self.id:
-            self.slug = normalize_slug(self.name, self.slug, max_length=SLUG_MAX_LENGTH)
-            self.slug = normalize_slug(slug=self.slug, max_length=SLUG_MAX_LENGTH)
+            self.slug = normalize_slug(self.name, self.slug, max_length=DOMAIN_MAX_LENGTH)
+            self.slug = normalize_slug(slug=self.slug, max_length=DOMAIN_MAX_LENGTH)
```
njharman, line 255: Why is normalized_slug called twice.  This looks enough like error it needs comment explaining it is not, if indeed it is correct.

apps/organizations/models.py
```diff
@@ -78,7 +78,7 @@ class Organization(IndexedTimeStampedModel, RoleAssignmentResourceModelMixin):
     salesforce_id = models.CharField(max_length=255, blank=True, verbose_name="Salesforce ID")
 
     slug = models.SlugField(
-        max_length=SLUG_MAX_LENGTH,
+        max_length=DOMAIN_MAX_LENGTH,
```
njharman, line 81: I believe, this slug/it's length is unrelated to other use.  Esp if the name is changed to RP_SLUG.  Should just be a literal number, proly 63.

apps/organizations/models.py
benjaminEwhite: No idea, but it's weird!!!  I'm removing the second invocation.

apps/organizations/models.py
```diff
@@ -18,7 +18,7 @@
 logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
 
 # RFC-1035 limits domain labels to 63 characters
-SLUG_MAX_LENGTH = 63
+DOMAIN_MAX_LENGTH = 63
```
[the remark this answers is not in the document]
benjaminEwhite: Changing name as suggested.

apps/api/serializers.py
```diff
@@ -38,9 +39,9 @@ class ContributorObtainTokenSerializer(serializers.Serializer):
 
     email = serializers.EmailField()
     access = serializers.CharField(required=False)  # Token used in magic-link url/email.
-    # Domains are limited to 63 characters max. No underscore allowed, unlike slugs.
+
     subdomain = serializers.RegexField(
-        r"^[-0-9a-zA-Z]+$", max_length=20, required=False, allow_blank=True
+        r"^[-0-9a-zA-Z]+$", max_length=DOMAIN_MAX_LENGTH, required=False, allow_blank=True
```
benjaminEwhite, line 44: @njharman could use a second opinion here, but looking at this more closely, I don't think there's a good reason to enforce max length on this field in the serializer. While I don't think there's much a security concern here, I don't think there's much reason to be emitting information about the server's requirements around the max length a subdomain should be. That should really only be a concern for the RevenueProgram model at the ORM layer.

Agree/disagree?

apps/contributions/serializers.py
```diff
@@ -231,7 +231,7 @@ class ContributionMetadataSerializer(ConditionalRequirementsSerializerMixin):
     sf_campaign_id = serializers.CharField(max_length=255, required=False, allow_blank=True)
     referer = serializers.URLField()
     revenue_program_id = serializers.IntegerField()
-    revenue_program_slug = serializers.SlugField(max_length=SLUG_MAX_LENGTH)
+    revenue_program_slug = serializers.SlugField(max_length=DOMAIN_MAX_LENGTH)
```
benjaminEwhite, line 234: @njharman also reaching the same conclusion here. Seems to me it shouldn't be any of this serializer's business how long the passed revenue_program_slug is. If a >63 char value is getting sent here, something else has gone wrong further up the chain, and I don't think that sending a validation message about the revenue program slug being too long is going to help.


My inclination is to remove max_length here altogether. Thoughts?

apps/api/serializers.py
```diff
@@ -38,9 +39,9 @@ class ContributorObtainTokenSerializer(serializers.Serializer):
 
     email = serializers.EmailField()
     access = serializers.CharField(required=False)  # Token used in magic-link url/email.
-    # Domains are limited to 63 characters max. No underscore allowed, unlike slugs.
+
     subdomain = serializers.RegexField(
-        r"^[-0-9a-zA-Z]+$", max_length=20, required=False, allow_blank=True
+        r"^[-0-9a-zA-Z]+$", max_length=RP_SLUG_MAX_LENGTH, required=False, allow_blank=True
```
benjaminEwhite, line 44: @njharman could use a second opinion here, but looking at this more closely, I don't think there's a good reason to enforce max length on this field in the serializer. While I don't think there's much a security concern here, I don't think there's much reason to be emitting information about the server's requirements around the max length a subdomain should be. That should really only be a concern for the RevenueProgram model at the ORM layer.

Agree/disagree?

apps/contributions/serializers.py
```diff
@@ -231,7 +231,7 @@ class ContributionMetadataSerializer(ConditionalRequirementsSerializerMixin):
     sf_campaign_id = serializers.CharField(max_length=255, required=False, allow_blank=True)
     referer = serializers.URLField()
     revenue_program_id = serializers.IntegerField()
-    revenue_program_slug = serializers.SlugField(max_length=SLUG_MAX_LENGTH)
+    revenue_program_slug = serializers.SlugField(max_length=RP_SLUG_MAX_LENGTH)
```
benjaminEwhite, line 234: @njharman also reaching the same conclusion here. Seems to me it shouldn't be any of this serializer's business how long the passed revenue_program_slug is. If a >63 char value is getting sent here, something else has gone wrong further up the chain, and I don't think that sending a validation message about the revenue program slug being too long is going to help.

My inclination is to remove max_length here altogether. Thoughts?

benjaminEwhite:
@njharman your point about unit tests makes sense, but let me know how you feel after considering my contention that those references to max length in the serializers should be removed. If those go away, then the only place this business logic lies is in the ORM layer, and there's no need to test that Django ORM enforces its `max_length` param on SlugFields.

njharman:
> ce

njharman (approved):
lgtm given changes discussed

x110dc (approved):

PR #508, opened 2022-07-08
DEV-2049 retire address model

benjaminEwhite:

#### What's this PR do?

- Retires the `Address` model
- Adds country as a column directly on `RevenueProgram`, and migrates existing rp.address.country values to this new column.
- Removes tests around `Address` and old implementation
- Updates `DonationPageFullDetailSerializer.get_organization_country` to become `DonationPageFullDetailSerializer.get_revenue_program_country` and updates serializer field to `revenue_program_country` from `organization_country` to reflect changed business logic where stripe payment manager is related to RP not org.
- Updates tests and fixtures in light of above.

Part of rationalizing model relationships. Initially, it was discovered that the address model should be refactored, and that led to conclusion that address could be retired entirely.

- SPA still works as before (donation pages load, and can make a contribution), since the Stripe integration requires rp country to be set.
- In Django admin, edit an rp's country.

No. Nothing new required.

N/A

No.

No.

[DEV-2049](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-2049)

[DEV-1581](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-1581)


apps/organizations/models.py
```diff
@@ -15,6 +16,17 @@
...
 
+def get_country_choices():
+    """
+    returns a tuple of country choices according to pycountry.countries db
+    """
+    country_choices = []
+    for country_code in settings.COUNTRIES:
+        country = pycountry.countries.lookup(country_code)
+        country_choices.append((country.alpha_2, country.alpha_2))
+    return country_choices
```
njharman, line 27: This is weird. first why for loop and not simple list comp. 2nd why is it "dynamic" (run every time it's needed) it's based on set once on import config setting [edit: oh it's, of course, only called once, in the field definition. more reason to be inline list comp...] 

3rd, main point:  settings.COUNTRIES seems to be alpha 2 country list, we're using those to look up alpha 2 code pycountry says is correct?  So, we have to remember to do that everywhere?
Why not just validate settings.COUNTRIES are valid alpha 2 (on import using pycountry) and using settings.COUNTRIES directly country_choices = list(zip(settings.COU..., settings.COU...)

This whole function smells to me!

apps/organizations/models.py
```diff
@@ -89,6 +100,14 @@ class Organization(IndexedTimeStampedModel, RoleAssignmentResourceModelMixin):
         help_text="If false, receipt email assumed to be sent via Salesforce. Other emails, e.g. magic_link, are always sent via NRE regardless of this setting",
     )
 
+    country = models.CharField(
+        max_length=2,
+        blank=True,
```
njharman, line 105: I question that country is not required. Is there requirement/use case when we would not know/want to have the country of organization?

apps/organizations/models.py
```diff
@@ -89,6 +100,14 @@ class Organization(IndexedTimeStampedModel, RoleAssignmentResourceModelMixin):
         help_text="If false, receipt email assumed to be sent via Salesforce. Other emails, e.g. magic_link, are always sent via NRE regardless of this setting",
     )
 
+    country = models.CharField(
```
njharman, line 103: From past experience, I've discovered it's helpful to have help text explaining what country is meant to be used for. Legal country (as in address), country of operations, country for the purposes of currency or language, etc.

njharman (dismissed):
Tests pass, coverage good, admin works as advertised.
Did not test SPA

Couple minor comments/questions.

apps/organizations/models.py
benjaminEwhite: This question prompted discussion with Mariclare, which resulted in moving country to RP instead of organization. And determined not required (and defaulting to US is fine).

apps/organizations/models.py
benjaminEwhite: This is in RP now, but added help text explaining.

apps/organizations/models.py
benjaminEwhite: Changed my mind and made not dynamic, and just removed py countries from the mix entirely, cause not clear it's adding any value.

x110dc (approved):
LGTM.

PR #515, opened 2022-07-12
Dev 1854 allow contributors not in nre

harsha09:
#### What's this PR do?
1. Creates a new contributor in NRE if the contributor doesn't exist already and sends magic link.
2. Users who aren't recognised by NRE will be able to get magic_link and able to access contributor portal.
3. Right now users who are not recognised by NRE will be able to see blank contributor portal page, when [DEV-1691] is merged they should be able to view contributions from stripe directly.

When [DEV-1691] is merged, users should not be limited to access contributor portal.

Request a magic link with new user who is not a contributor in NRE.
Access contributor portal using the magic link.

Not added any UTs but fixed few as they were breaking with the code change.

NA

NA

NA

[DEV-1854]

[DEV-1691] needs to be completed.

[DEV-1691]: https://news-revenue-hub.atlassian.net/browse/DEV-1691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-1691]: https://news-revenue-hub.atlassian.net/browse/DEV-1691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-1854]: https://news-revenue-hub.atlassian.net/browse/DEV-1854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


x110dc (approved):

apps/api/serializers.py
```diff
@@ -53,11 +53,9 @@ def get_token(cls, contributor):
...
         """
         If email is valid and matches that of a known contributor, we provide a access token.
+        If email doesn't exist in NRE, we create a contributor and provide access token to that contributor.
         """
         data = super().validate(attrs)
-        try:
-            contributor = Contributor.objects.get(email=data.get("email"))
-        except Contributor.DoesNotExist:
-            raise NoSuchContributorError("Could not find contributor", code="no_contributor_email")
+        contributor, _ = Contributor.objects.get_or_create(email=data.get("email"))
```
benjaminEwhite, line 59: Having a validate method create a new instance seems like an anti-pattern to me. IMO, serializer validation methods shouldn't have any side effects like that, and this business logic should live in the view layer.

Seems like it would be pretty straightforward to refactor this serializer so its only fields are `email` and `subdomain`. Then in `apps.api.views.RequestContributorTokenEmailView`, after line 146, do the get or create for the contributor, and create token after that step, instead of pulling out of serializer data.

benjaminEwhite (changes requested):
As noted, IMO the contributor creation logic should live in the view layer,  not the serializer layer.

I also think there should be a single, simple view test that demonstrates that this view creates a contributor if they do not already exist. That's important and not very obvious business logic, and should therefore be tested.

apps/api/serializers.py
harsha09: `access` key in the response body is needed for the client to process.
I've extracted out the `Contributor` creation to view layer. Also I've removed token generation and setting the access token in `validate` method as it's doing more than one thing other than validation.

apps/api/tests/test_views.py
```diff
@@ -94,9 +94,10 @@ def test_normal_response_when_valid_but_unrecognized_email(self):
         target_email = "testing123@testing123.com"
         # Verify that there are no contributors with target email
         self.assertEqual(Contributor.objects.filter(email=target_email).count(), 0)
-        response = self.client.post(self.url, {"email": target_email})
+        response = self.client.post(self.url, {"email": target_email, "subdomain": "rp"})
         # We don't want to indicate in any way whether or not an email is in the system.
         self.assertEqual(response.status_code, 200)
+        self.assertEqual(Contributor.objects.filter(email=target_email).count(), 1)
```
benjaminEwhite, line 100:
```suggestion
        # this view creates a contributor if none exists
        self.assertEqual(Contributor.objects.filter(email=target_email).count(), 1)
```

benjaminEwhite (approved):
Added a suggestion to test code that adds comment explaining why. 

Approved.

PR #530, opened 2022-07-18
DEV-2037 | API support for user creation

benjaminEwhite:

#### What's this PR do?

Adds API support for first step of self-service sign up flow. This first step is around creating a user instance, initially with just an email address and password.

A few things to note:

- `user.views.retrieve_user` has been subsumed by UserViewSet's list method, but corresponds to the same URL structure, so won't break SPA.
- For now, the email verification emails contain dummy copy, no styling, and no email verification link. That last bit, around email verification link, will be handled in subsequent ticket in this epic.

Scale the platform.

Since there is no UI for this, you'd need to use something like Postman. 

If you're running Postman locally (and this review app is still up), you can grab the JSON for a Postman collection and try it out: https://www.postman.com/collections/414d8c2490d38a4e4673

Yes

N/A

No

No

[DEV-2037](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-2037)

No


apps/users/tests/test_views.py
```diff
@@ -90,17 +99,225 @@ def test_happy_path(self):
...
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json(), {"accepted_terms_of_service": ["This field is required."]})
+
+    def test_create_when_not_accepted_terms_of_service_is_empty(self):
+        data = {**self.create_data}
+        data["accepted_terms_of_service"] = ""
+        response = self.client.post(self.url, data=data)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("accepted_terms_of_service", response.json().keys())
+        self.assertIn("Datetime has wrong format", response.json()["accepted_terms_of_service"][0])
```
njharman, line 169: ~Would maybe use pytest.paramaterize. Test these two; field missing, field blank and when fal...~  oh, it's probably a check box and blank is the "false" response, eh?

apps/users/tests/test_views.py
```diff
@@ -90,17 +99,225 @@ def test_happy_path(self):
...
+        self, mock_logger_warning, mock_bad_actor_response, mock_send_verification_email
+    ):
+        user_count = get_user_model().objects.count()
+        response = self.client.post(self.url, data=self.create_data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(get_user_model().objects.count(), user_count + 1)
+        mock_logger_warning.assert_called_once_with("Something went wrong with BadActorAPI", exc_info=True)
+        mock_send_verification_email.assert_called_once()
+
+    def test_create_when_email_already_taken(self):
```
njharman, line 269: Do we also need similar already taken test for PATCH?

apps/users/serializers.py
```diff
@@ -58,13 +80,63 @@ def get_permitted_revenue_programs(self, obj):
...
+        """We manually handle create step because password needs to be set with `set_password`"""
+        password = validated_data.pop("password")
+        User = get_user_model()
+        user = User(**validated_data)
+        user.set_password(password)
+        user.is_active = False
+        user.save()
+        return user
+
+    def update(self, instance, validated_data):
```
njharman, line 103: If email changes, should we set email_verified to False?

apps/users/tests/test_views.py
```diff
@@ -90,17 +99,225 @@ def test_happy_path(self):
...
+    @patch.object(UserViewset, "validate_password")
+    @patch.object(UserViewset, "validate_bad_actor")
+    @patch.object(UserViewset, "send_verification_email")
+    def test_create_happy_path(self, mock_send_verification_email, mock_validate_bad_actor, mock_validate_password):
+        user_count = get_user_model().objects.count()
+        response = self.client.post(self.url, data=self.create_data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(get_user_model().objects.count(), user_count + 1)
+        self.assertEqual((data := response.json())["email"], self.create_data["email"])
+        self.assertIsNotNone((user := get_user_model().objects.filter(id=data["id"]).first()))
```
njharman, line 142: I don't like this use of := one bit.

apps/users/serializers.py
```diff
@@ -58,13 +80,63 @@ def get_permitted_revenue_programs(self, obj):
...
+            "email",
+            "email_verified",
+            "flags",
+            "id",
+            "organizations",
+            "revenue_programs",
+            "role_type",
+            "password",
+        ]
+        read_only_fields = [
```
njharman, line 135: Is this just our additions to user model / it's layered on top of what Django default
Should is_active, is_staff, ... others be added to read_only_fields

apps/users/tests/test_views.py
```diff
@@ -90,17 +99,225 @@ def test_happy_path(self):
...
+        self.assertFalse(user.is_active)
+        self.assertFalse(data["email_verified"])
```
njharman, line 144: check email_verified on user object (perhaps in addition to data)
would also check that password and email set on user. Because we overrode code for former and both are listed in AC.

apps/users/serializers.py
njharman, line 103: What if any is the Business Rule around updating accepted_terms_of_service to False (or even resetting date of acceptance).  I did not see it defined in AC (maybe new ticket).
allow, error?

apps/users/tests/test_views.py
```diff
@@ -90,17 +99,225 @@ def test_happy_path(self):
...
+
+    def test_create_when_not_accepted_terms_of_service_is_empty(self):
+        data = {**self.create_data}
+        data["accepted_terms_of_service"] = ""
+        response = self.client.post(self.url, data=data)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("accepted_terms_of_service", response.json().keys())
+        self.assertIn("Datetime has wrong format", response.json()["accepted_terms_of_service"][0])
+
+    def test_create_when_no_password(self):
```
njharman, line 171: Can all these password tests be genericized and applied to patch as well? 
Should test, at least (and I think sufficient), that our password code is called for patch.

apps/users/tests/test_views.py
```diff
@@ -90,17 +99,225 @@ def test_happy_path(self):
...
+        another_user = get_user_model().objects.create(email="another_user@example.com")
+        self.client.force_authenticate(user=my_user)
+        update_email = "updated@example.com"
+        response = self.client.patch(reverse("user-detail", args=(another_user.pk,)), data={"email": update_email})
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertEqual(response.json(), {"detail": UserOwnsUser.message})
+        another_user.refresh_from_db()
+        self.assertNotEqual(another_user.email, update_email)
+
+    def test_cannot_partial_update_when_email_not_verified(self):
```
njharman, line 305: I agree should not be able to update until verified email. But I did not see in AC. was it somewhere else.  Think worth adding somewhere; communicating to/ validating with business folks that is desired behavior.

apps/users/tests/test_views.py
```diff
@@ -90,17 +99,225 @@ def test_happy_path(self):
...
+        update_data = {"password": raw_updated_password, "email": new_email}
+        self.client.force_authenticate(user=user)
+        response = self.client.patch(reverse("user-detail", args=(user.pk,)), data=update_data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        user.refresh_from_db()
+        self.assertNotEqual(old_hashed_pw, user.password)
+        self.assertNotEqual(user.password, raw_updated_password)
+        self.assertEqual(response.json()["email"], new_email)
+
+    def test_partial_update_when_not_my_user(self):
```
njharman, line 294: Depending on responses to some other comments. Probably addition tests for patch.

- password
- bad email
- accepted_terms_of_service
-

njharman (changes requested):

apps/users/serializers.py
```diff
@@ -58,13 +80,63 @@ def get_permitted_revenue_programs(self, obj):
...
+        User = get_user_model()
+        user = User(**validated_data)
+        user.set_password(password)
+        user.is_active = False
+        user.save()
+        return user
+
+    def update(self, instance, validated_data):
+        """We manually handle update step because password needs to be set with `set_password`, if part of update"""
+        password = validated_data.pop("password") if "password" in validated_data else None
```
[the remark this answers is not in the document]
benjaminEwhite: Changed.

apps/emails/templates/nrh-org-account-creation-verification-email.html
```diff
@@ -0,0 +1,4 @@
+<div>
+	<p>Thanks for signing up</p>
+	<p>Please <a href="{verification_url}">click on this link</a>to verify your email address.
```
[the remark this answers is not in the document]
benjaminEwhite: Fixing

apps/users/serializers.py
```diff
@@ -23,14 +33,23 @@ class UserSerializer(serializers.ModelSerializer):
...
+            return None
         if obj.is_superuser:
             return ("superuser", "Superuser")
         role_assignment = obj.get_role_assignment()
         return (role_assignment.role_type, role_assignment.get_role_type_display()) if role_assignment else None
 
     def get_permitted_organizations(self, obj):
+        # `obj` will be a dict of data when serializer being used in create view
+        if not isinstance(obj, get_user_model()):
+            return []
```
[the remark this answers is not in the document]
benjaminEwhite: Both should be returning empty list. That's some "haste makes waste" on my end, as I was rushing to get done on last day of sprint.  Updating the other one to return empty list.

apps/users/serializers.py
```diff
@@ -45,6 +64,9 @@ def get_permitted_organizations(self, obj):
         return serializer.data
 
     def get_permitted_revenue_programs(self, obj):
+        # `obj` will be a dict of data when serializer being used in create view
+        if not isinstance(obj, get_user_model()):
+            return None
```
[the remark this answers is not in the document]
benjaminEwhite: Fixing.

apps/users/tests/test_views.py
benjaminEwhite: Adding test

apps/users/views.py
```diff
@@ -64,13 +74,102 @@ class CustomPasswordResetCompleteView(PasswordResetCompleteView):
...
+        validation functions we're using need access to the user attributes, not just password. This allows
+        us to access all fields that were already validated in serializer layer.
+        """
+        # we temporarily initialize a user object (without saving) so can use Django's built
+        # in password validation, which requires a user object
+        temp_user = get_user_model()(email=email)
+        try:
+            validate_password(password, temp_user)
+        except DjangoValidationError as exc:
+            raise ValidationError(detail={"password": exc.messages})
```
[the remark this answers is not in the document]
benjaminEwhite: Yeah, I understand that in the general case, but here, the validation error should be predictable, as we're narrowly using Django's built in p/w validation. I suppose we could explicitly check that messages are expected, and only send those that are, and if unexpected, send a generic.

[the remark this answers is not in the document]
benjaminEwhite: Pushing up a change that ensures message is in expected list, else defaults to generic.

apps/users/tests/test_views.py
benjaminEwhite: Done, in commit about to push up.

apps/users/serializers.py
benjaminEwhite: Good question. Consulting with Mariclare to figure this out.

apps/users/tests/test_views.py
njharman: Bah! With great power, comes great responsibility.

apps/users/tests/test_views.py
benjaminEwhite: This is indeed business logic.

apps/users/tests/test_views.py
benjaminEwhite: Added tests for password and email. Will need to get clarification on accepted_terms_of_service, and will likely handle in follow up.

PR #538, opened 2022-07-21
new try at DEV-1691

x110dc:
See PR description here: https://github.com/newsrevenuehub/rev-engine/pull/459


apps/contributions/stripe_contributions_provider.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+class StripeCharge:
+    """
+    Wrapper on stripe charge object to extract the required details in
+    apps.contributions.serializers.PaymentProviderContributionSerializer and serializable.
+    """
+
+    def __init__(self, charge):
+        if not hasattr(charge, "invoice") or not charge.invoice:
```
x110dc, line 55: One time-payments (Payment Intents and their related Charges) will never have invoices associated. So, I think this will only work for recurring payments, no?

PR #543, opened 2022-07-21
redo DEV-1577 SPA - refactor code to remove paymentProvider redundant code

x110dc:
See https://github.com/newsrevenuehub/rev-engine/pull/501

fixes the white screen that shows up if paymentprovider is not connected

#### What's this PR do?
does the following
1. fixes the white screen
2. adds a logout link to the generic error page

better user experience

To test
1. Login as various roles and the dashboard should show up
2. Remove a payment-provider for an RP, this should not lead to an error/ white screen showing up
3. A logout link will show in generic error page, this is not easily testable as its unlikely to happen normally. So a cypress test covers this

This PR also refactors/removes code

no

na

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-1577

https://news-revenue-hub.atlassian.net/browse/DEV-2042
This would handle the flows around publishing based on


gurjittoptal (approved):

spa/src/components/errors/ChunkErrorBoundary.js
```diff
@@ -19,6 +21,10 @@ const ChunkErrorFallback = () => {
         >
           Refresh
         </S.Button>
+
+        <ErrorS.Logout onClick={logout} whileHover={{ scale: 1.05, x: -3 }} whileTap={{ scale: 1, x: 0 }}>
+          Sign out
+        </ErrorS.Logout>
```
benjaminEwhite, line 27: Was unable to test this locally because I don't know how to get app in this state.

This should also be tested.

benjaminEwhite (changes requested):
Left a couple of inline comments.

I ran locally and tried two different paths. 

First, I tried setting a payment provider as unverified and removing stripe_product_id (both in Django admin). After doing that, I was able to edit pages and do everything expected from dashboard. However, I was not prevented from publishing pages, which is one of the AC.

Second, I tried deleting a payment provider entirely (after temporarily changing some backend code in DonationPageFullDetailSerializer, which would otherwise fail if there is no payment provider).  In this state, I was still able to navigate around the dashboard, but preview of the donation page in editor has a bug.

Screen Shot 2022-08-01 at 4 17 13 PM

In this state, I was also able to publish a page, and in both cases when trying to access the published pages via public-facing URL, I got a 404.

spa/src/components/errors/ChunkErrorBoundary.js
gurjittoptal: cypress test has been added for this.

spa/src/components/dashboard/Dashboard.js
```diff
@@ -59,40 +42,36 @@ function Dashboard() {
     <S.Outer>
       <DashboardTopbar isEditPage={isEditPage} />
       <S.Dashboard data-testid="dashboard">
-        {isEditPage ? null : getShouldAllowDashboard() && <DashboardSidebar />}
```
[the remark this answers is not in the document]
gurjittoptal: Agreed and fixed.

gurjittoptal:
@benjaminEwhite 
The scope of this ticket changed as backend work is needed for non-payment-provider RP support/ other case.

So this PR cleans-up/ factors code to allow users to login without seeing an error message
Ref: 
https://news-revenue-hub.atlassian.net/browse/DEV-1577?focusedCommentId=49228


What you mentioned above can be handled will be handled in ticket below
https://news-revenue-hub.atlassian.net/browse/DEV-2042

benjaminEwhite (changes requested):
Test looks good.

Locally, I tested live donation pages with payment provider stripe verified set to False, and deleting payment provider. In both cases, you get a 404 page, which seems right.

Testing the org portal app with an RP admin user, I was still able to publish/add pages when stripe verified is set to False. Not sure if this is desired behavior, but seems like might not be ideal because an org user could think they have a live page but end user gets 404 because stripe is not verified.

Also tried deleting RP's payment provider, and logging in as RP admin. I was able to load initial dashboard screen, but get blank screen with empty error message when accessing page editor. Am able to click to create new page, but after initial modal submits, there's error in console, and the modal remains, with no indication to user that something has gone wrong.

Screen Shot 2022-08-03 at 11 00 41 AM

Screen Shot 2022-08-03 at 11 05 36 AM

benjaminEwhite (approved):
I'm 👍 on this, with the understanding that [DEV-2042](https://news-revenue-hub.atlassian.net/browse/DEV-2042) is a follow up task, that is inclusive of merging in this PR: https://github.com/newsrevenuehub/rev-engine/pull/577

x110dc (commented):
LGTM

PR #558, opened 2022-07-27
DEV-841: Unsaved Changes modal

GuiMend:

N/A

#### What's this PR do?
As a user, when you edit a page, if you have made any changes but haven't saved yet and click the back button this PR now interrupts that action and shows a modal informing that the user has unsaved changes and asks for a confirmation to exit without saving.

The "save" experience is kind of weird, especially for pages that are not live. This is a simple improvements to make the process a little more intuitive.

- Login to an account that has permission to edit pages
- Choose a page to edit: example `/dashboard/edit/billypenn/donate`
- Enter edit mode and make any edits to the page
- You will notice the "save" button is now enabled
- Click the "back" button, a modal should pop up instead of you being redirected
   - If you click "cancel", close modal and continue in the edit page
   - If you click "yes, exit", close modal and redirect user back to previous page

image

Yes

Added a warning so that users don't leave the edit page without saving

No

No

https://news-revenue-hub.atlassian.net/browse/DEV-841

No

Not sure


gurjittoptal:
Functionality looks good to me.

A few points on UI/ reuseability.
1. getUserConfirmation is used in the 3 other modals. We should either reuse it for unsaved-changes-alert OR refactor all for consistency
ref https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/MainLayout.js#L33
2. We should use existing assets OR html elements instead of @material-ui/icons. I we plan to use material-ui/icons we should replace all other icons and just use these
3. The look and feel should be consistent to the rest, ie no rounded borders and primary color ( blue ) for buttons

spa/package.json
```diff
@@ -11,6 +11,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.15.3",
     "@fortawesome/react-fontawesome": "^0.1.14",
     "@material-ui/core": "^4.11.4",
+    "@material-ui/icons": "^4.11.3",
```
gurjittoptal, line 14: I think we should use existing icons/ assets here. This shall remove both package.json and package-lock.json changes from this PR

GuiMend:
> Functionality looks good to me.
> 
> A few points on UI/ reuseability.
> 
> 1. getUserConfirmation is used in the 3 other modals. We should either reuse it for unsaved-changes-alert OR refactor all for consistency
>    ref [https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/MainLayout.js#L33](https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/MainLayout.js?rgh-link-date=2022-07-27T23%3A01%3A20Z#L33)
> 2. We should use existing assets OR html elements instead of @material-ui/icons. If we plan to use material-ui/icons we should replace all other icons and just use these
> 3. I feel the look and feel should be consistent to the rest, ie no rounded borders and primary color ( blue ) for buttons.

Thanks for the comments @gurjittoptal. I'll see what I can do with the `getUserConfirmation`.

Regarding 2. and 3. I spoke with Rachel about the design and she told me that we are going through a rebranding and that a lot of things will change in the design: 
   2. The icons she mentioned that we are using a mix of custom icons + material icons, in that case I think it makes sense to use the `@material-ui/icons` library
   3. The colors are also going to change (or at least we will add new ones).

You can check the JIRA ticket, there is a figma link with the expected design for the modal and I tried to make it as pixel perfect as I could.

spa/package.json
gurjittoptal: resolved based on comments below.

GuiMend:
Spoke with @gurjittoptal about **1. getUserConfirmation is used in the 3 other modals. We should either reuse it for unsaved-changes-alert OR refactor all for consistency** and as we are changing designs we won't do the refactoring of this piece of the code right now. It would probably need a separate ticket for changing the designs of the existing modals together with the refactoring.

gurjittoptal (approved):
Looks good to me.

x110dc (approved):

PR #560, opened 2022-07-29
add a salt to the uid hash

x110dc:
#### What's this PR do?

- adds a salt to the function that produces a hash 

- to make it more difficult for a bad actor to determine the original string given the hash

- make a contribution; examine the uid in the thank-you URL
- make a contribution on a different page but with the same email; confirm that the uid is the same

No, the existing tests cover this.

N/A

No.

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-1982

No.

No.


njharman (approved):
Did not complete pre-review checklist

There's no concern reusing the Django secret key? I don't know, security is hard. 

There's no concern truncating hash to 10chars which is 5 bytes of original hash due to hexdigest? I don't know, security is hard. https://news.ycombinator.com/item?id=14952547#:~:text=SHA%2D256%20has%20a%20256,)%20to%20find%20a%20collision.)

I would have added test get_sha256_hash(string)  != hashlib.sha256(string)  # i.e. there's a salt.

PR #566, opened 2022-08-01
DEV-1808_test_coverage

njharman:
#### What's this PR do?

Increase test coverage. Was suppose to get to 98% overall. Think got to 97%. Changes growing too large for review / commit so submitting now. Can follow up with more test coverage in follow-on ticket.

Fixed few test configuration issues.  Then sorted all files by test coverage and started adding coverage to the least covered "group". E.g. all the ./apps/pages/*

- Move pytest.ini options to Makefile run-tests so every incarnation of
  pytest is not forced to run coverage and be triply verbose.
- Add ./revengine to directories scanned for tests.
- Fix broken revengine/tests/test_django_reversion_setup.py test.
- Enable running tests from any path (only partially, wasn't worth effort to fix everything).
- Load fixtures once per TestClass instead of once per test_method. (should be used more)
- Include ./revengine/ in coverage report

Test Coverage

- apps/common/validators.py to 100%
- apps/pages/admin.py to 100%
- apps/pages/helpers.py to 100%
- apps/pages/models.py to 100%
- apps/pages/serializers.py to 95% [TODO: _update_related() looks annoying to test]
- apps/pages/views.py to 94% [TODO: ask about request.FILES]

Being on the cusp of our 95% coverage minimum. Meant devs were being blocked from completing critical tickets on time. 
Higher test coverage is almost always better.

N/A

Yes.

N/A

Much of these lacking coverage was in failure branches (exceptions, errors, etc)

Adding coverage after code is written / just to increase coverage results in poorer tests than those made with knowledge of original code.

Features (to make simpler/more complete) of pytest are not being utilized. We now have horrid mix of pyunit and pytest style tests. 

Complex tests, with tons of setup, etc. are created instead of taking that as a sign to refactor code, reducing interdependencies and state, to create (more) isolated "units" of code. Which, tbf, tends to be hard with frameworks and we're using many of which layered on top of each other. django ORM, base django, "advanced" django like model view sets, and REST framework.

N/A

[DEV-1808]

No.

Yes, optionally another ticket [DEV-2196] to increase coverage further. 

[DEV-1808]: https://news-revenue-hub.atlassian.net/browse/DEV-1808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Makefile
```diff
@@ -46,7 +46,7 @@ run-redis:
 
 run-tests:
 	make test_migrations
-	pytest
+	pytest --reuse-db -p no:warnings -vvv --cov-config=.coveragerc --cov-report=html --cov=apps
```
x110dc, line 50: I ran this w/ and w/o `-p no:warnings` and got the same output. What is that flag helping us with? Is there a danger we could miss something we need to notice, like a deprecation warning?

x110dc (changes requested):
Our least-covered file is still 90%. That's great! I just had one small question.

Makefile
njharman: These are the settings that I removed from the .pytest.ini

Surprised output is same. There's usually tons of warnings from 3rd party libs. That are annoying and hence suppressed.

I have no feelings if -p no:warnings kept here or not.

x110dc (approved):
LGTM.

x110dc (approved):
LGTM

PR #569, opened 2022-08-02
DEV-2133 | Make ReactAppView not fail when no social_meta

benjaminEwhite:

#### What's this PR do?

Fixes a bug in ReactAppView whereby the view would fail when a RevenueProgram didn't have an associated SocialMeta instance (which lies in one-to-one relation).

Adds a test for this case.

So donation pages won't be broken for RPs that don't have social meta.

In NRH admin, find the RP for a live donation page. Delete that RP's social meta instance (which can be done from the detail view for the RP).  

Now visit the donation page. It should load. If you inspect the head of the HTML file in developer tools, none of the social meta related values should appear (the values are pictured below, but this is what you shouldn't see on this page):

Screen Shot 2022-08-02 at 2 39 22 PM

Yes

N/A

No

N/A

[DEV-2133](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-2133)

No

No


njharman (approved):
Looks good. Tests pass. Could not check coverage because ./revengine/ isn't part of test coverage (fix that in other ticket/follow on to DEV-1808)

The except branch is covered so lgtm!

njharman:
Needs to have main merged into it

pyproject.toml
```diff
@@ -71,6 +71,7 @@ whitenoise = "^6.2.0"
 django-reversion-compare = "^0.15.0"
 django-migration-linter = "^4.1.0"
 django-test-migrations = "^1.2.0"
+pytest-mock = "^3.8.2"
```
x110dc, line 74: This should be a dev dep, no?

x110dc (changes requested):
One minor request.

pyproject.toml
x110dc: This is here but it's also in dev deps. I removed it from here.

x110dc (approved):

PR #572, opened 2022-08-03
DEV-2021: fix current csp violations, remove unsafe allowances

benjaminEwhite:

#### What's this PR do?

- Tightens CSP settings by including a nonce in `style-src` and `script-src` and removing `unsafe-inline` and `unsafe-eval` settings
- Adds `https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.4.1/` as a font and style source to solve for violations pertaining to that library.
- Refactors how spa env vars are added to template. Removes inclusion tag and opts for adding spa env as context in the ReactAppView template (this is because with the inclusion tag approach, for some unknown reason, the nonce values always appeared blank, causing a csp violation.
- Adds `window.__webpack_nonce__` to index.html in order to solve 2 CSP violations related to styled-components.
- Removes the report only setting, since all violations are solved for.

- Safer content security policy
- Removes possibly concerning logs from dev tools console

1.  Load any page from SPA in browser. For instance, [this one](https://billypenn-dev-2021.revengine-review.org/donate)
2. Open DevTools and click on the console.
3. You shouldn't see any CSP warnings.

Here's how it looks before this change (the red "[Report Only]" entries are the CSP violations):

Screen Shot 2022-08-03 at 9 17 13 AM

And here's how it looks after:

Screen Shot 2022-08-03 at 9 41 42 AM

No. Absent adopting true end to end solution like Selenium, there's not an obvious way to unit test for CSP violations at the moment, as that would require testing that spans the front and backend.

N/A

No

No

[DEV-2021](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-2021)

No

N/A


njharman (approved):
Don't feel particularly qualified to judge but, lgtm. Tests pass.

Congrats on removing the gross things!

x110dc (approved):

PR #580, opened 2022-08-04
use cf-connecting-ip and x-forwarded-for to fetch IP

x110dc:
#### What's this PR do?

- adds a new utility function to determine the IP to capture from an incoming request
- uses that function in the two places that `REMOTE_ADDR` was being called
- allows the root log level to be modified at run time with the env variable `LOG_LEVEL`

- we'd like to know the actual originating IP of a request; that's sometimes in REMOTE_ADDR, but REMOTE_ADDR will show the last in the chain of proxies if there are proxies; `CF-Connecting-IP` and `X-Forwarded-For`, if present will show the actual original IP
- I didn't realize until working on this PR that we didn't have a way to generate debug logs so added that

- you could potentially try to set up or use a proxy to connect to this locally and confirm that it's retrieving the right IP from the headers; I'll enable debugging in the review app so you can check the logs there to confirm that it's retrieving the IP from the correct header

Yes.

N/A

No.

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-2006

No.

N/A


njharman (approved):
100% coverage. Looks good. Didn't test logging but can't see why it wouldn't work as expected.

The minorest of minor things; in get_original....request() the if, if/else structure is "odd". can't see why it's not three ifs or if/elif/else.

x110dc:
> 100% coverage. Looks good. Didn't test logging but can't see why it wouldn't work as expected.
> 
> The minorest of minor things; in get_original....request() the if, if/else structure is "odd". can't see why it's not three ifs or if/elif/else.

I see what you mean. I just removed the redundant `else` since there's a return in the `if` above it.

PR #616, opened 2022-08-25
DEV-2043: API-based password reset flow to support account creation

benjaminEwhite:

N/A

#### What's this PR do?

Adds API support for password reset flow initiated and concluded from the SPA, using [django-rest-passwordreset](https://github.com/anexia-it/django-rest-passwordreset/blob/master/tests/test/test_auth_test_case.py)

This adds three new API endpoints:

- `POST /api/v1/users/password_reset/`: The request body for this endpoint should include an `email` parameter. If the email is for an extant user, a password reset email will be sent to that user, containing a link to the SPA of the form `<spa root url>/password_reset?token=c3671b77b461df3a078bcff86c`.
- `POST /api/v1/users/password_reset/confirm`: The request body for this endpoint should include a `password` parameter containing the new password. It should also contain a `token` parameter containing the token produced in step 1 (and that the SPA gets as a query parameter). Password rest tokens have been configured to expire after 24 hours. In the case of an expired token, this endpoint will return a 404. In case of success, it will reset the password.
- `POST /api/v1/users/password_reset/validate_token/`: This can be used to determine if a token is (still) valid. If so, it will be 200/ok. If not it will return a 404 / not found.  Strictly speaking, this endpoint is not necessary to complete the flow, but it comes for free with django-rest-passwordreset and might prove useful.

To support user account creation flow. 

These changes are API only, so manual test steps reflect this reality.

1. Open Postman.
2. Make a POST request to https://dev-2043.revengine-review.org/api/v1/users/password_reset/ including a JSON request body of the form `{"email": "your-staging-user-email-address@foo.com"}`.
3. Check your email. Copy the link address from the p/w reset link. This link is for the SPA, which is not yet configured to handle password reset, so for now, just copy the token value from the query parameters.
4. Back in Postman, make a POST request to https://dev-2043.revengine-review.org/api/v1/users/password_reset/confirm/ including a JSON request body of the form `{"password": "<your-new-password>", "token": "<your token">}`. This should return 200 success.
5. Log in to the admin or dashboard using your account and the new password.

To test the "200 when no user with email" flow, try using an email address for which there is not a user (for instance, you might use your NRH address with a + suffix like `me+something@fundjournalism.org`). Postman should still get a 200, but you shouldn't receive a reset email at that address.

No.

Yes. Minimal tests added, along with note in tests about why we're not testing the full flow (since that would be testing the dependency itself, which is already tested).

N/A

No.

No.

[DEV-2043](https://news-revenue-hub.atlassian.net/browse/DEV-2043)

No.

N/A


njharman:
Is sending emailaddy to open endpoint really how these things work? I don't know... just assumed someone smarter than me did something, er, better.  Seems so prone to DOS. It's not that uncommon to have / get people's emails.

apps/users/views.py
```diff
@@ -210,3 +214,35 @@ def get_permissions(self):
...
+    This causes an email to be sent to user who requested pw reset.
+
+    https://github.com/anexia-it/django-rest-passwordreset#example-for-sending-an-e-mail
+    """
+    email = reset_password_token.user.email
+    spa_reset_url = "{}{}?token={}".format(
+        instance.request.build_absolute_uri(reverse("index")),
+        SPA_PASSWORD_RESET_ROUTE,
+        reset_password_token.key,
+    )
```
njharman, line 236: potentially similar need in DEV-2053. I created new routes

```
urlpatterns += [
    path("cloudflare-500", cloudflare_500_view, name="cloudflare-500"),
    # React SPA:
    path("", index, name="index"),  # for reverse()
    path("verified", index, name="spa_account_verification"),  # for reverse()
```
Now that I look at it, I'm fairly certain the more specific routes should come first... Although since they all route to same place it do not matter.

apps/users/views.py
```diff
@@ -210,3 +214,35 @@ def get_permissions(self):
...
+
+    This causes an email to be sent to user who requested pw reset.
+
+    https://github.com/anexia-it/django-rest-passwordreset#example-for-sending-an-e-mail
+    """
+    email = reset_password_token.user.email
+    spa_reset_url = "{}{}?token={}".format(
+        instance.request.build_absolute_uri(reverse("index")),
+        SPA_PASSWORD_RESET_ROUTE,
+        reset_password_token.key,
```
njharman, line 235: Should token be urlsafed? or is it already guarnteed to be by 3rdp lib

apps/users/tests/test_views.py
```diff
@@ -109,6 +112,40 @@ def test_happy_path(self):
...
+        """Show that we get a 200, and that email containing link with reset token gets sent"""
+        user = create_test_user()
+        response = self.client.post(self.url, {"email": user.email})
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(self.mailbox), 1)
+        token = ResetPasswordToken.objects.get(user=user)
+        # we get the html version of the email
+        link = BeautifulSoup(self.mailbox[0].alternatives[0][0], "html.parser").a
+        self.assertIsNotNone(link)
+        self.assertIn(token.key, link.attrs["href"])
```
njharman, line 138: I found email with template typo.  Test that ensures no { or } are present? Might catch something.

njharman (approved):
looks good. 100% covered

Lots of "please consider" but no required changes. Pre-approving whatever decision you make on those.

[ I wish there was an easy way to discover tasks (in the backlog) that are gonna be similar in the back end and discuss/design them together. I now need to change things in DEV-2053 to be more in line with this. ]

apps/users/constants.py
```diff
@@ -23,3 +23,4 @@
     PASSWORD_TOO_SHORT_VALIDATION_MESSAGE,
     PASSWORD_TOO_SIMILAR_TO_EMAIL_VALIDATION_MESSAGE,
 ]
+SPA_PASSWORD_RESET_ROUTE = "password_reset"
```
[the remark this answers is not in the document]
benjaminEwhite: Think that stems from I initially thought would "roll my own" solution, before discovering django-rest-password-reset.  Removing the constant.

apps/users/views.py
benjaminEwhite: [Looking at implementation](https://github.com/anexia-it/django-rest-passwordreset/blob/1c5bbc324a1e3859536299cf73099fcbda1ce3e6/django_rest_passwordreset/tokens.py#L52), I believe this should be safe.

benjaminEwhite:
@njharman re: 

> Is sending emailaddy to open endpoint really how these things work? I don't know... just assumed someone smarter than me did something, er, better. Seems so prone to DOS. It's not that uncommon to have / get people's emails.


AFAICT, yes that's the way this is typically handled. And not sure how else one really could handle since typical user in need of reset won't be authenticated, and will only have their email address as info they can provide.

We could look into throttling p/w reset endpoint, but I believe that would require rolling our own solution, or else forking https://github.com/anexia-it/django-rest-passwordreset, as it does not support configurable throttling out of the box.

I'm going to leave as is for now and let @x110dc weigh in on this when he reviews.

apps/emails/templates/nrh-org-portal-password-reset.html
```diff
@@ -0,0 +1,6 @@
+<p>Hello,</p>
+<br>
+<p>We received a request to reset the password for the RevEngine account associated with {{ email }}.<p>
```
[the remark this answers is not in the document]
benjaminEwhite: Updating accordingly.

apps/emails/templates/nrh-org-portal-password-reset.txt
```diff
@@ -0,0 +1,7 @@
+Hello,
+
+We received a request to reset the password for the RevEngine account associated with {{ email }}.
```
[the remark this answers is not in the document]
benjaminEwhite: Updating accordingly.

apps/users/constants.py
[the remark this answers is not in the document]
benjaminEwhite: Removing.

apps/users/tests/test_views.py
benjaminEwhite: @njharman I didn't see a comment re: typo. Can you point me to it?

apps/users/views.py
```diff
@@ -210,3 +214,35 @@ def get_permissions(self):
...
+    )
+    context = {
+        "email": email,
+        "reset_password_url": mark_safe(spa_reset_url),
+    }
+    logger.info("Sending password reset email to %s with the following reset url: %s", email, spa_reset_url)
+    send_templated_email(
+        email,
+        "Reset your password",
```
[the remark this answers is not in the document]
benjaminEwhite: Added site name.

apps/users/views.py
benjaminEwhite: I don't think we need anything more specific in URL config. The p/w reset is handled in the SPA, so as long as the URL gets picked up by `index`, the SPA should know what to do with the remaining path.

apps/users/tests/test_views.py
njharman: I'm sorry, was not clear.  In a ticket I was working on "I found template typo" So added test to fail if "{" or "}" was in rendered  template text.  Fragile, and wouldn't catch all typo's. But better than nothing? Was pointing out for your consideration to add similar test here.  But given frequency of template edits not sure it's worth bother. But, given embarrassment of messed up email to contributors maybe it is?

apps/users/views.py
njharman: Ignore "more specific routes" comment (we thinking/saying same, I just use less good words).  

I was mainly pointing out way to use django's reverse() instead of rolling own URL (which is more prone to bugs, not by you, but future refactorer). Although, I believe you'd still have to add URL params. So,....?

apps/users/tests/test_views.py
x110dc: Was this resolved @benjaminEwhite ?

njharman: I'm gonna answer "yes". Because it was just a "hey, I found this additional thing which may or may not be useful to you" And not an "issue" requiring action.

x110dc (approved):

PR #618, opened 2022-08-25
DEV 2220 cancel update not pk

x110dc:

#### What's this PR do?

- moves handling of Subscription cancellation and update out of PaymentManager and into an APIView
- adds a Subscription ID to the contributions serializer 

Since we changed to show payments directly from Stripe instead of from Contributions the API needed to be updated to handle updating the payment method and cancelling for Subscriptions that may or may not have an associated Contribution. The Subscription Id was added to the serializer so the API would know the correct object to update/cancel.

Since these changes are only for the API and the frontend will be updated in subsequent ticket this can't be tested with the SPA directly

1. Create a recurring contribution
2. Request a magic link
3. From the results, pull out the JWT, the Subscription ID and the CSRF token and plug them in here:
```
curl -v -F 'revenue_program_slug=billypenn' --header "X-CSRFToken: blah"  -X DELETE --cookie "Authorization=quux" --cookie "csrftoken=blah" http://localhost:8000/api/v1/subscriptions/sub_baz/
```
Then to update the payment method, first create a new one: 
```
stripe payment_methods create --type=card -d "card[number]=4242424242424242" -d "card[exp_month]=8" -d "card[exp_year]=2027" -d "card[cvc]=234" --stripe-account=acct_quux
```

Then update similarly: 
```
 curl -v -F 'revenue_program_slug=billypenn' -F 'payment_method_id=pm_1LavTgPbhdfvD1kVhmGQNdhR' --header "X-CSRFToken: foo" -X PATCH --cookie "Authorization=bar" --cookie "csrftoken=baz" http://localhost:8000/api/v1/subscriptions/sub_swizzle
```

 No.

Yes.

N/A

- I added more tests in pytest format; so there's a mix of test styles. 
- I wasn't how best to structure the endpoints and verbs; I went with what seemed most RESTful: a DELETE and a PATCH on the bare identifier. 
- Likewise, I wasn't sure what response codes to use. 
- I wasn't sure how to use the DRF permission framework (other than confirming a Contributor) so I just did a bare-simple check of the email against the email for the customer associated with the Subscription. I created a ticket to improve this.
- I wasn't sure about requiring the revenue program as a parameter on the call or inferring it from the subdomain; it looks like both of these methods happen in the code in different places; not sure pros/cons of each but went with the parameter

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2220

No.

- https://news-revenue-hub.atlassian.net/browse/DEV-2287 
- https://news-revenue-hub.atlassian.net/browse/DEV-2286


apps/contributions/serializers.py
```diff
@@ -474,6 +474,7 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
     # id will be charge object id in our case, which will start with ch_ and doesn't exceed 255 chars
     # https://stripe.com/docs/upgrades#what-changes-does-stripe-consider-to-be-backwards-compatible
     id = serializers.CharField(max_length=255)
+    subscription_id = serializers.CharField(max_length=255, required=False, allow_blank=True)
```
njharman, line 477: help_text, comment or rename to make it clear this is a Stripe ID and not something NRE related.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -229,6 +235,7 @@ def upsert(self, contributions):
         """Serialized and Upserts contributions data to cache."""
         data = self.serialize(contributions)
         cached_data = json.loads(self.cache.get(self.key) or "{}")
+
```
njharman, line 238: boo!

apps/contributions/views.py
```diff
@@ -331,50 +332,71 @@ def process_flagged(self, request, pk=None):
...
+        if request.user.email.lower() != subscription.customer.email.lower():
+            # TODO: [DEV-2287] should we find a way to user DRF's permissioning scheme here instead?
+            return Response({"detail": "Forbidden"}, status=status.HTTP_403_FORBIDDEN)
+        try:
+            stripe.Subscription.delete(
+                subscription_id, stripe_account=revenue_program.payment_provider.stripe_account_id, expand=["customer"]
+            )
+        except stripe.error.StripeError:
+            logger.exception("stripe.Subscription.delete returned a StripeError")
+            return Response({"detail": ""}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
njharman, line 400: Should detail be something?

apps/contributions/tests/test_views.py
```diff
@@ -758,75 +760,99 @@ def setUp(self):
...
+            subscription_id=self.subscription_id,
+            data={
+                "payment_method_id": self.payment_method_id,
+                "revenue_program_slug": self.revenue_program.slug,
+            },
+        )
+        assert response.status_code == 403
+        assert response.data["detail"] == "Forbidden"
         mock_modify.assert_not_called()
         mock_attach.assert_not_called()
```
njharman, line 814: The pytest way would be
> assert not mock_modify.called

apps/contributions/tests/test_views.py
```diff
@@ -758,75 +760,99 @@ def setUp(self):
...
+    @mock.patch("stripe.Subscription.retrieve")
     @mock.patch("stripe.PaymentMethod.attach")
     @mock.patch("stripe.Subscription.modify")
-    def test_failure_when_any_parameter_other_than_pm_id(self, mock_modify, mock_attach):
-        response = self._make_request(self.contribution, {"amount": 10})
+    def test_failure_when_any_parameter_other_than_pm_id(self, mock_modify, mock_attach, mock_retrieve):
+        response = self._make_request(
+            subscription_id=self.subscription_id,
+            data={
+                "fiddle": self.payment_method_id,
```
njharman, line 790: is "fiddle" well know "bad value"?  I typically use something like "test_unknown_parameter"

njharman (approved):
lgtm, 100% coverage
Approved with commentary 

> I wasn't sure about requiring the revenue program as a parameter on the call or inferring it from the subdomain; it looks like both of these methods happen in the code in different places; not sure pros/cons of each but went with the parameter

Is the API served from RP subdomains? I thought that was only SPA.  HTTP_REFFERER is not reliable. That is why I required it as param for magic email. 

fwiw 204 is more correct than 200 for DELETE and PATCH (with no content). I believe the Proper PATCH response would be 200 and return the full modified object. But Proper REST is so unpragmatic for APIs (as opppsed to HyperText Documents).

apps/contributions/serializers.py
x110dc: done

apps/contributions/stripe_contributions_provider.py
x110dc: removed.

apps/contributions/tests/test_views.py
x110dc: I was wondering about that. Thanks!

apps/contributions/views.py
x110dc: I changed it to "Error"

apps/contributions/tests/test_views.py
x110dc: It's not meaningful. When I've exhausted "foo", "bar", "baz" and "quux" then I don't know what to use next so "fiddle" often comes to mind; but yes, I can see the value in adding something descriptive there. Will change.

PR #627, opened 2022-08-31
DEV 2289 separate API endpoint for Subscriptions

x110dc:
#### What's this PR do?

- moves AttrDict into a utility file
- creates new endpoints for Stripe Subscriptions (and new Serializer) 
- changes the existing Contribution endpoints to use Stripe Payment Intents instead of Charges
- adds is_modifiable and is_cancellable properties the Contribution objects
- stores Subscriptions (in addition to Payment Intents) in cache when magic link is requested

As it is now the Contributions endpoints are conflating two things: payments that have already happened and recurring charges; this meant that to cancel or modify recurring contributions you acted on past contributions that may have already been canceled and could act on any of them. But you're not canceling or modifying past Contributions. So if you select a past monthly (already paid) contribution and modify the payment method? What are you modifying? Not that past payment but future payments in the Subscription. It felt cleaner to offer two endpoints for the frontend to use for this. Also, there was no way to indicate a Subscription had been canceled or if it was cancelable or modifiable. That's been added to both endpoints (which the frontend and the UX catches up). Later the `is_` properties can be removed from Contributions (and left only on Subscriptions). I also tried to make it RESTful using PATCH and DELETE. 

This is all backend changes, so need to test with `curl` or Postman or similar. So to fetch 
```
curl -X GET -v 'http://localhost:8000/api/v1/subscriptions/sub_1234/?revenue_program_slug=billypenn' \
  -H 'Accept: application/json, text/plain, */*' \
  -H "Content-Type: application/json" \
  -H 'Cookie: csrftoken=<...>; Authorization=<...>' \
  -H 'X-CSRFTOKEN: <...>' \
  --insecure | jq .
```
And similarly for list: `curl -X GET -v 'http://localhost:8000/api/v1/subscriptions/?revenue_program_slug=billypenn' ... | jq .`
To modify: 
```
curl 'http://localhost:8000/api/v1/subscriptions/sub_1234/'  -d '{"revenue_program_slug":"billypenn", ... "payment_method_id":"pm_1234"}' 
```
And to cancel:
```
curl -X DELETE -v 'http://localhost:8000/api/v1/subscriptions/sub_1234/' ... 
-d '{"revenue_program_slug":"billypenn", "payment_method_id":"pm_1234"}'
```

No.

Yes.

N/A

I don't think so, except maybe leaving `is_modifiable` and `is_cancelable` on Contributions. But I created a ticket to remove those. 

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-2289
This should also satisfy the https://news-revenue-hub.atlassian.net/browse/DEV-2193 

No.

- https://news-revenue-hub.atlassian.net/browse/DEV-2320
- https://news-revenue-hub.atlassian.net/browse/DEV-2329
- https://news-revenue-hub.atlassian.net/browse/DEV-2328


apps/contributions/serializers.py
```diff
@@ -471,21 +473,111 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
...
+    created = serializers.SerializerMethodField()
+    last_payment_date = serializers.SerializerMethodField()
+
+    def get_id(self, instance):
+        return instance.id
+
+    def get_card_brand(self, instance):
+        return self._card(instance).brand
+
+    def get_next_payment_date(self, instance):
```
njharman, line 529: https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2065039378/Python+Tips

apps/contributions/serializers.py
```diff
@@ -471,21 +473,111 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
...
+        return True
+
+    def get_interval(self, instance):
+        interval = instance.get("plan", {}).get("interval")
+        interval_count = instance.get("plan", {}).get("interval_count")
+        if interval == "year" and interval_count == 1:
+            return ContributionInterval.YEARLY
+        if interval == "month" and interval_count == 1:
+            return ContributionInterval.MONTHLY
+        raise serializers.ValidationError(f"Invalid interval: {interval}/{interval_count}")
```
njharman, line 568: Maybe also include instance["plan"] for context?

apps/contributions/serializers.py
```diff
@@ -471,21 +473,111 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
...
     interval = serializers.ChoiceField(choices=ContributionInterval.choices)
     revenue_program = serializers.CharField(max_length=63)
     amount = serializers.IntegerField()
     provider_customer_id = serializers.CharField(max_length=255)
     credit_card_expiration_date = serializers.CharField(max_length=7)
     created = serializers.DateTimeField()
     last_payment_date = serializers.DateTimeField()
+
+
+class SubscriptionsSerializer(serializers.Serializer):
```
njharman, line 502: new code 100% covered good job! But some easy coverage additions would 100% this file.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -34,36 +37,52 @@ class InvalidMetadataError(ContributionIgnorableError):
...
+            return False
+        if self.payment_intent.invoice.subscription.status in [
+            "incomplete",
+            "incomplete_expired",
+            "canceled",
+            "unpaid",
+        ]:
+            return False
+        # statuses "trialing", "active", and "past_due" are cancelable
+        return True
```
njharman, line 72: Is it worth checking that with if and logging/raising error if an unknown status is given to us?

apps/contributions/stripe_contributions_provider.py
```diff
@@ -242,9 +249,50 @@ def upsert(self, contributions):
...
+                logger.warning("Unable to process Subscription [%s] due to [%s]", subscription.id, type(ex))
+        return data
+
+    def upsert(self, subscriptions):
+        """Serialized and Upserts subscriptions data to cache."""
+        data = self.serialize(subscriptions)
+        cached_data = json.loads(self.cache.get(self.key) or "{}")
+        cached_data.update(data)
+
+        with self.cache.lock(f"{self.key}-lock"):
```
njharman, line 285: I'm surprised to see lock().  the caches I know memcache/redis set() is atomic?

I don't see lock() in docs https://docs.djangoproject.com/en/3.2/topics/cache/#cache-key-transformation
Nor could I find it in django/core/cache/*py 

What am I missing?

apps/contributions/views.py
```diff
@@ -333,15 +335,53 @@ def process_flagged(self, request, pk=None):
...
         IsAuthenticated,
     ]
+    serializer_class = serializers.SubscriptionsSerializer
 
-    # This is used to update the payment method for a Subscription
-    @action(methods=["patch"], detail=True, permission_classes=[IsAuthenticated])
-    def patch(self, request, subscription_id):
+    def retrieve(self, request, pk):
+        revenue_program_slug = request.query_params.get("revenue_program_slug")
+        revenue_program = RevenueProgram.objects.get(slug=revenue_program_slug)
```
njharman, line 346: Should we not 404 if rp matching slug not found?

apps/contributions/views.py
```diff
@@ -333,15 +335,53 @@ def process_flagged(self, request, pk=None):
...
+        cache_provider = SubscriptionsCacheProvider(self.request.user.email, revenue_program.stripe_account_id)
+        subscriptions = cache_provider.load()
+        if not subscriptions:
```
njharman, line 349: Instead of just returning 404 (when we haven't loaded from cache yet) should we return different message so UI can display "loading, refresh in X" or whatever?

apps/contributions/views.py
```diff
@@ -333,15 +335,53 @@ def process_flagged(self, request, pk=None):
...
+            )
+        try:
+            subscription = [
+                x
+                for x in subscriptions
+                if x.get("revenue_program_slug") == self.request.query_params["revenue_program_slug"]
+                and x.get("id") == pk
+            ][0]
+        except IndexError:
+            return Response({"detail": "Not Found"}, status=status.HTTP_404_NOT_FOUND)
```
njharman, line 361: Wouldn't this be less convoluted and weird? (or see next comment about duplicated code) and keep the index exception around deduplication/extract method get_me_subscriptions_for()[0]
```
for x in subscriptions:
  if x.get... and x.get..:
    return Response(x)
return 404
```

apps/contributions/views.py
```diff
@@ -333,15 +335,53 @@ def process_flagged(self, request, pk=None):
...
+        subscriptions = cache_provider.load()
+        if not subscriptions:
+            task_pull_serialized_stripe_contributions_to_cache(
+                self.request.user.email, revenue_program.stripe_account_id
+            )
+        subscriptions = [
+            x
+            for x in subscriptions
+            if x.get("revenue_program_slug") == self.request.query_params["revenue_program_slug"]
+        ]
```
njharman, line 377: Lot of duplicated code between this and retrieve()

apps/contributions/views.py
```diff
@@ -379,22 +419,22 @@ def patch(self, request, subscription_id):
...
+        """
         revenue_program_slug = request.data.get("revenue_program_slug")
         revenue_program = RevenueProgram.objects.get(slug=revenue_program_slug)
         # TODO: [DEV-2286] should we look in the cache first for the Subscription (and related) objects?
         subscription = stripe.Subscription.retrieve(
-            subscription_id, stripe_account=revenue_program.payment_provider.stripe_account_id, expand=["customer"]
+            pk, stripe_account=revenue_program.payment_provider.stripe_account_id, expand=["customer"]
         )
+
         if request.user.email.lower() != subscription.customer.email.lower():
```
njharman, line 433: I know you didn't touch this code. But, maybe add todo/or acknowledgement its a limitation we are accepting. I notice that if user changes email in our or stripe? they may be unable to cancel subscription?

njharman (changes requested):
100% coverage!

1. Would prefer the follow up dev-2320 ticket referred back to this one
2. PR note: Doesn't this require changes to user facing docs? Even if not the front implementation?

Don't forget to log your time on ticket.

apps/common/utils.py
```diff
@@ -127,3 +127,17 @@ def get_original_ip_from_request(request):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__dict__ = self
+
+    @classmethod
+    def construct_from_dict(cls, data):
+        if isinstance(data, dict):
+            return cls({key: cls.construct_from_dict(data[key]) for key in data})
+        if isinstance(data, list):
+            return [cls.construct_from_dict(i) for i in data]
```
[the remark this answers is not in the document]
x110dc: I didn't author it, I just moved it! 😉 But point taken. I found a library that does just this that served as drop-in replacement. I think I actually prefer the method of testing in Stripe objects in the webhook receiver, but I was too lazy to wholesale replace what's here, but I'll open a separate ticket to look into that.

apps/contributions/serializers.py
```diff
@@ -471,21 +473,111 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
...
+        required=True, help_text="if recurring then can the payment method be modified"
+    )
+    is_cancelable = serializers.BooleanField(
+        required=True, help_text="if recurring then can the payment method be canceled"
+    )
     status = serializers.ChoiceField(choices=ContributionStatus.choices)
     card_brand = serializers.ChoiceField(choices=CardBrand.choices, required=False, allow_null=True)
     last4 = serializers.IntegerField()
     payment_type = serializers.ChoiceField(choices=PaymentType.choices, required=False, allow_null=True)
-    next_payment_date = serializers.DateTimeField()
```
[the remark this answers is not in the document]
x110dc: It was never used on the frontend and was always null anyway.

apps/contributions/serializers.py
```diff
@@ -471,21 +473,111 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
...
+
+    def get_is_modifiable(self, instance):
+        if instance.status in ["incomplete_expired", "canceled", "unpaid"]:
+            return False
+        return True
+
+    def get_is_cancelable(self, instance):
+        if instance.status in ["incomplete", "incomplete_expired", "canceled", "unpaid"]:
+            return False
+        return True
```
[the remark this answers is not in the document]
x110dc: Agreed. I changed.

apps/contributions/serializers.py
```diff
@@ -471,21 +473,111 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
...
+    def get_last_payment_date(self, instance):
+        return datetime.utcfromtimestamp(int(instance.current_period_start))
+
+    def get_created(self, instance):
+        return datetime.utcfromtimestamp(int(instance.created))
+
+    def get_last4(self, instance):
+        return instance.default_payment_method.card.last4
+
+    def _card(self, instance):
```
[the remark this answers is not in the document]
x110dc: That's fine with me (done), and 👍 on updating style guide.

apps/contributions/views.py
x110dc: Not sure that can/should happen, but agreed, good defensive programming anyway. I'll do that.

apps/contributions/views.py
x110dc: But that's not what it does. It only returns the 404 if it's not in the cache AND not found in Stripe either.

apps/contributions/views.py
x110dc: Agreed. On both points. I'll change.

apps/contributions/views.py
x110dc: You're right. I'll make a helper function (and put it at the top of the list).

apps/contributions/views.py
njharman: Is my assumption that task_pull_serialized_stripe_contributions_to_cache() is an asynchronous celery task incorrect?

njharman: only when called with .delay()

apps/contributions/views.py
x110dc: Thanks, that cleaned up those two methods nicely.

apps/contributions/views.py
x110dc: There's no way for a user to change their email. Eventually we'll allow a user to link multiple email addresses together and it'll be addressed then.

apps/contributions/views.py
x110dc: Yes, it's only async when called with `.delay` or `.apply_async`.

apps/contributions/views.py
njharman: I don't believe there's anything stopping them API wise.  there's a PATCH on user and it doesn't block changing email. Just sayin...

apps/contributions/serializers.py
x110dc: But that'll be a big JSON blob. Maybe plan.id?

apps/contributions/serializers.py
njharman: oh, I thought I fixed all of them, back when I made that tip...  Maybe people have been adding wrong since then, horrors!

Def assign ticket to me, I'll do it right quick!

PR #638, opened 2022-09-02
DEV-2329 Do not use utcfromtimestamp

njharman:
#### What's this PR do?
Updated code and tests to use timezone aware datetimes, using datetime.timezone.utc Cut our usage of pytz (deprecated) still used by 3rd party libs (Django). Resolved in Django 4.0 I believe. 

https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

Warning Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing a specific timestamp in UTC is by calling datetime.fromtimestamp(timestamp, tz=timezone.utc).

Cause The Python Authors probably know what they are taking about. timezones cause subtle and hard to track down bugs.

N/A

No.

Existing ones updated.

N/A

No

dev docs https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2047311873/Python+Style+Guide

[DEV-2329]

No

No

[DEV-2329]: https://news-revenue-hub.atlassian.net/browse/DEV-2329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/serializers.py
```diff
@@ -3,6 +3,7 @@
 from django.conf import settings
 from django.db.models import TextChoices
 
+import pytz
```
x110dc, line 6: Can we use zoneinfo instead? https://docs.python.org/3/library/zoneinfo.html 

https://docs.djangoproject.com/en/3.2/topics/i18n/timezones/

I think use of pytz is kinda/sorta deprecated.

x110dc (changes requested):

apps/contributions/serializers.py
njharman: Yes

njharman: Also updated https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2047311873/Python+Style+Guide

x110dc (approved):

PR #646, opened 2022-09-05
expose Stripe Account ID on 'public' endpoint

x110dc:
#### What's this PR do?

Adds the Stripe Account ID and the RP nonprofit status to the "public" endpoint for revenue programs.

- It allows other systems to query for the Stripe Account number in order to link it to a Revenue Program/Org.
- It allows other systems to determine the nonprofit status of an RP.

1. Create a superuser: `python manage.py createsuperuser`
2. From the Django Admin make sure you have at least one Org/RevenueProgram/PaymentProvider; note the Stripe Account ID
3. Get a token/cookie for that user: `curl -c cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -X POST -d '{"email":"foo@bar.com", "password": "foobar"}' http://localhost:8000/api/v1/token/`
4. Retrieve Revenue Programs with that Stripe Account ID: `curl -b cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -X GET  "http://localhost:8000/api/v1/public/revenue-programs/?stripe_account_id=acct_1234"|jq .`
5. Confirm that what's returned matches what you see in the Django Admin.

No.

Yes.

N/A

No.

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-592

No.

No.


apps/public/views.py
```diff
@@ -9,7 +9,13 @@
 class RevenueProgramViewset(ReadOnlyModelViewSet):
     model = RevenueProgram
     permission_classes = [IsAuthenticated, IsActiveSuperUser]
-    queryset = RevenueProgram.objects.all()
+
+    def get_queryset(self):
+        queryset = RevenueProgram.objects.all()
```
leo-dcfa, line 14: it's a small performance gain but instead of retrieving the complete record set and then filtering it, you could check if there query param is provided and if yes, you can retrieve by stripe_account id otherwise retrieve all

leo-dcfa (approved):
a small suggestion, not a blocked

leo-dcfa:
Tested and working as expected

PR #670, opened 2022-09-15
DEV-1106 | SPA-side changes for Stripe Account Link user flow

benjaminEwhite:

This is the second of 3 planned stacked PRs to implement the requires in [DEV-1106](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-1106).

The base for this PR is DEV-1106-main, which has already had server-side changes merged in. 

After this PR is merged, there should be a PR of DEV-1106-main into main.

**NOTE**: The diff on this PR of 46 files changed is somewhat misleading. I renamed some fixtures, and that required updating all the places they were imported, so that accounts for a substantial portion of the changed files. There were also some unused fixtures that were deleted.

**ALSO NOTE**: In testing this flow out in review app, I noticed that Stripe allowed me to submit the form and I was seemingly done, but then I got back to the SPA and the Connect Account CTA was still showing. This was not a UI bug, but instead because the Stripe Account was not fully configured to accept charges. I had to click the Connect Account CTA button again and Stripe asked for additional information that it not asked for first time around. This same flow was required one or two more times. If this is indicative, we may need to design/implement an "in-between" state for account creation flow where the dashboard can understand that a user has started connection process, but that it's not totally complete, and we alter the text in the CTAs accordingly. See [DEV-2401](https://news-revenue-hub.atlassian.net/browse/DEV-2401)

#### What's this PR do?

This PR is primarily SPA-side changes to support the Stripe Account Link flow. In working on front end, a small amount of iteration was required on backend code.

In the status quo, we used an [OAuth flow](https://stripe.com/docs/connect/oauth-reference) to connect RP Stripe accounts to the Hub Stripe account so we can process payments on behalf of orgs. 

With this PR, we start using the [Stripe Account Link flow](https://stripe.com/docs/connect/standard-accounts). When the SPA sees that an org-admin user has logged in and the `stripe_verified` value of their RP's payment provider is false, it will display a CTA (via a modal or if user has deferred, a collapsible toast component) prompting the user to connect their account. When the user clicks on this, they are taken off site to Stripe, where they fill out a (long) form providing details about their business. 

When the user completes the Stripe configuration, Stripe sends them back to the SPA, which signals to the backend that the user has completed Stripe setup. The backend retrieves the relevant Stripe account, and if it sees that "charges_enabled" is True, it updates the payment_provider.stripe_verified value to be true.  

The SPA will pick up on this changed state because we trigger a refetch of user data after the API request to signal stripe completion returns.

The changes in this PR lean on [react-query ](https://tanstack.com/query/v4/?from=reactQueryV3&original=https://react-query-v3.tanstack.com/), which provided a streamlined, nifty way to deal with retrieving and mutating state that is on the server (especially around user data). This required some light refactoring of other parts of the SPA to work with two custom hooks (a refactored `useUser` and a new `useFeatureFlags`). 

Some specific changes to note:

**Server-side changes** (@njharman )

- Adds a `payment_provider_stripe_verified` property to the `RevenueProgram` model.
- Updates the RevenueProgramInlineSerializer to include `payment_provider_stripe_verified` as a property. This serializer is used inline when serializing a user, so including this value means that we can the Stripe verification status of an RP by just looking at the serialized user.
- Updates `apps.users.tests.test_serializers.UserSerializerTest.test_has_expected_fields` to ensure that inline RPs in serialized user have expected keys (inclusive of `payment_provider_stripe_verified`).
- Adds `get_stripe_account_link_return_url` to organizations/views.py. This function is used to generate the return and refresh url values passed to stripe.AccountLink.create. This function was needed to make the account link flow work when running locally.

**SPA-side changes** 

- Renames `hub-admin.json` => `login-success-hub-admin.json` and same pattern for `org-admin.json`, `rp-admin-unverified.json`. The data in these fixtures is what is returned after successful login, and contains user data in addition to auth data. The updated file names make clear what the fixture represents (and hopefully saves other devs from confusion I encountered when trying to use one of these fixtures when intercepting the /users endpoint).
- Deletes some unused fixtures.
- Adds some new fixtures to reflect expected states of serialized self-service users vis-a-vis Stripe Account Link flow (`self-service-user-not-stripe-verified.json` and `self-service-user-stripe-verified.json`).
- Adds tests to the Cypress account spec for happy path a.) through the initial on-site part of account link flow (seeing and clicking on the CTA to connect account), b.) through the back-on-site-after-filling-out-off-site-Stripe-form part of the flow that results in the CTA no longer being displayed. (FWIW, I was on the fence about whether these tests should live in account or dashboard spec. The flows happen in the dashboard, but they have to do with new account flow, so put in account spec).
- Refactors `useUser` hook to use react-query. 
- Adds a new hook `useFeatureFlags` that has `useUser` as a dependency.
- With the refactored `useUser` and the new `useFeatureFlags`, I was able to simplify state management around user and feature flags data. 
- Hooks up pre-existing ConnectStripeElements, ConnectStripeModal, ConnectStripeToast, and Dashboard components to the new API endpoints from first PR for this ticket to fully implement the flow.

Implements Stripe Account Link flow which will facilitate a good onboarding experience for self-service users.

1. Go to https://dev-1106.revengine-review.org/create-account and fill out form to create a new account. Use your NRH email address with a `+something` suffix so the account verification email will come to your inbox.
2. Go to your inbox. Open the email verification email and click on the link.
3. Go to [the Django admin](https://dev-1106.revengine-review.org/nrhadmin/). Create a new org
4.  Next create a new RP that is for that org. When creating the RP, click on the inline payment provider plus button to create a new payment provider. Don't add a stripe id or change any of the default settings. Just click save.
5. Save the new RP.
6. Create a new role assignment for the user you created. The role_type should be org_admin, and give access to the org you just created.
7. Now go to https://dev-1106.revengine-review.org/pages (as the user you just created). You should see the Stripe connect modal.  Click on "Connect to Stripe" and in a moment you'll be taken to Stripe's site where you'll be prompted to create an account and then go through their onboarding flow. (alternatively, you can close the modal. this will cause the Stripe toast CTA to appear in the upper right hand corner of screen, and you can click on button inside of it to get to Stripe onboarding flow).
8. Once you've completed the Stripe flow, click the submit button, and you should be taken back to the dashboard but the Stripe CTA should no longer display.
9. Go to the Django admin and find the payment provider. It should now have a value for `stripe_acccount_id` and `stripe_verified` should be true.
10. You should also be able to now publish pages in page editor

**BUT! ** see note at top of PR about Stripe making it seem like you're done with account link, but in fact there are more steps. 

Yes. Cypress tests added for happy path flows. 

Updated a Python unit test to guarantee presence of `payment_provider_stripe_verified` inline rev programs in user serializer.

TBD

[The code makes some assumptions about user role types and the number of revenue programs they'll have](https://github.com/newsrevenuehub/rev-engine/pull/670/files#diff-70e45ec7295b73903f78553a7cc886383a4c154c77469dd45ecb00f0f7bf3459R78), with self-service users taken as the norm.  There is a long code comment explaining these assumptions.

The upshot is that once we need to think about handling per/RP Stripe verification for users that have access to > 1 RP, we'll need to refactor this code. I didn't want to pay that toll now since we don't know the exact business requirements around this.

No.

- [DEV-1106](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-1106)
- [DEV-2399 | Insert Stripe FAQ Link](https://news-revenue-hub.atlassian.net/browse/DEV-2399)
- [DEV-2395 | Display something in Dashboard UI if server can't create Stripe Account link](https://news-revenue-hub.atlassian.net/browse/DEV-2395)
- [DEV-2401 | Handle partially complete Stripe Account Link states](https://news-revenue-hub.atlassian.net/browse/DEV-2401)

Nay.

Yes.

- [DEV-2399 | Insert Stripe FAQ Link](https://news-revenue-hub.atlassian.net/browse/DEV-2399)
- [DEV-2395 | Display something in Dashboard UI if server can't create Stripe Account link](https://news-revenue-hub.atlassian.net/browse/DEV-2395)
- [DEV-2401 | Handle partially complete Stripe Account Link states](https://news-revenue-hub.atlassian.net/browse/DEV-2401)


spa/src/components/DashboardRouter.js
```diff
@@ -49,6 +47,7 @@ function DashboardRouter() {
 
             <Redirect from="/verified/:slug" to="/verify-email-success?result=:slug" />
             <Redirect from={ROUTES.VERIFIED} to={ROUTES.VERIFY_EMAIL_SUCCESS} />
+            <Redirect from={ROUTES.STRIPE_ACCOUNT_LINK_RETURN} to={`${ROUTES.CONTENT_SLUG}?stripeAccountLinkSuccess`} />
```
benjaminEwhite, line 50: FWIW, I initially tried doing this without adding a new route, thinking that it would be enough to send users back to dashboard with a query param. However, for reasons not entirely clear to me, the query param did not make its way all the way to the dashboard by the time it loads, and I saw this redirect pattern being used, tried it, and it solved my problem.

benjaminEwhite:
Closing and reopening to trigger new review app build. I've torn down review app from local. This is needed because I didn't use a subtask in JIRA to lead to separate ticket numbers for first PR and this one.

spa/cypress/integration/02 - donation-page-edit.spec.js
```diff
@@ -484,6 +440,77 @@ describe('Donation page edit', () => {
   });
 });
 
+describe('Edit interface: Setup', () => {
```
benjaminEwhite, line 443: Changes in this PR caused these tests to break. Fixing them was initially difficult because there were two (nested) `.before` blocks, which can create some gnarly path dependency between tests. Moved these tests out into own describe block to solve.

(Editorial comment: IMO nested before / beforeEach are to be avoided in Cypress because they lead to bugs and compromised tests where test order may impact outcome in unintended ways).

spa/src/components/dashboard/Dashboard.js
```diff
@@ -15,38 +16,125 @@ import Content from 'components/content/Content';
...
+  /*If user is an rp_admin whose first RP is not verified, we set the id
+  of that RP to trigger display of Stripe Connect CTA elements.
+
+  Note that the logic here is unsophisticated. It presumes that the primary
+  user (for now) is an rp_admin with a single revenue program. It will not
+  cause the Stripe Connect CTA elements to be displayed to users with other role types
+  than rp_admin. It also does not support connecting > 1 revenue program beyond
+  the first to appear in user.revenue_programs.
+  */
+  useEffect(() => {
```
nrh-cklimas, line 78: I'd suggest moving these three effects into a separate hook, which is easier to test and also less scary to read :). It looks like these all depend on the global user state, so the hook could either call `useUser()` directly or take the user object as an argument.

spa/src/components/dashboard/Dashboard.js
```diff
@@ -15,38 +16,125 @@ import Content from 'components/content/Content';
...
+  to the dashboard app route with a `stripeAccountLinkSuccess` present. If this component
+  loads and that param is present, we signal to the server that Stripe Account Link is complete,
+  and the server will update the revenue program's payment provider's stripe_verified value
+  to be `true` if it sees that the account is ready to accept charges.
+  */
+  useEffect(() => {
+    const params = new URLSearchParams(search);
+    if (
+      revenueProgramIdForVerification &&
+      [...params.keys()].includes('stripeAccountLinkSuccess') &&
```
nrh-cklimas, line 119:
```suggestion
      'stripeAccountLinkSuccess' in params &&
```

spa/src/hooks/useFeatureFlags.js
```diff
@@ -0,0 +1,12 @@
+import useUser from './useUser';
+
+/* This hook provides any feature flags attached to the user. It
+  depends on `useUser` which is responsible for retrieving the user from
+  the API, and that is the source of flag data.
+*/
+function useFeatureFlags() {
```
nrh-cklimas, line 7: I'm not sure about this. If it's just a wrapper around `useUser()`, are we better off just refactoring everything feature flags-related to use the user object instead? I don't have much of a handle on how we use feature flags, so this could in fact be a good idea... e.g. if we want to one day set feature flags on both a user and their organization, this could be a useful indirection.

spa/src/hooks/useUser.js
```diff
@@ -1,7 +1,31 @@
...
+    // this means that user won't be re-fetched for two minutes. After 2 minutes
+    // React Query's default "window focus" behavior could trigger a refetch if
+    // user is in different window and comes back to this one. Also, this query
+    // could refetch before or after the two minutes if it is manually invalidated
+    // in the SPA.
+    staleTime: 120000,
+    // When user logs in, the authentication endpoint returns user data and it gets
+    // stored in localstorage. We should update the localstorage data each time user
+    // is refetched so it isn't stale.
+    onSuccess: (data) => localStorage.setItem(LS_USER, JSON.stringify(data))
```
nrh-cklimas, line 26: I don't know react-query well enough to say, do we need to add error handling here for `fetchUser` failing too?

spa/src/components/dashboard/connectStripe/ConnectStripeElements.js
```diff
@@ -10,20 +12,27 @@ import StripeLogo from 'assets/icons/stripeLogo.svg';
 import useModal from 'hooks/useModal';
 import Cookies from 'universal-cookie';
```
nrh-cklimas, line 13: Just noticed this--I know not in scope for this PR. We should someday refactor so we're using one cookie library. Sorry I missed that in the other PR... still getting up to speed, obviously.

spa/src/components/dashboard/connectStripe/ConnectStripeToast.js
```diff
@@ -1,16 +1,15 @@
...
+import * as S from './ConnectStripeToast.styled';
 
 // Assets
 import StripeLogo from 'assets/icons/stripeLogo.svg';
 import Triangle6Dots from 'assets/icons/triangle6Dots.svg';
 import RemoveIcon from '@material-ui/icons/Remove';
 import RETooltip from 'elements/RETooltip';
 
-const ConnectStripeToast = () => {
+const ConnectStripeToast = ({ revenueProgramId, createStripeAccountLinkMutation }) => {
```
nrh-cklimas, line 12: Is there a way we can avoid prop drilling `createStripeAccountMutation`? e.g. having to have it passed down from Dashboard -> ConnectStripeElements -> ConnectStripeToast (if I have that right). I don't know react-query well enough to have a concrete suggestion, so tell me if I'm off-base here. The pattern I'm used to is that each component interacts with the shared state independently from each other.

nrh-cklimas (changes requested):
The only thing I feel strongly about here is moving the Stripe-related logic out of `<Dashboard>` and into a separate hook--rest are possible ideas. I haven't had time to test it out in the review app and might add a comment or two once I do, but I wanted to give you as much time as I could to look these over.

nrh-cklimas:
Forgot to add, I really like the react-query usage here, and it's clear it's going to allow us to simplify things.

nrh-cklimas:
I walked through the test steps up until you set up Stripe, because I'm not sure how to do that in a test context, and things looked good there.

spa/src/components/DashboardRouter.js
```diff
@@ -60,11 +59,7 @@ function DashboardRouter() {
                 ROUTES.EDITOR_ROUTE,
                 ROUTES.VERIFY_EMAIL_SUCCESS
               ]}
-              render={() => (
-                <UserContextProvider>
```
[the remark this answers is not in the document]
benjaminEwhite: Hopefully you should be able to get what you need from useUser, but let's huddle on this.

apps/organizations/views.py
```diff
@@ -150,7 +156,7 @@ def create_stripe_account_link(request, rp_pk=None):
             {"detail": "Cannot create a stripe account link at this time. Try again later."},
             status=status.HTTP_500_INTERNAL_SERVER_ERROR,
         )
-    return Response(stripe_response, status=status.HTTP_202_ACCEPTED)
+    return Response(dict(stripe_response), status=status.HTTP_202_ACCEPTED)
```
njharman, line 159: Is stripe_response "dict-like"? And that's why make it a real dict? What issue was occurring [this is more education for me than "change request"]

apps/organizations/views.py
```diff
@@ -76,6 +76,14 @@ class RevenueProgramViewSet(viewsets.ReadOnlyModelViewSet):
     pagination_class = None
 
 
+def get_stripe_account_link_return_url(request):
+    return (
+        "http://localhost:3000/stripe-account-link-complete"
+        if settings.ENVIRONMENT == "local"
+        else request.build_absolute_uri(reverse("index") + "stripe-account-link-complete")
```
njharman, line 83: Why not create named url [in revengine/urls.py similar to verified urls], that can reverse directly?

apps/organizations/views.py
```diff
@@ -137,11 +145,9 @@ def create_stripe_account_link(request, rp_pk=None):
...
-            # NB: this may get updated when second PR focused on SPA-side changes is done
-            refresh_url=request.build_absolute_uri(reverse("index")),
+            refresh_url=get_stripe_account_link_return_url(request),
             # The URL that the user will be redirected to upon leaving or completing the linked flow.
-            # NB: this may get updated when second PR focused on SPA-side changes is done
-            return_url=request.build_absolute_uri(reverse("index")),
+            return_url=get_stripe_account_link_return_url(request),
             type="account_onboarding",
         )
     except StripeError:
```
njharman, line 153: This file is so close to 100% coverage 😮. Can't you feel it? Doesn't it taunt/tempt you? Covering the if/and error branch of code block you edited here would do it!

apps/users/tests/test_serializers.py
```diff
@@ -67,8 +67,12 @@ def test_has_expected_fields(self):
             "revenue_programs",
             "role_type",
         }
+        expect_rp_fields = {"id", "name", "slug", "organization", "payment_provider_stripe_verified"}
         data = self._get_serialized_data_for_user(self.org_admin_user)
         self.assertEqual(expected_fields, set(data.keys()))
+        self.assertGreaterEqual(len(data["revenue_programs"]), 1)
```
njharman, line 73: Please use pytest assert style tests `assert len(data["revenue_programs"]) >= 1`. In this test method, that you touched. No need to update rest of file, at this moment.

apps/organizations/tests/test_org_views.py
```diff
@@ -350,6 +350,9 @@ def test_happy_path(self, monkeypatch, rp_role_assignment):
         account_id = "someID"
         mock_account_create = mock.MagicMock(return_value={"id": account_id})
         monkeypatch.setattr("stripe.Account.create", mock_account_create)
+        mock_return_url = "https://foo.bar"
+        mock_get_return_url = mock.MagicMock(return_value=mock_return_url)
```
njharman, line 354: I'm not sure I understand need/advantage of mocking get_...link_return_url?  

Difference in URLs between local and "test" [aka CI] environments? Should the test environment not also return the local url?

njharman (changes requested):
Reviewing backend/django only.

need to merge main, I believe.

spa/cypress/integration/02 - donation-page-edit.spec.js
nrh-cklimas: Stray thought: the issue with the nested before/beforeEach may be that we're missing some matching afters to clean up the behaviors so they don't leak into other tests. But a style point.

apps/organizations/views.py
benjaminEwhite: Yeah, it's dict like and doesn't serialize unless explicitly cast to a dict.

spa/src/components/dashboard/Dashboard.js
benjaminEwhite: Yeah,  I actually tried `in` here and was surprised by the results. 

Check this out:

```
const foo = new URLSearchParms("?foo=bar");
console.log(foo.keys());
=> Iterator {}
"foo" in foo.keys()
=> false
"foo" in [...foo.keys()];  // to cast iterator to array
=> false

// BUT ¯\_(ツ)_/¯ this works...

[...foo.keys()].includes("foo")

```

spa/src/hooks/useFeatureFlags.js
benjaminEwhite: I think it's easy to imagine using feature flags at org or rp-level in the future.


I guess the other thing I like about this approach is that it hides the implementation details from calling context. So if I care about feature flags, I don't have to think about user directly, I just think about using a hook that deals with feature flags.

spa/src/hooks/useUser.js
benjaminEwhite: With react-query, we get two chances to deal with error state. One is internal to the query/mutation, by providing an `onError` function. The other is by using the `isError` boolean returned by calling by `useQuery`, which would allow calling context to decide what to do in case of error. 

I think in this case, we'd want to deal with error in parent context. That said, at a minimum, I think we'd always want to do a `console.error(err)` so that Sentry knows about the issue.

With that in mind, I'm going to add console.error behavior to onError for `useUser` and `useCreateStripeAccountLink`.  

`useUser` gets called in a number of places, but crucially, it gets called at the top of the tree in src/components/Main.js.  If useUser is still loading (based in `.isLoading` attribute), we don't display the dashboard.  I'm adding new behavior so that if `isError` is true (from useUser), then we'll display an error fallback.

Will take a similar approach to `isError` in other places I'm using these hooks.

spa/src/components/dashboard/Dashboard.js
nrh-cklimas: Ah, my bad. I was thinking of plain old objects. How about `params.has('stripeAccountLinkSuccess')`?

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has

spa/src/hooks/useFeatureFlags.js
nrh-cklimas: Sounds good for now.

spa/src/hooks/useUser.js
nrh-cklimas: Think we talked through this one on Slack today.

spa/src/components/dashboard/Dashboard.js
benjaminEwhite: 👍 - updated. see 3562573f2d28

apps/users/tests/test_serializers.py
benjaminEwhite: Updated. See 1657c2cfc96

apps/organizations/tests/test_org_views.py
benjaminEwhite: For the test environment, it's a non-issue. We only need this to return :3000 for local so that the full flow for stripe account connection can be tested locally without having to manually change :8000 to :3000.

nrh-cklimas (approved):

apps/organizations/views.py
benjaminEwhite: Done. See 13e2286fdb

apps/organizations/views.py
```diff
@@ -76,6 +76,14 @@ class RevenueProgramViewSet(viewsets.ReadOnlyModelViewSet):
     pagination_class = None
 
 
+def get_stripe_account_link_return_url(request):
+    return (
```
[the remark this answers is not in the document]
benjaminEwhite: Now spans one line. See 13e2286fdb

apps/organizations/views.py
```diff
@@ -76,6 +76,14 @@ class RevenueProgramViewSet(viewsets.ReadOnlyModelViewSet):
     pagination_class = None
 
 
+def get_stripe_account_link_return_url(request):
```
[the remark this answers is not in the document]
benjaminEwhite: See https://github.com/newsrevenuehub/rev-engine/pull/670/files#diff-04fbcde936dbafc7cad39fed75b333113887a1ae8009c4c0e7c729f86f9d4ea1R439-R452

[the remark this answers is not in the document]
benjaminEwhite: To be clear: I removed the mocking of this function in the test you saw, and then created unit tests for this function.

apps/organizations/views.py
benjaminEwhite: Done. See 53a300d7740

benjaminEwhite: Was able to get to 98%, but didn't add tests for FilterQuerySetByUserMixin.filter_queryset_for_user which would be necessary to get to 100%.

benjaminEwhite: See 283f690cf8deadc0201cee38259ed243c71d2500

njharman (approved):

njharman (approved):
lgtm The somewhat rude and not helpful "Codacy Static Code Analysis" check failed with 

>This pull request quality could be better.

benjaminEwhite:
@njharman fwiw, I'm pushing a couple of small changes up to solve for some of the complaints of Codacy, but some of its warnings are false positives (for instance, [this one](https://app.codacy.com/gh/newsrevenuehub/rev-engine/file/76231301985/issues/source?bid=31147861&fileBranchId=31147904#l49)).  See 4e386b40 .

.envrc.example
```diff
@@ -9,3 +9,4 @@ export PGPORT=54000
 export PGUSER=postgres
 export PGDATABASE=revengine
 export REDIS_URL=redis://localhost:6379
+export ENVIRONMENT=local
```
x110dc, line 12: I'd prefer that this not be "hard-coded" -- it should instead be a dev's name so something like "export ENVIRONMENT=<dev's first name here>" -- otherwise won't be able to tell where errors are coming from in Sentry if a dev enables Sentry.

apps/organizations/views.py
```diff
@@ -76,6 +76,17 @@ class RevenueProgramViewSet(viewsets.ReadOnlyModelViewSet):
...
+    """This function exists purely to create a distinct return url when running locally to enable the full...
+
+    flow spanning SPA and backend and off-site Stripe form completion to be traversed.
+    """
+    reversed = reverse("spa_stripe_account_link_complete")
+    return (
+        f"http://localhost:3000{reversed}" if settings.ENVIRONMENT == "local" else request.build_absolute_uri(reversed)
```
x110dc, line 86: Now I know why you set the `ENVIRONMENT` specifically in `.envrc.sample`. Per 12-factor I prefer not to see code that relies on "environments". Instead I'd prefer a pattern of creating a new variable for this very purpose if this has to be different locally. (With a default that does the right thing in production if undefined). See https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2047311873/Python+Style+Guide#Environment-variables

x110dc (changes requested):
Just a couple minor change requests to using environment variable. Ping me if you'd like to discuss.

.envrc.example
benjaminEwhite: 38b192eb2

apps/organizations/views.py
benjaminEwhite: SInce the value in non-local case needs to be `request.build_absolute_url(reversed)`, whose value can only be known at run time, what pattern would you suggest? Would it be something like `STRIPE_ACCOUNT_LINK_RETURN_BASE_URL` in `base.py` (defaulting to None if not set) and check here in `get_stripe_account_link_return_url` if that var is set, and if so return a URL using that var, otherwise use `request.build_absolute_url`?

x110dc: Yes, that's exactly what I'm saying. Thanks for stating it better!

benjaminEwhite: See 9f36a39a5

revengine/settings/base.py
```diff
@@ -520,3 +520,10 @@
     "ENVIRONMENT": ENVIRONMENT,
     "DASHBOARD_SUBDOMAINS": DASHBOARD_SUBDOMAINS,
 }
+
+# This is meant to facilitate testing of the Stripe Account Link flow in local dev environment
+# By setting this to "http://localhost:3000" in env, when you go through Stripe Account Link flow,
+# on completing the Stripe form, you'll be sent back to localhost:3000 (aka the SPA being served by
+# webpack) instead of getting sent to localhost:8000, which is what would happen by default in local
+# dev environment.
+STRIPE_ACCOUNT_LINK_RETURN_BASE_URL = os.getenv("STRIPE_ACCOUNT_LINK_RETURN_BASE_URL", None)
```
x110dc, line 529: This should be added to `.envrc.sample`, no?

x110dc (changes requested):
One small change request.

revengine/settings/base.py
benjaminEwhite: 4272eeb99

x110dc (approved):

PR #672, opened 2022-09-15
DEV-2341 User accepted_terms_of_service is read only

njharman:
#### What's this PR do?
Prevent User.accepted_terms_of_service date from being changed via API. Achieved through making field read_only in the Serializer.  It is not made read-only at DB layer. This is normal/standard otherwise it becomes difficult to manage your data through admin/other means. 

It seems that something in Serializer is ignoring META.read_only_fields for fields modified in get_fields(). Would maybe change to use Serializer(partial_updates=True), except fields we are making not required in get_fields() include "password" which is tricky special case.

Add test that past, future and same date for accepted_terms_of_service returns 200 and makes no changes.

Added comment to hopefully prevent future dev from frustration of debugging this behavior.

Legal protection. Business needs.

I doubt there is form that includes this read-only field. Will have to hit API directly via Postman or the like.

Create user with validated email and accepted terms (note timestamp of acceptance)

PATCH /v1/users/<user.id>/  <payload: {"accepted_terms_of_service": timestamp}

Verify that 200 is returned and that user.accepted_terms_of_service timestamp may not be changed.  [non-200 will be returned if not valid timestamp format]

No

Yes.

N/A

There is weirdness around META.read_only_fields being ignored, I believe because for fields messed with in get_fields().  Added comments to warn others. Don't believe it is worth ROI chasing down root cuase.
 
N/A

[DEV-2341]

No.

None.

[DEV-2341]: https://news-revenue-hub.atlassian.net/browse/DEV-2341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/users/serializers.py
```diff
@@ -117,25 +117,29 @@ def get_fields(self, *args, **kwargs):
         request = self.context.get("request", None)
         if request and getattr(request, "method", None) == "PATCH":
             fields["accepted_terms_of_service"].required = False
+            fields[
+                "accepted_terms_of_service"
+            ].read_only = True  # Read only does not work when set in META.read_only_fields.
```
benjaminEwhite, line 122: Think this comment could be slightly more verbose. It "doesn't work" because we can't set `accepted_terms_of_service` to be a read only field in general, because for `post` we need it to be writable.  The comment as is makes it seem like there's something strange going on with the serializer, whereas it seems that this complexity just stems from our business requirements.

benjaminEwhite (changes requested):
Code looks fine, but think the the code comment in the serializer could be a bit more verbose so it's crystal clear why we're setting read_only based on method.

apps/users/serializers.py
njharman: The POST got me thinking... I added test to ensure accepted...service is writeable in POST and rewrote comments to match our intended logic.  R/W Post, R/O Patch. Also added comment to view, that if we ever add PUT need to address the custom logic in serializer.get_fields [forgetting that is the sort of thing that bites future Norm in the ...]  

https://github.com/newsrevenuehub/rev-engine/pull/672/commits/fd61caae399c7e662a1020f0f07a44a0c84986db

benjaminEwhite (approved):

x110dc (approved):

x110dc (approved):

PR #677, opened 2022-09-19
DEV-2368: Default currency symbol to $ if missing in API data

nrh-cklimas:

n/a

#### What's this PR do?

Adds logic to the page editor such that if currency is not defined for the page, it defaults display to use a `$`.

Current state is to display `undefined` which looks broken. We decided to use $ as a default since most if not all of our users right now will eventually use it.

This requires some data massaging so the setup steps are a little weird. I have a revenue program called `nocurrencyrp` and an organization called `nocurrencyorg` already configured, which can be used instead of doing steps 1-4.

1. Log into the Django admin at https://dev-2368.revengine-review.org/nrhadmin/.
2. Create a test organization. Settings don't matter.
3. Create a test revenue program linked to the test organization you created in step 2. For the payment provider field, click the plus sign next to the field and create a dummy one; no settings are needed on it besides the default.
4. After creating the RP, edit it again and use the X button next to the payment provider to sever the link. Django will complain that the field is required, but it will still make the data change.
5. Create a test user in the Django admin and give it the role org admin for the org created in step 2.
6. Log into https://dev-2368.revengine-review.org with this user.
7. Create a new page.
8. Verify that in the preview, the monthly donation choices show a $ in front of them (as opposed to the word `undefined`).
  a. If you want to confirm that the data setup is correct, open dev tools and look at the request for `https://dev-2368.revengine-review.org/api/v1/pages/PAGE_NUMBER_HERE/`. In the response, it should show that the `currency` property is `null`. If you see actual values on `currency`, then your data is not set up to reproduce the bug this PR fixes.

No.

Added Cypress tests.

Not needed I think.

No.

Not needed.

https://news-revenue-hub.atlassian.net/browse/DEV-2368

No.

No.


GuiMend (approved):
LGTM 👍 

image

x110dc (approved):

PR #681, opened 2022-09-20
DEV-2298: Only fire "New page" modal when user has access to multiple revenue programs

GuiMend:

N/A

#### What's this PR do?
When a user creates a new page we show a modal for them to add a page name and select Revenue program. If a user only has 1 revenue program, this PR now bypasses the modal and gives the page name the format "Page [number]", which number is the amount of pages already existent + 1.

Reduces the friction for a user to create a page and have to come up with a name for it, not knowing what that is for at this point.

- Login with a user that has only 1 revenue program
- Go to `/pages` and create a new page. 
- Should go directly to the new page edit mode and the page name should be in the format "Page #number" or show an alert error if something went wrong (you can test this by removing network access just before clicking "new page" button)
- Login with user with multiple revenue programs
- "New page" button should open the existing modal and work as expected.

No

No, the current modal is only temporary. There is a follow up ticket to update the create page modal and I believe it would be best not to use time creating tests for this modal if it's going to be removed really soon. (Follow up ticket: https://news-revenue-hub.atlassian.net/browse/DEV-2299)

Create Donation Page is now simplified

Yes, the current AddPageModal implementation could be better. This will be addressed in the follow up ticket: https://news-revenue-hub.atlassian.net/browse/DEV-2299 when the modal is updated

No

[DEV-2298](https://news-revenue-hub.atlassian.net/browse/DEV-2298) Only fire "New page" modal when user has access to multiple revenue programs

No

Follow up ticket: https://news-revenue-hub.atlassian.net/browse/DEV-2299


spa/src/components/content/pages/AddPageModal.js
```diff
@@ -84,40 +88,61 @@ function AddPageModal({ isOpen, closeModal }) {
...
+      );
+    },
+    [canSavePage, createPage, handleSaveFailure, history, name, revenueProgram.id, slug]
+  );
 
   const handleDiscard = () => {
     closeModal();
   };
 
+  useEffect(() => {
```
nrh-cklimas, line 129: Something to consider is putting this logic in `<Pages>` instead, e.g. never fire the modal in the first place--it feels a little more elegant to never render the modal in the first place instead of having to force it to render null for a render cycle. But your call here.

If we do keep it here, we should catch any errors during handleSave() and show them as an alert--the existing error handling tries to display an error in a nonexistent dialog.

nrh-cklimas (changes requested):
One tweak here related to error handling.

spa/src/components/content/pages/AddPageModal.js
GuiMend: When working on the follow up ticket I'll rewrite this modal and do as you sugested. But for now I believe it's good enough. The error handling will show an alert for the user because the `if` should never be true in the way we are handling this part.

nrh-cklimas (approved):

x110dc (approved):

PR #687, opened 2022-09-20
DEV 2437 add stripe account for update payment method

x110dc:
#### What's this PR do?

Returns the Stripe Account ID on the contributions and subscriptions endpoints. 

The frontend needs the Stripe Account ID in order to create a new payment method to attach/update on a Subscription. It would've been difficult to get that otherwise. 

1. Run the server locally or use the Review App. 
2. Request a magic link for a RP/email that has Subscriptions
3. View the contributor portal with that magic link
4. Using DevTools examine the API response from the `/contributions/` and `/subscriptions` endpoints. 
5. Confirm both have a `stripe_account_id` field that's populated. 

No.

Yes.

No.

Since the Stripe objects themselves don't have a field indicating the Stripe Account they came from (when they come from a Connect webhook they do have this field) I opted to just add the field as an attribute before storing to the cache. That felt slightly ugly. Not sure if there's a better way to do it. 

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-2437

No.

This should unblock https://news-revenue-hub.atlassian.net/browse/DEV-1569


apps/contributions/stripe_contributions_provider.py
```diff
@@ -243,6 +243,10 @@ def serialize(self, contributions):
     def upsert(self, contributions):
         """Serialized and Upserts contributions data to cache."""
         data = self.serialize(contributions)
+        # add the stripe account ID
+        for k in data:
+            data[k]["stripe_account_id"] = self.stripe_account_id
```
njharman, line 248: Are there a lot? I can't imagine it matters but iterating over values and doing `v["stripe_account_id"] = self.stripe_account_id` is more efficient than looking up value each time via key.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -261,11 +264,12 @@ def load(self):
 
 
 class SubscriptionsCacheProvider:
```
njharman, line 266: Probably out of scope. But the amount of duplicated code between these CacheProviders makes me twitch.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -281,6 +285,9 @@ def serialize(self, subscriptions):
     def upsert(self, subscriptions):
         """Serialized and Upserts subscriptions data to cache."""
         data = self.serialize(subscriptions)
+        # add the stripe account ID
```
njharman, line 288: Comment would be more useful and less "just repeating what code says" if it included some of this explanation
 
> Since the Stripe objects themselves don't have a field indicating the Stripe Account they came from (when they come from a Connect webhook they do have this field)

apps/contributions/tests/test_stripe_contributions_provider.py
```diff
@@ -375,7 +375,10 @@ def test_load(self):
         redis_mock = RedisMock()
         with patch.dict("apps.contributions.stripe_contributions_provider.caches", {"default": redis_mock}):
             cache_provider = ContributionsCacheProvider(
-                "test@email.com", serializer=self.serializer, converter=self.converter
+                email_id="test@email.com",
+                stripe_account_id="bogus",
```
njharman, line 379: Should test that strip_account_id is present and the correct value in data returned from cache, no?

apps/contributions/serializers.py
```diff
@@ -497,6 +497,7 @@ class PaymentProviderContributionSerializer(serializers.Serializer):
     credit_card_expiration_date = serializers.CharField(max_length=7)
     created = serializers.DateTimeField()
     last_payment_date = serializers.DateTimeField()
+    stripe_account_id = serializers.CharField(max_length=255, required=False, allow_blank=True)
```
njharman, line 500: [Mostly just verifying I understand. (The AC where not specific)]

It's find only adding to this serializer, (and not the Contribution/Subscription ones) because we only need the stripe_account_id for the contribution portal name?. Thing contributors use to see their contributions. And not the other use cases of endpoints (users of nrh viewing/managing contributions to their org/rp) 

Also I don't understand why(aka don't like that) both use cases are jammed into the same endpoints.

njharman (changes requested):
pre-review checklist not completed.
time not logged.

Full coverage, but some logic is not tested.

apps/contributions/serializers.py
x110dc: As far why we need it on these two endpoints and nowhere else: that's explained by @nrh-cklimas here: https://news-revenue-hub.atlassian.net/browse/DEV-1569?focusedCommentId=55830

I'm not sure I understand your second paragraph, but there's intentionally some overlap between the `/contributions/` and the `/subscriptions/` endpoint right now because the FE is relying on the former for too many things but reducing that overlap is addressed in these tickets: 

- https://news-revenue-hub.atlassian.net/browse/DEV-2320
- https://news-revenue-hub.atlassian.net/browse/DEV-2391
- https://news-revenue-hub.atlassian.net/browse/DEV-2438

While the FE transitions to use the latter API more. 

Also, I'm confused myself a bit about why the stripe_account_id was showing on the `/subscriptions/` endpoint when I hadn't explicitly declared it on the serializer, but I added that here: https://github.com/newsrevenuehub/rev-engine/pull/687/commits/4315d2e400ce2df6a84a3abffa7d84739428c9d4 I guess something about DRF that I don't fully understand.

apps/contributions/tests/test_stripe_contributions_provider.py
x110dc: Fixed here: https://github.com/newsrevenuehub/rev-engine/pull/687/commits/ba97fb05b2cc27820bbfe2866c98d400c8107b60

apps/contributions/stripe_contributions_provider.py
x110dc: Agreed. Fixed here: https://github.com/newsrevenuehub/rev-engine/pull/687/commits/ba97fb05b2cc27820bbfe2866c98d400c8107b60

apps/contributions/stripe_contributions_provider.py
x110dc: Agreed, out of scope for this ticket. But I added a TODO/ticket to address: https://github.com/newsrevenuehub/rev-engine/pull/687/commits/f0d62fc604821b6b1174797d93199474bc5f73b4

apps/contributions/stripe_contributions_provider.py
x110dc: 👍 Improved here: https://github.com/newsrevenuehub/rev-engine/pull/687/commits/ba97fb05b2cc27820bbfe2866c98d400c8107b60

njharman (approved):
no pre-review checklist

PR #671, opened 2022-09-15
DEV-2028: Add profile finalization modal

nrh-cklimas:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

n/a, ready to review

#### What's this PR do?

This is https://github.com/newsrevenuehub/rev-engine/pull/660 but with added tests and some code cleanup.

Adds a modal that asks the user to finalize their profile with their name, job title, company name, and type. This is available by going directly to `/setup` but also appears if the user has no organizations (see utilities/needsProfileFinalization.js for the criteria).

I added axios-mock-adapter as a dev dependency so we can mock API responses in Jest.

#### Why are we doing this? How does it help us?

Improves our user onboarding flow.

#### How should this be manually tested? Please include detailed step-by-step instructions.

1. Go to https://dev-2028.revengine-review.org/create-account.
2. Fill out the form, use a +test or similar in your email address.
3. Click the link to verify your email address.
4. Once you do this, you should immediately see a modal to customize your profile.
  a. The form should only be submittable once (e.g. the submit button should be disabled) you fill in name, company name, and type. Job title is optional.
5. After submitting the form, the modal should prompt you to connect to Stripe. Close this.
6. You should now be able to interact with the admin dashboard as usual and should not see the profile modal in step 4 again.
#### Did this PR make changes to the user interface that may require changes to the user-facing docs?

Yes.

#### Have automated unit tests been added? If not, why?

I added Jest tests on new code and added Cypress tests for a basic happy path on completing the customization modal as well as error handling if customizing the profile fails.

#### How should this change be communicated to end users?

May want to expand the KB article on this.

#### Are there any smells or added technical debt to note?

The way the select in the modal is styled is pretty bad and hacky. I plan to address this in followup work.

There is a backend change in the existing PR that may need test coverage.

#### Has this been documented? If so, where?

https://news-revenue-hub.atlassian.net/wiki/spaces/REVENGINE/pages/2222456857/How+do+I+create+an+account

#### What are the relevant tickets? Add a link to any relevant ones.

https://news-revenue-hub.atlassian.net/browse/DEV-2028

#### Do any changes need to be made before deployment to production (adding environment variables, for example)?

No.

#### Are there next steps? If so, what? Have tickets been opened for them? List next-step tickets here:

- https://news-revenue-hub.atlassian.net/browse/DEV-2368
- https://news-revenue-hub.atlassian.net/browse/DEV-2387
[...]

PR #688, opened 2022-09-20
Ad hoc PR: handle edge case with rp slug name

benjaminEwhite:

This is to solve for [this comment in parent PR](https://github.com/newsrevenuehub/rev-engine/pull/671#discussion_r975722697). I would have used the make a suggestion feature, but it required change in object code and test so that would be unwieldy in GH.

### DIFF
--- a/apps/users/views.py
+++ b/apps/users/views.py
@@ -14,6 +14,7 @@ from django.contrib.auth.views import (
 )
 from django.core import signing
 from django.core.exceptions import ValidationError as DjangoValidationError
+from django.db.utils import IntegrityError
 from django.dispatch import receiver
 from django.http import HttpResponseRedirect
 from django.urls import reverse, reverse_lazy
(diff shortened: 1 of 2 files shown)


apps/users/views.py
```diff
@@ -308,14 +309,27 @@ def customize_account(self, request, pk=None):
...
+        except IntegrityError:
+            logger.exception(
+                "[UserViewSet.customize_account] found unexpected already extant RP "
+                "with slug %s after creating organization (ID: %s) with slug %s",
+                organization_name,
+                organization.id,
+                organization.slug,
+            )
+            raise ValidationError(
+                "Something unexpected happened customizing the account",
```
njharman, line 330: Assuming this is displayed to user somehow, shouldn't it be more explicit.  That name already taken or something.

njharman (commented):
Meta I don't like this "ad hoc PR". I don't know what ticket this is, where to log time.

I thought our policy (I saw in some other slug related task I believe for donation page) was to check db for availability first and add serial integer to end of slug until unique name found rather than fail?)

In general +1 catching easy/expected integrityerrors. Looks like we have several things with slugs.  I'd look into encapsulating into function instead of repeating code everywhere.

PR #689, opened 2022-09-21
DEV-2408: Stripe webhooks fail for subscriptions

benjaminEwhite:

This fixes a bug discovered in UAT for DEV-2183. It should be merged into DEV-2183-root-branch, and that branch will eventually be merged into main.

#### What's this PR do?

This PR fixes a bug that was causing Stripe webhooks to fail for recurring contributions ("Subscriptions" in Stripe lingo).

The root cause of the bug was that we were not saving the Stripe Customer ID back on to the NRE contribution after creating a Stripe PaymentIntent or Subscription. In our webhooks code, for subscriptions, we were attempting to retrieve by customer ID instead of subscription ID.

To solve for this, this PR changes the contribution retrieval behavior of our webhook so that it searches contributions by Stripe subscription ID for subscriptions. It also updates the payment intent and subscription creation methods in apps.contributions.models.Contribution so that the customer id gets written back to the contribution (even though the customer id is no longer relevant for the webhook in question).

Fixes a bug that stops us from shipping DEV-2183.

Create a recurring contribution ([here](https://billypenn-dev-2408.revengine-review.org/donate)), then check in NRH admin ([here](https://dev-2408.revengine-review.org/nrhadmin/contributions/)) and its status should be "paid".

In addition to fixing bug, changes a .info to a .exception so we get sentry errors when webhooks don't match with contributions.

No

Tests have been updated

N/A

No

N/A

- [DEV-2408 | Webhooks are broken for Subscriptions in DEV-2183](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/boards/3?modal=detail&selectedIssue=DEV-2408)
- [DEV-2183](https://news-revenue-hub.atlassian.net/browse/DEV-2183)

No

No

### DIFF
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -289,6 +289,7 @@ class Contribution(IndexedTimeStampedModel, RoleAssignmentResourceModelMixin):
         )
         self.provider_payment_id = intent["id"]
         self.provider_client_secret_id = intent["client_secret"]
+        self.provider_customer_id = intent["customer"]
         self.save()
         return intent
 
@@ -320,6 +321,7 @@ class Contribution(IndexedTimeStampedModel, RoleAssignmentResourceModelMixin):
         )
         self.payment_provider_data = subscription
         self.provider_subscription_id = subscription["id"]
+        self.provider_customer_id = subscription["customer"]
         self.provider_client_secret_id = subscription["latest_invoice"]["payment_intent"]["client_secret"]
         self.save()
         return subscription
--- a/apps/contributions/webhooks.py
+++ b/apps/contributions/webhooks.py
@@ -11,6 +11,7 @@ logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
 
 class StripeWebhookProcessor:
     def __init__(self, event):
+        logger.info("StripeWebhookProcessor called with %s", event)
         self.event = event
         self.obj_data = self.event.data["object"]
 
--- a/apps/contributions/webhooks.py
+++ b/apps/contributions/webhooks.py
@@ -15,6 +15,7 @@ class StripeWebhookProcessor:
         self.obj_data = self.event.data["object"]
 
     def get_contribution_from_event(self):
+        logger.info(self.obj_data)
         if self.obj_data["object"] == "subscription":
             return Contribution.objects.get(provider_subscription_id=self.obj_data["id"])
         else:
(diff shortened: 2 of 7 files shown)


apps/contributions/models.py
```diff
@@ -320,6 +321,7 @@ def create_stripe_subscription(self, stripe_customer_id, metadata):
         )
         self.payment_provider_data = subscription
         self.provider_subscription_id = subscription["id"]
+        self.provider_customer_id = subscription["customer"]
```
njharman, line 324: It doesn't seem that test_stripe_subscription checks that this value is set.

njharman, line 324: It doesn't seem that test_stripe_subscription checks that this value is set.

apps/contributions/models.py
```diff
@@ -289,6 +289,7 @@ def create_stripe_one_time_payment_intent(self, stripe_customer_id, metadata):
         )
         self.provider_payment_id = intent["id"]
         self.provider_client_secret_id = intent["client_secret"]
+        self.provider_customer_id = intent["customer"]
```
njharman, line 292: It doesn't seem that test_create_stripe_one_time_payment_intent checks that this value is set. In fact deleting this line causes no failure in test_models.py (where I'd expect failure) and no failure in any test.

apps/contributions/webhooks.py
```diff
@@ -11,17 +11,22 @@
...
+            return Contribution.objects.get(provider_subscription_id=self.obj_data["id"])
+        elif event_type == "payment_intent":
+            try:
+                return Contribution.objects.get(provider_payment_id=self.obj_data["id"])
+            except Contribution.DoesNotExist as e:
+                if customer_id := self.obj_data.get("customer"):
+                    # This is fine as long as we continue to generate a unique customer per charge.
+                    return Contribution.objects.get(provider_customer_id=customer_id)
+                else:
+                    raise e
```
njharman, line 29: Pretty sure the "e" is superfluous and thus confusing (implying that code is modifying or raising some other exception than one being handled) https://docs.python.org/3/reference/simple_stmts.html#raise

njharman (changes requested):
no time logged
no pre-review check list

apps/contributions/webhooks.py
benjaminEwhite: b7405441

apps/contributions/models.py
benjaminEwhite: b7405441

apps/contributions/models.py
benjaminEwhite: b7405441

x110dc (approved):

PR #709, opened 2022-09-27
DEV-2234: Update Contributions layout

GuiMend:

N/A

#### What's this PR do?
- Create all components related to the updated Contributions page layout and their stories
- Update Contributions layout
- Update constants files so that we can do absolute imports through out the project

We are updating all pages and components to match new designs, making the app more consistent and user friendly.

- Login
- Go to Conrtibutions page
- Make sure the design matches figma https://www.figma.com/file/gpbDEAUkqCwHc1j0GTi98v/%5BDES-7%5D-Landing-Page-Experience-(Phase-1)?node-id=50%3A374
- Check Storybook for individual components `Filters > StatusFilter` and `Filters > AmountFilter`

Yes

Yes

Updates to Contribution page now makes that page consistent with the rest of the application.

No

No

https://news-revenue-hub.atlassian.net/browse/DEV-2234

No

https://news-revenue-hub.atlassian.net/browse/DEV-2235


spa/src/styles/themes.js
```diff
@@ -110,7 +111,8 @@ export const revEngineTheme = {
     sm: '2px',
     md: '4px',
     lg: '6px',
-    xl: '10px'
+    xl: '10px',
+    '2xl': '12px'
```
nrh-cklimas, line 115: Nitpick: how about `xxl` instead? That way we don't need to quote the property name.

spa/src/components/common/Filters/AmountFilter/AmountFilter.js
```diff
@@ -0,0 +1,68 @@
...
+import PropTypes from 'prop-types';
+
+import CurrencyField from 'components/common/TextField/CurrencyField/CurrencyField';
+
+import { Flex, Label, Content } from './AmountFilter.styled';
+import useDebounce from 'hooks/useDebounce';
+import usePreviousState from 'hooks/usePreviousState';
+
+const AmountFilter = ({ onChange, className }) => {
+  const [amount, setAmountRange] = useState({ amount__gte: '', amount__lte: '' });
```
nrh-cklimas, line 11: Couple things I wonder about:

- It feels weird for this state to have an initial value of an empty string, but once the user starts making changes, they are always numbers. How about we set them to `null` first?
- Why use double underscores instead of camel case for the state names, and why put them in one state object instead of two useState()s?
- Could you talk about the intent of debounce() here? The reason I ask is I think the intent is to prevent onChange() from being called too often. If that's the case, we don't need to debounce state, we can just debounce the call to onChange... if that makes sense.

spa/src/components/common/Filters/AmountFilter/AmountFilter.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+  it('should render AmountFilter', () => {
+    tree();
+
+    expect(screen.getByText(/amount/i)).toBeVisible();
+    expect(screen.getByRole('spinbutton', { name: 'Filter minimum amount' })).toHaveValue(null);
+    expect(screen.getByRole('spinbutton', { name: 'Filter maximum amount' })).toHaveValue(null);
+  });
+
+  it('should call onChange when amount changes', async () => {
```
nrh-cklimas, line 19: We should also test that onChange is debounced.

nrh-cklimas (changes requested):
Overall looks like a big improvement over what we have. Most of my concerns are related to the debouncing on the monetary filters.

spa/src/styles/themes.js
GuiMend: I can do that, but what if we need even higher values as we have in other places?
`xxxl, xxxxl, ...` doesn't seem as good as `2xl, 3xl, 4xl, ...`

spa/src/components/common/Filters/AmountFilter/AmountFilter.js
GuiMend: > It feels weird for this state to have an initial value of an empty string, but once the user starts making changes, they are always numbers. How about we set them to null first?
It can always be a string and I only convert to number at the onChange call. Can't leave it `null` because I'm changing uncontrolled/controlled component

> Why use double underscores instead of camel case for the state names, and why put them in one state object instead of two useState()s?
It's the way the filters was currently setup and I didn't feel like it was necessary to change the implementation, the `amount__gte` are the queryparams for the backend call
image


> Could you talk about the intent of debounce() here? The reason I ask is I think the intent is to prevent onChange() from being called too often. If that's the case, we don't need to debounce state, we can just debounce the call to onChange... if that makes sense.
👍

spa/src/components/common/TextField/CurrencyField/CurrencyField.js
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <FormControl fullWidth variant="outlined">
+      <OutlinedInput
+        inputProps={{ min: 0, 'aria-label': ariaLabel }}
+        className={classes.root}
+        type="number"
+        value={value}
+        onChange={onChange}
+        placeholder={placeholder}
+        startAdornment={<InputAdornment position="start">$</InputAdornment>}
```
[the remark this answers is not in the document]
GuiMend: Do you mean the currency in the `/contributions` endpoint? It looks like each result could have a different currency
image

Also, just noticed that the donation table that shows all results always use `UDS`
```
const formatCurrencyAmount = (rawAmount) => {
  return new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD' }).format(rawAmount / 100);
};
```

spa/src/styles/themes.js
nrh-cklimas: It's really hard to say as we don't know enough about what we need from a design standpoint. I'm good with having a discussion around it as part of the component library work as letting this stand as-is.

spa/src/components/common/TextField/CurrencyField/CurrencyField.js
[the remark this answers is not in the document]
nrh-cklimas: It feels like this is a bigger discussion around currency. For now, I guess let's go with the assumption that everything is $s in the dashboard. However, since we're calling this a common component, I think the currency symbol should be a prop so we can potentially re-use it elsewhere.

spa/src/components/common/Filters/StatusFilter/StatusFilter.test.js
```diff
@@ -0,0 +1,42 @@
...
+const onClick = jest.fn();
+
+describe('StatusFilter', () => {
+  const tree = (props) => render(<StatusFilter onClick={onClick} {...props} />);
+
+  it('should render default status filter', () => {
+    tree();
+
+    expect(screen.getByText(/status/i)).toBeVisible();
+    STATUS_FILTERS.forEach((status) => {
```
nrh-cklimas, line 16: Sorry, just noticed the same issue re: forEach vs `for (const status of STATUS_FILTERS)`

spa/src/components/common/Filters/AmountFilter/AmountFilter.js
GuiMend: I think I solved the issue with AmountFilter component. Now the `AmountFilter` only handles string numbers. There isn't a way the user can enter anything other than number in the currency input because it has `type=number`, so the conversion to Int/Float/Number will always work

GuiMend: image

nrh-cklimas: These improvements look great.

nrh-cklimas (approved):

x110dc (approved):

PR #747, opened 2022-10-10
DEV-2555: Profile Step Updates

GuiMend:

n/a

#### What's this PR do?
Fix design mismatch between implementation and figma for Profile Steps modal

To make design consistent accross all pages and modals

- Create new account
- Ask for verification email and verify account
- Go back to application (should be `/setup/` page)
- Check that modal design matches figma

Slightly. Don't think requires doc changes

n/a

n/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-2555

no

Parent ticket: https://news-revenue-hub.atlassian.net/browse/DEV-2107


spa/src/components/account/Profile/InfoTooltip.test.js
```diff
@@ -19,12 +19,12 @@ describe('InfoTooltip', () => {
     await waitFor(() => expect(screen.getByText('test tooltip')).toBeVisible());
   });
 
-  it('keeps the tooltip open if the button is clicked repeatedly', async () => {
+  it('toogle the tooltip if the button is clicked repeatedly', async () => {
```
nrh-cklimas, line 22:
```suggestion
  it('toggles the tooltip if the button is clicked repeatedly', async () => {
```

spa/src/components/base/Tooltip/Tooltip.js
```diff
@@ -23,12 +24,12 @@ export const Tooltip = styled(WrappedTooltip)`
 `;
 
 Tooltip.propTypes = {
-  ...MuiTooltip.propTypes,
```
nrh-cklimas, line 26: Why remove this? My thought here was to give consumers an escape hatch to use MUI directly for special cases.

spa/src/components/base/Tooltip/Tooltip.js
```diff
@@ -14,6 +14,7 @@ const WrappedTooltip = ({ className, tooltipWidth, ...other }) => (
 export const Tooltip = styled(WrappedTooltip)`
   && {
     background: rgb(50, 50, 50);
+    margin-top: 6px;
```
nrh-cklimas, line 17: I don't think this will play well with different tooltip placements. If what we want is to ensure that tooltip is 6px away from its anchor, I think we want something like `PopperProps={{ modifiers: [{ name: 'offset', options: { offset: [0, 6]  }]  }}` on `MuiTooltip` (I didn't test this, though).

Docs:
- https://v4.mui.com/api/popper/
- https://popper.js.org/docs/v2/modifiers/offset/

nrh-cklimas (changes requested):
Changes look good in the review app, just have two comments on base component changes.

spa/src/components/base/Tooltip/Tooltip.js
GuiMend: Btw: I tried using the `popperprops` but didn't get the expected results and didn't want to spend too much time figuring out the props if the margin worked in all cases

GuiMend: I like that Chris, I've added your suggestion and updates storybook to so that the icon tooltip also shows the correct placement

x110dc (approved):

PR #751, opened 2022-10-11
DEV-2596: Restore /:pageSlug/thank-you/ route

nrh-cklimas:

n/a

#### What's this PR do?

This fixes a bug where requests to `/:donationSlug/thank-you/` were not getting routed properly.

This also adds TypeScript support to the project 🌈 and adds unit tests for the donation page router.

Fixes a very confusing situation for contributors where donating from a page with a slug didn't show a confirmation page.

Repeat the steps below for both a page with a donation slug (like `/donate/`) and one using the default donation page (at the top-level domain, no subdirectories).

1. Go to the donation page.
2. Complete a test donation.
3. Confirm that after doing so, you see a thank you page. When you start at `/donate/`, the browser address bar should be `/donate/thank-you/`. When starting at the top-level domain, the browser address bar should be `/thank-you/`.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2596

No.

No.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,9 @@ COPY ./spa/package.json ./spa/package-lock.json /code/
 RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit --no-fund --silent
 COPY ./spa /code/spa/
 WORKDIR /code/spa/
-RUN NODE_ENV=production npm run build
+
+# Limit memory usage to avoid Heroku deploy issues.
+RUN npm run build:for-heroku
 
 FROM python:3.10-slim as base
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,9 @@ FROM node:16-slim as static_files
 WORKDIR /code
 ENV PATH /code/node_modules/.bin:$PATH
 COPY ./spa/package.json ./spa/package-lock.json /code/
-RUN npm install --silent
+
+# Skip installing Cypress binaries and security/funding audits, use package-lock.json strictly
+RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit --no-fund --silent
 COPY ./spa /code/spa/
 WORKDIR /code/spa/
 RUN NODE_ENV=production npm run build
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ COPY ./spa /code/spa/
 WORKDIR /code/spa/
 
 # Limit memory usage to avoid Heroku deploy issues.
-RUN npm run build:for-heroku
+RUN npm run build:low-memory
 
 FROM python:3.10-slim as base
 
--- a/spa/package.json
+++ b/spa/package.json
@@ -83,7 +83,7 @@
       "**/?(*.)+(test).[jt]s?(x)"
     ],
     "collectCoverageFrom": [
-      "**/*.{js,jsx}",
+      "**/*.[jt]s?(x)",
       "!**/node_modules/**",
       "!**/*.stories.js",
       "!**/*.styled.js"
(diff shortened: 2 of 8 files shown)


spa/src/components/DonationPageRouter.tsx
```diff
@@ -37,7 +37,7 @@ function DonationPageRouter() {
           <React.Suspense fallback={<GlobalLoading />}>
             <Switch>
               <Route
-                path={[ROUTES.DONATION_PAGE_SLUG + ROUTES.THANK_YOU_SLUG, ROUTES.THANK_YOU_SLUG]}
+                path={[join(ROUTES.DONATION_PAGE_SLUG, ROUTES.THANK_YOU_SLUG), ROUTES.THANK_YOU_SLUG]}
```
nrh-cklimas, line 40: This is the fix for the bug--everything else is tests and TypeScript setup.

spa/src/components/DonationPageRouter.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    history
+  };
+}
+
+describe('DonationPageRouter', () => {
+  const GenericThankYouMock = GenericThankYou as jest.Mock;
+
+  beforeEach(() => GenericThankYouMock.mockReturnValue(<div data-testid="mock-generic-thank-you" />));
+
+  it('displays a tracked GenericThankYou at /thank-you/', async () => {
```
nrh-cklimas, line 59: I'm hardcoding these paths in the tests to avoid a situation where slashes change in our route config, and we accidentally route to URLs like `/donate//thank-you` but tests continue to pass.

GuiMend (approved):
Really nice 👍

spa/package.json
```diff
@@ -69,6 +69,7 @@
     "start": "craco start",
     "start:subdomains": "DANGEROUSLY_DISABLE_HOST_CHECK=true craco start",
     "build": "INLINE_RUNTIME_CHUNK=false craco build",
+    "build:for-heroku": "INLINE_RUNTIME_CHUNK=false NODE_ENV=production craco --max_old_space_size=2560 --optimize_for_size build",
```
nrh-cklimas, line 72: This came from a combo of https://devcenter.heroku.com/articles/node-memory-use#heroku-memory-limits (describing what CLI switches to use) and pages like https://stackoverflow.com/questions/59205530/heroku-server-crashes-with-javascript-heap-out-of-memory-when-deploying-react which were suggesting 2560 as a limit.

This does potentially cause the build to be slower than it could be because Node is conserving memory.

Dockerfile
```diff
@@ -3,10 +3,14 @@ FROM node:16-slim as static_files
...
 ENV PATH /code/node_modules/.bin:$PATH
 COPY ./spa/package.json ./spa/package-lock.json /code/
-RUN npm install --silent
+
+# Skip installing Cypress binaries and security/funding audits, use package-lock.json strictly
+RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit --no-fund --silent
 COPY ./spa /code/spa/
 WORKDIR /code/spa/
-RUN NODE_ENV=production npm run build
+
```
x110dc, line 11: This same Dockerfile/image gets deployed to production. So, I'd rather not have something name "for Heroku" that will run there (since I think it will be confusing for ppl that look at this later). I'm okay with a separate Dockerfile for Heroku (and may want to do that anyway for other performance reasons). 

For now maybe just rename it ":low-memory" or similar?

spa/package.json
x110dc: 👍

spa/src/components/DonationPageRouter.test.tsx
x110dc: I think this comment would be great as a comment in the actual code.

x110dc (changes requested):

Dockerfile
nrh-cklimas: Makes sense. Updated in 50ddab3a3acf88376dd2bbb9bf535f69fe8e0495

spa/src/components/DonationPageRouter.test.tsx
nrh-cklimas: 👍 Added in 7b52bb9916a3cf270d1ec7f19f81388422f2b2be

x110dc (approved):

PR #760, opened 2022-10-15
DEV-2468 | Register apple pay domain after successful stripe connect

benjaminEwhite:

#### What's this PR do?

Changes how Stripe Apple Pay Domain registration happens. Before, there was a signal on revenue program save. On first creation of RP, that signal would call `RevenueProgram.stripe_create_apple_pay_domain`.

This is problematic because an RP's payment provider will not have a Stripe account ID until they've connected accounts.

Therefore, this PR changes when `stripe_create_apple_pay_domain` gets called. Now, we call it from within `apps.organizations.views.create_stripe_account_link_complete` if the retrieved Stripe account has charges enabled.

Otherwise we have to manually run script for all orgs to get apple pay working.

In the review app, create a new account. 

In another window, as an admin user, log into to Django admin. You should see that the new account's RP has `None` for `domain_apple_verified_date`.

Now as the org admin user, go through account creation and the full Stripe Account connect flow.

Once the account is connected, back in Django admin, look at `domain_apple_verified_date` for the RP. It should now have a date (question: does this actually work in review app env though?0.

No

Yes

N/A

I noticed we have both `apps.organizations.tests.test_org_models.py` and `apps.organizations.tests.test_models.py`. I'm not sure if one is duplicative of the other or if they're meant to have different sorts of tests, but I was confused by this, and don't recall this being the case before.

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-2468

No

No


njharman:
>  both apps.organizations.tests.test_org_models.py and apps.organizations.tests.test_models.py. I

Apparently (git blame) this was me. No idea why... kind of thing that would make me angry/yell (in my head) "what a yahoo" at person doing it.

Created https://news-revenue-hub.atlassian.net/browse/DEV-2412 to resolve.

njharman (approved):
Good coverage. Looks fine.

I'm guessing good reason not, but it looked like "stripe_create_apple_pay_domain() were just made idempotent and could run on every save to an RP" Would have been more robust.  I'm not sure if SPA can "save RP" to try to recreated apple pay. Or if create_stripe_account_link_complete / its process is repeatable  (to create apple pay after say some stripe issue was resolved)

x110dc (approved):

PR #769, opened 2022-10-20
DEV-2568: Config Sentry performance monitoring

GuiMend:

n/a

#### What's this PR do?
Configures Sentry performance monitoring + removes some code debt regarding routing setup

So that we have better metrics regarding frontend performance and monitor if something goes wrong.

Not sure

no

n/a

n/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-2568

no

no


spa/src/components/routes/RouterSetup.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
nrh-cklimas, line 1:
```suggestion
import { Suspense } from 'react';
```
(since that's our usual pattern)

nrh-cklimas, line 1: Same thoughts as about the new hook--could this be a TSX file, and let's put unit tests on this. We don't need to test all the functionality of these components, just that this component wraps with the components we expect. An approach would be to mock the imported components with code like:

```tsx
jest.mock('components/errors/ChunkErrorBoundary', () => ({ children }) =>
  <div data-testid="mock-error-boundary>{children}</div>
);
```

And then from there, expect that these test ids are in the document as well as children.

spa/src/components/routes/RouterSetup.js
```diff
@@ -0,0 +1,20 @@
...
+
+import ChunkErrorBoundary from 'components/errors/ChunkErrorBoundary';
+import AddSlashToRoutes from 'components/routes/AddSlashToRoutes';
+import GlobalLoading from 'elements/GlobalLoading';
+import { BrowserRouter, Switch } from 'react-router-dom';
+
+const RouterSetup = ({ children }) => (
```
nrh-cklimas, line 8: I like standardizing our routing in this component a lot.

nrh-cklimas (changes requested):
Same as Daniel, I don't see any data there for the review app--I think this is because useSentry looks for an environment variable named `SENTRY_ENABLE_FRONTEND`, and we're not doing that in our builds. I think that also means that the question about environment changes in the PR desc needs to update to include that.

I think first thing here is to get the env variable set on the review app so we can verify that we're getting data, and then from there, improve test coverage.

spa/src/hooks/useSentry.js
```diff
@@ -2,10 +2,14 @@ import { useEffect } from 'react';
```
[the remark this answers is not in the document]
GuiMend: Thanks for helping with the tests. They were commited [bbd80d4](https://github.com/newsrevenuehub/rev-engine/pull/769/commits/bbd80d43c0d98b7197e73593012aa0e45bb93f05)

spa/src/hooks/useSentry.ts
```diff
@@ -0,0 +1,39 @@
...
+import { createBrowserHistory } from 'history';
+import { Route } from 'react-router-dom';
+
+// Const
+import { ENVIRONMENT, SENTRY_DSN_FRONTEND, SENTRY_ENABLE_FRONTEND } from 'settings';
+
+const history = createBrowserHistory();
+
+/**
+ * Typically, just loading sentry at build time is sufficient.
```
nrh-cklimas, line 15:
```suggestion
 * Typically, just loading sentry at load time is sufficient.
```
I think that's what you mean here?

spa/src/hooks/useSentry.ts
```diff
@@ -0,0 +1,39 @@
...
+// Sentry
+import * as Sentry from '@sentry/react';
+import { BrowserTracing } from '@sentry/tracing';
+import { createBrowserHistory } from 'history';
+import { Route } from 'react-router-dom';
+
+// Const
+import { ENVIRONMENT, SENTRY_DSN_FRONTEND, SENTRY_ENABLE_FRONTEND } from 'settings';
+
+const history = createBrowserHistory();
```
nrh-cklimas, line 12: I think this is why we're still not seeing any info in Sentry on the review app (I tried clicking around after the most recent changes deployed). I _think_ what might work is to pull history out of `useHistory()` so that it connects to the history we are actually using in the router.

I think this means that the hook will need to be called by RouterSetup now, not App, so useHistory() will see the history the router has.

GuiMend: I've tried changing the history object to `useHistory`. Let's see if anything changes

nrh-cklimas: I get a white page on the review app now with a console error, I think because `useHistory()` is still in App, and it doesn't have a history to access. It looks like this doesn't fail locally, maybe because Sentry won't try to initialize on localhost? So I think you'll have to test in the review app to be sure it works.

GuiMend: think now it's solved

nrh-cklimas (approved):

x110dc (approved):

PR #772, opened 2022-10-21
Dev 2694 fix pay fees metadata

nrh-cklimas:

n/a

#### What's this PR do?

Fixes a bug with the contribution page where two things were not accurately recorded:

- Whether the user paid fees
- The original amount the user chose to contribute (e.g. before fees)

I converted the underlying functions to TypeScript and expanded test coverage on these files.

Gives us accurate tracking of contributions.

To

1. Make a test contribution while choosing to pay fees.
2. Look at the contribution in Stripe and check that the metadata shows that the `agreed_to_pay_fees` field is true.
3. Confirm the Stripe metadata also shows that the `donor_selected_amount` field shows the amount you chose to contribute in step 1 minus fees.
4. Make a test contribution while not choosing to pay fees.
5. Repeat step 2, except this time Stripe should show the metadata `agreed_to_pay_fees` is false and `donor_selected_amount` shows the correct number.

No.

I added extra Cypress tests for paying/not paying fees to check that the right `agreed_to_pay_fees` property is being sent.

I also added expanded unit tests on the functions that control that submission.

n/a

n/a I think?

No.

- https://news-revenue-hub.atlassian.net/browse/DEV-2694
- https://news-revenue-hub.atlassian.net/browse/DEV-2699

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2698


spa/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "baseUrl": "src",
-    "target": "es5",
+    "target": "es2015",
```
nrh-cklimas, line 4: Reason for this is to allow `for (const key of object)`.

spa/src/utilities/calculateStripeFee.ts
```diff
@@ -22,19 +22,28 @@ function calculateStripeFee(amount, interval, isNonProfit) {
     NOTE: We are not including any VAT or GST, or any other taxes here, since these are donations.
   */
 
-  const amountInt = parseFloat(amount);
-  if (isNaN(amountInt)) return null;
+  const amountInt = parseFloat(amount as string);
+
+  if (isNaN(amountInt) || amountInt < 0) {
```
nrh-cklimas, line 27: We didn't have a check for negative numbers previously.

spa/src/components/paymentProviders/stripe/stripeFns.ts
```diff
@@ -1,50 +1,50 @@
...
+  frequency: ContributionInterval,
+  rpIsNonProfit: boolean
+) {
+  let total = parseFloat(amount as string);
+
+  if (shouldPayFee) {
+    const fee = calculateStripeFee(amount, frequency, rpIsNonProfit);
+
+    // Fee might be null if amount isn't a number.
+    // TODO handle this better?
```
nrh-cklimas, line 26: I'm not sure what we should do in this case.

spa/src/components/paymentProviders/stripe/stripeFns.ts
```diff
@@ -95,18 +121,31 @@ export function serializeData(formRef, state) {
  * @param {object} data - JSON-serialized form data
  * @returns a response from stripe.createPaymentMethod
  */
-export async function createPaymentMethod(stripe, card, data) {
-  const billing_details = {};
+export async function createPaymentMethod(stripe: Stripe, card: StripeCardElement, data?: PaymentMethodFormData) {
+  const billing_details: PaymentMethodCreateParams.BillingDetails = {};
   if (data?.first_name || data?.last_name) {
-    billing_details.name = `${data.first_name} ${data.last_name}`;
+    billing_details.name = `${data.first_name ?? ''} ${data.last_name ?? ''}`.trim();
```
nrh-cklimas, line 127: This fixes an incidental bug I noticed, that if only first or last name was set we were setting names like `John undefined`.

spa/src/components/donationPage/DonationPage.js
```diff
@@ -34,6 +34,7 @@ import { CONTRIBUTION_INTERVALS } from 'constants/contributionIntervals';
 import { CSRF_HEADER } from 'settings';
 
 function authorizePayment(paymentData, paymentType, csrftoken) {
+  console.log({ paymentData });
```
GuiMend, line 37: COnsole log got commited

spa/src/components/paymentProviders/stripe/stripeFns.ts
GuiMend: You mean when "amount" is not a number? Couldn't we just force `amount: number` and just make sure when we cann the function we conver the value to number?

spa/src/components/paymentProviders/stripe/stripeFns.ts
```diff
@@ -130,7 +169,7 @@ export function getPaymentSuccessUrl({
       rpSlug,
       pathName,
       stripeClientSecret
-    }).filter(([_, v]) => [undefined, null].includes(v))
+    }).filter(([, v]) => [undefined, null].includes(v as any))
```
GuiMend, line 172: Can we avoid using `any` here?

spa/src/components/paymentProviders/stripe/stripeFns.ts
nrh-cklimas: The way amount works right now in the contribution form is confusing. It is usually a number, but if the user hasn't chosen a contribution amount or typed anything into the freeform field, it's the empty string `''`. `calculateStripeFee('')` is `null`. A confusing aspect is, `calculateStripeFee(0)` returns around 30 cents (depending on nonprofit status).

The logic here is if the fee is null, don't do anything with that. But overall I feel like we could be doing better here, but I don't know exactly how to approach it without redoing everything.

spa/src/components/donationPage/DonationPage.js
nrh-cklimas: Thanks for catching this--fixed in 60bc2f3dda74e87601c8564429e5e2fc05f34d65

spa/src/components/paymentProviders/stripe/stripeFns.ts
nrh-cklimas: I don't think so, but there could be a trick I don't know.

The reason why a cast is needed is because TypeScript correctly says "these types are specified as never being null or undefined, those will never be in `[undefined, null]`. So we need the cast to bypass the type check. I don't like removing the check until we have more code in TypeScript and can trust it to flag these issues in all cases.

GuiMend (approved):
👍

spa/cypress/e2e/01-donationPage.cy.js
```diff
@@ -383,13 +383,18 @@ describe('User flow: happy path', () => {
     fillOutDonorInfoSection();
     fillOutAddressSection();
     fillOutReasonForGiving();
+
+    // The donation page fixture has fees paid by default, but make sure that's true.
```
benjaminEwhite, line 387: Should change this to "contribution page" instead of "donation page", to align with [DEV-1446](https://news-revenue-hub.atlassian.net/browse/DEV-1446).

FWIW, was going to leave change as suggestion, but for some reason, that button is not available in this PR...

spa/cypress/e2e/01-donationPage.cy.js
```diff
@@ -440,7 +445,86 @@ describe('User flow: happy path', () => {
...
+    cy.intercept(
+      { method: 'POST', url: getEndpoint(AUTHORIZE_ONE_TIME_STRIPE_PAYMENT_ROUTE) },
+      {
+        body: { provider_client_secret_id: fakeStripeSecret, email_hash: fakeEmailHash },
+        statusCode: 201
+      }
+    ).as('create-one-time-payment');
+    cy.intercept({ method: 'POST', url: 'https://r.stripe.com/0' }, { statusCode: 201 });
+    cy.intercept({ method: 'POST', url: 'https://m.stripe.com/0' }, { statusCode: 201 });
+    cy.intercept({ method: 'GET', url: 'https://api.stripe.com/**' }, { statusCode: 200 });
```
benjaminEwhite, line 459: There's a pre-existing Cypress command you could use here (refactoring it to also intercept api.stripe.com/**). See [here](https://github.com/newsrevenuehub/rev-engine/blob/DEV-2694-fix-pay-fees-metadata/spa/cypress/support/commands.js#L101).

spa/cypress/e2e/01-donationPage.cy.js
```diff
@@ -366,7 +366,7 @@ describe('User flow: happy path', () => {
     cy.visitDonationPage();
   });
 
-  specify('one-time contribution', () => {
+  specify('one-time contribution, paying fees', () => {
```
benjaminEwhite, line 369: Not a big deal, but think you could parametrize the `pay fees` boolean, and compress the two one-time contribution tests into a single block. 

https://dev.to/tinesoft/be-smart-write-parameterized-tests-with-jest-3eab

spa/cypress/e2e/01-donationPage.cy.js
```diff
@@ -440,7 +445,86 @@ describe('User flow: happy path', () => {
...
+          pageSlug: livePageOne.slug,
+          rpSlug: livePageOne.revenue_program.slug,
+          pathName: `/${livePageOne.slug}/`,
+          stripeClientSecret: fakeStripeSecret
+        })
+      );
+    });
+  });
+
+  specify('recurring contribution, paying fees', () => {
```
benjaminEwhite, line 527: Could also parametrize the two recurring contribution tests.

I suppose you could even combine all 4 contribution tests here into a single parametrized test, but that would require a bit of overhead to decide which fixture to use, etc.

spa/cypress/e2e/03-donationPageList.cy.js
```diff
@@ -119,8 +118,7 @@ describe('Donation page list', () => {
         cy.intercept({ method: 'GET', pathname: getEndpoint(USER) }, { body: orgAdminWithContentFlagAndOneRP });
         cy.intercept({ method: 'POST', pathname: getEndpoint(LIST_PAGES) }, { statusCode: 200 }).as('createNewPage');
         cy.visit(CONTENT_SLUG);
-        cy.wait('@listPages');
-        cy.get('button[aria-label="New Page"]').click();
+        cy.wait('@listPages').then(() => cy.get('button[aria-label="New Page"]').click());
```
benjaminEwhite, line 121: Why the `.then` here? As I understand it, simply calling `cy.wait` as in original should cause the test not to move on until list pages has run.

benjaminEwhite (changes requested):
Looks good overall. A couple of suggestions (which you may choose to reject) re: tests. And one `.then` that we might not need.

spa/cypress/e2e/03-donationPageList.cy.js
nrh-cklimas: I was troubleshooting this with Gui. What we both saw is that Cypress would click New Page immediately, before the rest of the pages loaded, and would try to create a page numbered 1, when it should be creating a page numbered 2 (because 1 already exists). Annoyingly, this only seems to happen in CI-- Gui thought it was because CI is slower than running locally.

I see https://github.com/cypress-io/cypress-documentation/issues/1508 says you shouldn't need to chain on it, but if so, I don't understand why we're seeing this test fail. I think a lot of is it I don't know Cypress very well.

spa/cypress/e2e/01-donationPage.cy.js
nrh-cklimas: :+1: Updated in 58f5d1361dda4712589f6fb36243360f5f6ac7a5

spa/cypress/e2e/01-donationPage.cy.js
nrh-cklimas: I liked this idea a lot. I did that also in 58f5d1361dda4712589f6fb36243360f5f6ac7a5

spa/cypress/e2e/01-donationPage.cy.js
nrh-cklimas: Same thing, 58f5d1361dda4712589f6fb36243360f5f6ac7a5--I looked at trying to generate all four but it seemed a little hairy to make sure the logic for all was correct.

spa/cypress/e2e/03-donationPageList.cy.js
nrh-cklimas: Take a look: 4fcccdc45c2e15093782775f725f04c96e50b2f2

benjaminEwhite (approved):
LGTM

x110dc (approved):

PR #788, opened 2022-10-27
Dev 2475 resend contributions receipt

harsha09:
#### What's this PR do?
Creates an endpoint to email contribution to users.
The endpoint accepts contribution_id which is an integer.
Retrieves `PaymentIntent` and `Invoice` from `stripe` to populate `amount, interval, receipt_email_id and created_date`.

Users should be able to re-send the contribution receipt to their registered E-Mail address.

Right now this is not testable from UI. Should be able to test when [DEV-2474] is completed.
Can test using curl or postman using below endpoint details

Endpoint: api/v1/payments/email-contribution/
Method: POST
Parameter: contribution_id <integer>
Response: 200, 417 (When contribution status is not PAID)
Response Body:
{"detail": "success": "contributor_email": <email>}

No

Yes

May be using user manual or some documentation.

NA

Not yet

Parent ticket [DEV-1606]

NA

[DEV-2474] for UI.

[DEV-2474]: https://news-revenue-hub.atlassian.net/browse/DEV-2474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-1606]: https://news-revenue-hub.atlassian.net/browse/DEV-1606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/views.py
```diff
@@ -401,3 +412,44 @@ def destroy(self, request, pk):
...
+        )
+
+    payment_intent = stripe.PaymentIntent.retrieve(
+        id=contribution.provider_payment_id, stripe_account=stripe_account_id
+    )
+
+    interval = ContributionInterval.ONE_TIME
+    if payment_intent.invoice:
+        invoice = stripe.Invoice.retrieve(id=payment_intent.invoice, stripe_account=stripe_account_id)
+        interval = payment_interval_from_stripe_invoice(invoice, ContributionInterval)
```
benjaminEwhite, line 438: Why aren't we just using `contribution.interval`?

apps/contributions/views.py
```diff
@@ -401,3 +412,44 @@ def destroy(self, request, pk):
         # TODO: [DEV-2438] return the canceled sub and update the cache
 
         return Response({"detail": "Success"}, status=status.HTTP_204_NO_CONTENT)
+
+
+@api_view(["POST"])
+@permission_classes([IsAuthenticated, HasRoleAssignment])
```
benjaminEwhite, line 418: With these permission classes alone, it would be possible for an authenticated user with a role assignment to send email contributions for any contributor. We should be validating that the requesting user is the same user who created the underlying NRE contribution and related Stripe entities. Should also have tests that prove this new permission logic.

apps/contributions/views.py
```diff
@@ -401,3 +412,44 @@ def destroy(self, request, pk):
...
+def email_contribution(request):
+    email_id = request.user.email
```
benjaminEwhite, line 420: I think we should be confirming that the requesting user is the contributor for the requested contribution. Once we know that, we should send email to that user's address.

Given current permissions, it would be possible for a user to gain access to receipts they shouldn't be able to see: if they guess a contribution ID and that contribution, and if the payment_intent doesn't have `.receipt_email` defined, then the receipt will be sent to the requesting user, not the contributor email.

benjaminEwhite (changes requested):

apps/contributions/stripe_contributions_provider.py
```diff
@@ -82,6 +83,7 @@ def is_modifiable(self):
 
     @property
     def interval(self):
+        # [TODO]: logic being ambigious to be replaced by utils.payment_interval_from_stripe_invoice
```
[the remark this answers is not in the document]
harsha09: I've added this comment to avoid code duplication and don't want to touch this particular code in this PR.
But, I've removed this as it's not required now, because I've modified to get interval from contribution.

apps/contributions/views.py
```diff
@@ -401,3 +412,44 @@ def destroy(self, request, pk):
...
+    if contribution.status != ContributionStatus.PAID:
+        return Response(
+            {"detail": f"Contribution status is {contribution.status} expectected status is {ContributionStatus.PAID}"},
+            status=status.HTTP_417_EXPECTATION_FAILED,
+        )
+
+    payment_intent = stripe.PaymentIntent.retrieve(
+        id=contribution.provider_payment_id, stripe_account=stripe_account_id
+    )
+
```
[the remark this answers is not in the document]
harsha09: It'll return 500 status code but response is not JSON. I've modified it to return a JSON response.

apps/contributions/views.py
harsha09: Yeah, that's a nice suggestion. I've modified it to get it from the Contribution object. I hope this will be fine @x110dc as we are anyhow getting contribution object.

apps/contributions/utils.py
```diff
@@ -22,3 +23,23 @@ def get_sha256_hash(string):
...
+
+    if interval == "year" and interval_count == 1:
+        return contribution_interval_model.YEARLY
+    if interval == "month" and interval_count == 1:
+        return contribution_interval_model.MONTHLY
+
+    return None
+
+
+def convert_stripe_date_to_datetime(stripe_date):
```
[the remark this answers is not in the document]
harsha09: Addressed using 48d0e2fcc12d4eb080b1091f91d756864562222a

apps/contributions/views.py
harsha09: Will contributors will be able to login to the dashboard, they'll be able to view contributor portal with the magic token if I'm not wrong.
This API will be integrated into the frontend of contributor dashboard.
LMK if I'm wrong.

apps/contributions/views.py
harsha09: As this API will be called from Contributions Dashboard, Contributors doesn't have access to it as they'll be routed to magic link url.
As per [DEV-1606], it'll be logically correct if the access is provided to RP Admin (Revenue Program Managers) where a particular contribution falls under his one of RPs.

benjaminEwhite: There's nothing stopping someone from making requests to the API from somewhere other than contributions dashboard. For instance, if I sign up for a free account (as an org), I could be authenticated, and have a role assignment, and make a request to this endpoint and gain access to receipts that don't belong to me.

apps/contributions/views.py
benjaminEwhite: Ah, I see. So even if it's not contributors accessing this, the concern remains: we're not doing anything to ensure that the requesting user's org should have access to the requested contribution. We need to ensure that the requested contribution belongs to an RP that the requesting user has access to (via role assignment).

benjaminEwhite (changes requested):

apps/contributions/views.py
harsha09: I've created `IsRPAdmin` and `HasAccessToContribution` permissions and enforced them on the view.
I think this should solve the problem.

apps/contributions/views.py
harsha09: I've created IsRPAdmin and HasAccessToContribution permissions and enforced them on the view.
I think this should solve the problem.

apps/api/permissions.py
```diff
@@ -134,3 +135,26 @@ def has_permission(self, request, view):
...
+class HasAccessToContribution(permissions.BasePermission):
+    """
+    Check whether the user is an RP Admin and the contribution belongs to him.
+    Prerequisite permissions: IsAuthenticated, HasRoleAssignment, IsRPAdmin.
+    """
+
+    def has_permission(self, request, view):
+        contribution = get_object_or_404(Contribution, pk=request.data.get("contribution_id"))
+        return (
+            request.user.roleassignment.revenue_programs.filter(id=contribution.revenue_program.id).first() is not None
```
benjaminEwhite, line 159:
```suggestion
            request.user.roleassignment.revenue_programs.filter(id=contribution.revenue_program.id).exists()
```

apps/api/permissions.py
```diff
@@ -134,3 +135,26 @@ def has_permission(self, request, view):
...
+
+
+class IsRPAdmin(permissions.BasePermission):
+    """
+    Permission to check whether the logged user is a Revenue Program admin.
+    Prerequisite permissions: IsAuthenticated and HasRoleAssignment.
+    """
+
+    def has_permission(self, request, view):
+        return request.user.roleassignment.role_type == Roles.RP_ADMIN.value
```
benjaminEwhite, line 147: We shouldn't be limiting to RPs. Our default user type is now an ORG_ADMIN (that's what people who go through self-onboarding get).

As long as we know that someone is authenticated, has a role assignment, and has access to the contribution by virtue of their role assignment, that should be enough.


FWIW, you can see [this custom validator ](https://github.com/newsrevenuehub/rev-engine/blob/main/apps/common/validators.py#L51)(which won't work here cause context if function-based view instead of class-based + serializer) which does something similar.

apps/contributions/tests/test_views.py
```diff
@@ -1150,12 +1155,55 @@ def test_one_time_contribution(self, payment_intent_retrieve, send_email_task):
...
         assert email_attrs["contribution_interval"] == contribution.interval
         assert email_attrs["contribution_interval_display_value"] == contribution.interval
+
+    def test_when_contribution_is_from_another_rp(self):
+        self.client.force_authenticate(user=self.rp_user)
+        contribution = Contribution.objects.filter(
+            donation_page_id=DonationPage.objects.filter(revenue_program_id=self.org1_rp2.id).first().id
+        ).first()
+        response = self.client.post(self.url, {"contribution_id": contribution.id})
+        assert response.status_code == 403
```
benjaminEwhite, line 1176: More relevant at the moment is when contribution is for another org (and user is org user)

apps/contributions/tests/test_views.py
```diff
@@ -1150,12 +1155,55 @@ def test_one_time_contribution(self, payment_intent_retrieve, send_email_task):
...
+
+    def test_when_logged_user_is_not_revenue_program_manager(self):
```
benjaminEwhite, line 1178: See above comment. We shouldn't restrict to rp role type.

apps/contributions/tests/test_views.py
```diff
@@ -1150,12 +1155,55 @@ def test_one_time_contribution(self, payment_intent_retrieve, send_email_task):
...
+        self.client.force_authenticate(user=self.org_user)
+        response = self.client.post(self.url, {"contribution_id": 1234})
+        assert response.status_code == 403
+
+    @mock.patch("apps.contributions.views.send_templated_email.delay")
+    @mock.patch("stripe.PaymentIntent.retrieve")
+    def test_contributor_email_when_receipt_email_in_payment_intent(self, payment_intent_retrieve, send_email_task):
+        self.client.force_authenticate(user=self.rp_user)
```
benjaminEwhite, line 1186: ditto re: rp_user

apps/contributions/views.py
```diff
@@ -411,9 +413,8 @@ def destroy(self, request, pk):
 
 
 @api_view(["POST"])
-@permission_classes([IsAuthenticated, HasRoleAssignment])
+@permission_classes([IsAuthenticated, HasRoleAssignment, IsRPAdmin, HasAccessToContribution])
```
benjaminEwhite, line 416: Ditto re: `IsRPAdmin`

benjaminEwhite (changes requested):

apps/api/permissions.py
```diff
@@ -134,3 +135,26 @@ def has_permission(self, request, view):
...
+
+
+class HasAccessToContribution(permissions.BasePermission):
+    """
+    Check whether the user is an RP Admin and the contribution belongs to him.
```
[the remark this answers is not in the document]
harsha09: Have modified to `Check whether the user has permission to access contribution via roleassignment.`

apps/api/permissions.py
harsha09: Modified the permissions for `ORG_ADMIN` and `RP_ADMIN` who has access to the contribution will be able to send the receipt.

apps/api/permissions.py
harsha09: This is not needed for the modified code.

apps/contributions/tests/test_views.py
harsha09: Added testcase for org user as well

apps/contributions/tests/test_views.py
harsha09: This will be addressed by above mentioned comments.

apps/contributions/views.py
harsha09: This will be addressed by above comments.

apps/contributions/tests/test_views.py
harsha09: This will be addressed by above comments.

apps/contributions/tests/test_views.py
```diff
@@ -1181,3 +1181,122 @@ def test_payment_success_view():
...
+        )
+
+    def test_when_no_contribution_found(self):
+        self.client.force_authenticate(user=self.rp_user)
+        response = self.client.post(self.url, {})
+        assert response.status_code == 404
+
+    def test_when_contribution_status_is_not_paid(self):
+        self.client.force_authenticate(user=self.rp_user)
+        contribution = [x for x in Contribution.objects.all() if x.status != ContributionStatus.PAID][0]
```
benjaminEwhite, line 1230:
```suggestion
        contribution =  Contribution.objects.filter(status=ContributionStatus.PAID).first()
```

PR #793, opened 2022-10-29
DEV-2470: add google cloud publisher

leo-dcfa:

- ~~Discuss how to handle service accounts and if we should have a `fake-service-account.json` file to allow the application to boot up; discuss alternatives~~
- ~~Discuss if Signals is an acceptable solution here~~ - moved away from this solution after running into issues
- ~~Decide if we should implement emulators for local development as part of this PR or create another ticket for it~~
- ~~Finally, deploy this to a review app and validate that it works e2e~~

#### What's this PR do?

- Publish messages to Google Cloud PubSub when a new contributor is created
- Added Google Cloud PubSub to docker-compose emulator for local development
- Added script to test Google Cloud PubSub by subscription to new user topic

When a new user is created on Production, we’d like to build passive awareness, as this is a key performance indicator for our self-service strategy. 

- Add the following variables to your `.envrc`
```text
export NEW_USER_TOPIC=new-nre-customer-test
export GOOGLE_CLOUD_PROJECT=revenue-engine
export PUBSUB_PROJECT_ID=revenue-engine
export GOOGLE_APPLICATION_CREDENTIALS=fake-service-account.json
export PUBSUB_EMULATOR=[::1]:8085
export PUBSUB_EMULATOR_HOST=[::1]:8085
```
- Start the dev server + google cloud pub sub emulator with `make run-gcloud-pub-sub`
- On a separate terminal window run ` python manage.py shell`
- On the same window where the shell is running, copy and the following script and hit enter
```python
# this is a helpful script to connect to topic from the Django shell
import os

from google.cloud import pubsub_v1


topic_name = "projects/{project_id}/topics/{topic}".format(
    project_id=os.getenv("GOOGLE_CLOUD_PROJECT"), topic=os.getenv("NEW_USER_TOPIC")
)

print(f"Subscribing to topic {topic_name}")

subscription_name = "projects/{project_id}/subscriptions/{sub}".format(
    project_id=os.getenv("GOOGLE_CLOUD_PROJECT"),
    sub="TEST",
)

print(f"Will attempt to delete existing subscription with name {subscription_name}")

try:
    pubsub_v1.SubscriberClient().delete_subscription(request={"subscription": subscription_name})
except:  # noqa
    print("Subscription did not exist prior; will continue")


def callback(message):
    print(message.data)
    message.ack()


with pubsub_v1.SubscriberClient() as subscriber:
    subscriber.create_subscription(name=subscription_name, topic=topic_name)
    print(f"New subscription created; listening messages in topic {topic_name}")
    future = subscriber.subscribe(subscription_name, callback)
    future.result()
```
- Go to [create-account page](http://localhost:3000/create-account/) and create a new account (please note as part of tests I had to temporarily disable bad-actor api, user views.py, line 272)
- Go back to the terminal window where you ran the subscribe script and validate that the email address you entered has been printed out to the terminal
- Go back to the create-account page and create an account ending with `@fundjournalism.org`
- Go back to the terminal window where you ran the subscribe script and validate that no email address is printed out

N/A

Yes

N/A

- Yes, see comment on file regarding tests not working as expected
- I was not able to find a way of creating topics as part of starting the emulator container so I added a script to create a new topic as part of the makefile, prior to starting the dev server. There must be a better way of doing this but my research did not render results

N/A

[DEV-2470]

Yes, TBC

To be decided

[DEV-2470]: https://news-revenue-hub.atlassian.net/browse/DEV-2470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/google_pub_sub/tests/tests.py
```diff
@@ -0,0 +1,22 @@
...
+
+class PublisherTests(TestCase):
+    @pytest.mark.no_patch_google_cloud_pub_sub_publisher
+    @patch("apps.google_pub_sub.publisher.pubsub_v1.PublisherClient", MagicMock())
+    def test_publishes_to_google_cloud_pub_sub(self):
+        topic = "topic"
+        publisher = GoogleCloudPubSubPublisher()
+        message = Message(data="some message")
+        result = publisher.publish(topic, message)
+        assert result
```
leo-dcfa, line 22: I am running into issues with testing this properly, so this is just a placeholder test.

My test was initially this:
```python
@pytest.mark.no_patch_google_cloud_pub_sub_publisher
    @patch("apps.google_pub_sub.publisher.pubsub_v1.PublisherClient", MagicMock())
    def test_publishes_to_google_cloud_pub_sub(self):
        topic = "topic"
        expected = "result"
        future = asyncio.Future()
        future.set_result(expected)
        full_topic_path = f"projects/{GOOGLE_CLOUD_PROJECT}/topics/{topic}"
        publisher = GoogleCloudPubSubPublisher()
        publisher.client.topic_path.return_value = full_topic_path
        publisher.client.publish.return_value = future
        message = Message(data="some message")
        result = publisher.publish(topic, message)
        assert expected == result
        publisher.client.topic_path.assert_called_with(GOOGLE_CLOUD_PROJECT, topic)
        publisher.client.publish.assert_called_with(full_topic_path, message.data)
```
But for some reason, it looks like the `.publish` call from `GoogleCloudPubSubPublisher` patch from autouse fixture is not being removed. Any ideas on how to do this better are very welcome

leo-dcfa: https://github.com/newsrevenuehub/rev-engine/pull/793/files#diff-a31c7ed5d35f5ed8233994868c54d625b18e6bacb6794344c4531e62bd9dde59R15

I also tried this:

```python
def patch_google_cloud_pub_sub_publisher(request):
    marker = request.node.get_closest_marker("no_patch_google_cloud_pub_sub_publisher")
    if marker:
        yield
    else:
        with mock.patch("apps.google_pub_sub.publisher.GoogleCloudPubSubPublisher.publish", MagicMock()):
            yield
```

But that resulted in errors on other classes:

```
self = <django.db.backends.postgresql.base.DatabaseWrapper object at 0x7f7956c05c00>
name = None

    @async_unsafe
    def create_cursor(self, name=None):
        if name:
            # In autocommit mode, the cursor will be used outside of a
            # transaction, hence use a holdable cursor.
            cursor = self.connection.cursor(name, scrollable=False, withhold=self.connection.autocommit)
        else:
>           cursor = self.connection.cursor()
E           psycopg2.InterfaceError: connection already closed

```

See this job for details

revengine/settings/base.py
```diff
@@ -79,6 +79,15 @@
 MEDIA_LOCATION = os.getenv("MEDIA_LOCATION", "")
 DEFAULT_FILE_STORAGE = os.getenv("DEFAULT_FILE_STORAGE", "django.core.files.storage.FileSystemStorage")
 
+# Google Pub Sub
+GOOGLE_CLOUD_NEW_USER_NOTIFICATION_TOPIC = os.getenv(
+    "GOOGLE_CLOUD_NEW_USER_NOTIFICATION_TOPIC", "new-nre-customer-test"
+)
+GOOGLE_CLOUD_PROJECT = os.getenv("GOOGLE_CLOUD_PROJECT", "revenue-engine")
+GOOGLE_APPLICATION_CREDENTIALS = os.getenv(
```
x110dc, line 87: This variable isn't used anywhere. Why create it? The Google libraries that need this will read it from the environment directly.

revengine/settings/base.py
```diff
@@ -79,6 +79,15 @@
 MEDIA_LOCATION = os.getenv("MEDIA_LOCATION", "")
 DEFAULT_FILE_STORAGE = os.getenv("DEFAULT_FILE_STORAGE", "django.core.files.storage.FileSystemStorage")
 
+# Google Pub Sub
+GOOGLE_CLOUD_NEW_USER_NOTIFICATION_TOPIC = os.getenv(
+    "GOOGLE_CLOUD_NEW_USER_NOTIFICATION_TOPIC", "new-nre-customer-test"
```
x110dc, line 84: The default here should be `None` or similar. And let's rename it to something shorter like `NEW_USER_TOPIC`.

revengine/settings/base.py
```diff
@@ -92,6 +101,7 @@
     "apps.element_media",
     "apps.public",
     "apps.config",
+    "apps.google_pub_sub",
```
x110dc, line 104: I don't think this needs to be a full-fledged Django app.

apps/google_pub_sub/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
```
x110dc, line 3: I don't think this needs to be a full Django app. Just a class at most I'd think?

.pre-commit-config.yaml
```diff
@@ -17,6 +17,7 @@ repos:
       - id: detect-aws-credentials
         args: [--allow-missing-credentials]
       - id: detect-private-key
+        exclude: "fake-service-account.json"
```
x110dc, line 20: I'd rather not complicate local set up with this at this point unless really necessary. Can we remove this, the fake json and related? How about just adding a `ENABLE_PUBSUB` variable that defaults to `True` and locally and in test we set it to False?

apps/users/models.py
```diff
@@ -52,6 +55,13 @@ def get_role_type(self):
     def __str__(self):
         return self.email
 
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
```
x110dc, line 59: Won't this notify on every `save()` to the User? We only want to notify on new user creation. See here for a way to tell if it object is new: https://stackoverflow.com/a/35647389

google_cloud_pub_sub/fake-service-account.json
```diff
@@ -0,0 +1,9 @@
+{
```
x110dc, line 1: Let's remove this file.

x110dc (changes requested):
I think we can simplify this. I think we can assume that as long as properly configured the Google library will do what it advertises. We just need safety rails around something being unconfigured or misconfigured so if that's wrong it doesn't break anything. This is equivalent to sending an email or a Slack message. As long as the message is properly formatted a lot of testing shouldn't be necessary. Ping me if you'd like to discuss more.

apps/google_pub_sub/apps.py
leo-dcfa: my initial idea is to have it a more general purpose if we need to publish messages in other contexts.

revengine/settings/base.py
leo-dcfa: ah good point, of course!

apps/users/models.py
leo-dcfa: ah good point, forgot this one

.pre-commit-config.yaml
leo-dcfa: good idea

google_cloud_pub_sub/fake-service-account.json
leo-dcfa: done

revengine/settings/base.py
leo-dcfa: removed

Makefile
```diff
@@ -35,6 +35,7 @@ setup:
 run-dev:
 	@echo 'Running local development'
 	docker-compose up -d --remove-orphans
+	sleep 5 && curl -s -X PUT 'http://localhost:8085/v1/projects/revenue-engine/topics/new-nre-customer-test' # sleep for 5 seconds to allow google cloud to boot up
```
x110dc, line 38: I don't think we're going to want this most of the time. Can we either make this a comment about how to do this or make it a separate Makefile target?

docker-compose.yml
```diff
@@ -14,3 +14,9 @@ services:
       - "54000:5432"
     volumes:
       - dev_revengine_data:/var/lib/postgresql/data
+  gcloud_pub_sub_emulator:
```
x110dc, line 17: Let's put this in `docker-compose-dev.yml`

apps/users/models.py
```diff
@@ -52,6 +55,16 @@ def get_role_type(self):
...
         return self.email
 
+    def save(self, *args, **kwargs):
+        if (
+            settings.ENABLE_PUBSUB
+            and self._state.adding
+            and self.email
+            and not self.email.endswith("@fundjournalism.org")
+        ):
+            self.google_cloud_pub_sub_publisher.publish(settings.NEW_USER_TOPIC, Message(data=self.email))
```
x110dc, line 65: If `settings.NEW_USER_TOPIC` or `settings. GOOGLE_CLOUD_PROJECT` isn't defined I'd rather skip the notification than fail. So maybe do the `super()` first or catch exceptions here?

x110dc (changes requested):

apps/common/utils.py
```diff
@@ -129,6 +129,10 @@ def get_original_ip_from_request(request):
     return remote_addr
 
 
+def google_cloud_pub_sub_is_configured() -> bool:
+    return settings.ENABLE_PUBSUB and settings.NEW_USER_TOPIC and settings.GOOGLE_CLOUD_PROJECT
```
benjaminEwhite, line 133: What you have here is totally fine. I have a slight stylistic preference for following syntax when there's > 1 `and`, but if you prefer the multiple `ands` you should keep them 😄 

```python
    return all([settings.ENABLE_PUBSUB, settings.NEW_USER_TOPIC, settings.GOOGLE_CLOUD_PROJECT])
```

apps/users/google_pub_sub.py
```diff
@@ -0,0 +1,40 @@
+import logging
+from dataclasses import dataclass
+
+from django.conf import settings
+
```
benjaminEwhite, line 5: Nitpicking, but no need for extra line here.

apps/users/google_pub_sub.py
```diff
@@ -0,0 +1,40 @@
...
+from google.cloud import pubsub_v1
+
+
+logger = logging.getLogger(__name__)
```
benjaminEwhite, line 9: Why this syntax here vs. `logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")` in apps/users/models.py?

apps/users/models.py
```diff
@@ -52,6 +60,16 @@ def get_role_type(self):
...
 
+    def save(self, *args, **kwargs):
+        email_publishable = self._email_is_publishable()
+        super().save(*args, **kwargs)
+        if google_cloud_pub_sub_is_configured() and email_publishable:
+            logger.debug("Will publish email to Google Cloud PubSub: %s", self.email)
+            GoogleCloudPubSubPublisher.get_instance().publish(settings.NEW_USER_TOPIC, Message(data=self.email))
+
+    def _email_is_publishable(self) -> bool:
+        return self._state.adding and self.email and not self.email.endswith("@fundjournalism.org")
```
benjaminEwhite, line 71: Could also use the `all` approach here...

apps/users/google_pub_sub.py
```diff
@@ -0,0 +1,40 @@
...
+        future = self.client.publish(topic_path, message.data)
+        result = future.result(timeout=3)
+        logger.info("Published data result with id %s to %s", result, topic)
+        return result
+
+    @classmethod
+    def get_instance(cls):
+        if not cls.__instance:
+            cls.__instance = GoogleCloudPubSubPublisher()
+        return cls.__instance
```
benjaminEwhite, line 40: Might be good to have a doc string explaining why/how this is used. I'm not entirely sure why we are using a singleton pattern here.

apps/users/tests/test_google_pub_sub.py
```diff
@@ -0,0 +1,32 @@
...
+from unittest.mock import patch
+
+from google.cloud.pubsub_v1 import futures
+
+from apps.users.google_pub_sub import GoogleCloudPubSubPublisher, Message
+
+
+class MessageTests(TestCase):
+    def test_encodes_string_to_bytes(self):
+        assert isinstance(Message("irrelevant").data, bytes)
```
benjaminEwhite, line 11: Notwithstanding the fact that there's a bunch of existing test code that uses unitttest, when possible we're trying to move toward using pytest-based tests instead.  Here, can easily do that, and also do away with the class, since there's only a single test function. 

```suggestion
def test_message_encodes_string_to_bytes(self):
        assert isinstance(Message("irrelevant").data, bytes)
```

apps/users/tests/test_models.py
```diff
@@ -49,6 +52,41 @@ def test_str_method_with_different_user_types(self):
...
+    def test_will_publish_external_emails_to_google_pubsub(self, publisher, new_user_topic):
+        topic = "topic"
+        new_user_topic.return_value = topic
+        email = "contributor@some-address.com"
+        create_test_user(email=email)
+        publisher.publish.called_with(new_user_topic, Message(data=email))
+
+    @patch("apps.users.models.google_cloud_pub_sub_is_configured", MagicMock(return_value=True))
+    @patch("apps.users.google_pub_sub.GoogleCloudPubSubPublisher.publish")
+    def test_will_not_publish_external_emails_to_google_pubsub(self, publish):
```
benjaminEwhite, line 67: Should this be "internal" (or maybe "hub") instead of "external"?

apps/users/tests/test_models.py
```diff
@@ -49,6 +52,41 @@ def test_str_method_with_different_user_types(self):
         str(self.rp_admin_user)
         str(create_test_user())
 
+    @patch("apps.users.models.google_cloud_pub_sub_is_configured", MagicMock(return_value=True))
+    @patch("django.conf.settings.NEW_USER_TOPIC")
+    @patch("apps.users.google_pub_sub.GoogleCloudPubSubPublisher")
+    def test_will_publish_external_emails_to_google_pubsub(self, publisher, new_user_topic):
```
benjaminEwhite, line 58: This is minor, but I prefer present tense for these kind of test names. `test_publishes_external_emails_to_google_pubsub`

benjaminEwhite (changes requested):

apps/common/utils.py
leo-dcfa: nice, thanks for the suggestion. am fairly new to python so this sort of suggestion is very helpful

apps/users/google_pub_sub.py
leo-dcfa: good catch, hadn't noticed it

apps/users/google_pub_sub.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(self):
+        self.client = pubsub_v1.PublisherClient()
+        self.project_id = settings.GOOGLE_CLOUD_PROJECT
+
+    def publish(self, topic, message: Message):
+        logger.info("Received data to publish %s", message)
+        topic_path = self.client.topic_path(self.project_id, topic)
+        future = self.client.publish(topic_path, message.data)
+        result = future.result(timeout=3)
```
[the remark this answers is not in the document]
leo-dcfa: nice, thanks

apps/users/google_pub_sub.py
leo-dcfa: Adding this
``text
"""Returns an instance of GoogleCloudPubSubPublisher;
           Singleton pattern was chosen here since it is not necessary to instantiate/authenticate multiple times with
           GoogleCloud, and it will only do it the first time the application requires it. Additionally, memory is saved by
           sharing the same instance across RevEngine"""
```

Let me know if you don't agree but I think it doesn't make sense to authenticate with GoogleCloud and keep multiple instances of the publisher, which is just a thin client that does a bit of formatting/logging

File: google_cloud_pub_sub/subscribe.py [hunk 21]
```diff
@@ -0,0 +1,35 @@
+# this is a helpful script to connect to topic from the Django shell
```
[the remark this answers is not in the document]
leo-dcfa: good point about code coverage, removed it

apps/users/tests/test_models.py
leo-dcfa: good catch

apps/users/tests/test_models.py
leo-dcfa: no problem

apps/users/tests/test_google_pub_sub.py
```diff
@@ -0,0 +1,32 @@
...
+        full_topic_path = f"projects/{google_cloud_project}/topics/{topic}"
+        publisher = GoogleCloudPubSubPublisher()
+        publisher.client = publisher_client
+        publisher_client.topic_path.return_value = full_topic_path
+        publisher_client.publish.return_value = future
+        message = Message(data="some message")
+        result = publisher.publish(topic, message)
+        assert expected == result
+        publisher_client.topic_path.assert_called_with(google_cloud_project, topic)
+        publisher_client.publish.assert_called_with(full_topic_path, message.data)
```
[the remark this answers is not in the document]
leo-dcfa: Is this something you are very passionate about? I am happy to change it, just this sprint I have a lot in my plate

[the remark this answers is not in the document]
x110dc: @leo-dcfa It's our policy to use pytest: https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2047311873/Python+Style+Guide

This ticket isn't urgent, so if you need to come back to it to address something else that's fine.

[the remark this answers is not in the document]
leo-dcfa: no worries, I'll reassign the ticket to myself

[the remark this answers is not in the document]
leo-dcfa: @benjaminEwhite would you mind pointing to what the equivalent in pytest is for `assert_called_with`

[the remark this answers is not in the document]
x110dc: I think there are examples in other tests here like in apps/contributions/tests/test_webhooks.py: 

`assert not mock_process_pi.called`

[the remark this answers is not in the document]
leo-dcfa: Ah I know about that one but slightly different no? Assert that something is called is different from asserting the calling arguments are correct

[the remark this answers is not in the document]
benjaminEwhite: @leo-dcfa we probably should make language in the wiki a bit clearer. I think what we're trying to avoid is the `self.assert...` methods that attach to a unittest test case class.

For this example, I believe you could do something along lines explained here: https://stackoverflow.com/a/69620212

benjaminEwhite (approved):

revengine/settings/base.py
```diff
@@ -79,6 +79,11 @@
 MEDIA_LOCATION = os.getenv("MEDIA_LOCATION", "")
 DEFAULT_FILE_STORAGE = os.getenv("DEFAULT_FILE_STORAGE", "django.core.files.storage.FileSystemStorage")
 
+# Google Pub Sub
+ENABLE_PUBSUB = os.getenv("ENABLE_PUBSUB", "False").lower() in ("true", "1")
```
x110dc, line 83: I think we should probably be consistent here with the rest of our boolean/enable variables and do:
`SENTRY_ENABLE_FRONTEND = os.getenv("SENTRY_ENABLE_FRONTEND", "false").lower() == "true"`

pyproject.toml
```diff
@@ -71,6 +71,7 @@ sorl-thumbnail-serializer-field = "^0.2.1"
 stripe = "^4"
 uwsgi = "^2"
 whitenoise = "^6.2.0"
+google-cloud-pubsub = "^2.13.11"
```
x110dc, line 74: Let's make this `^2.13`

apps/users/models.py
```diff
@@ -52,6 +60,16 @@ def get_role_type(self):
...
 
+    def save(self, *args, **kwargs):
+        email_publishable = self._email_is_publishable()
+        super().save(*args, **kwargs)
+        if google_cloud_pub_sub_is_configured() and email_publishable:
+            logger.debug("Will publish email to Google Cloud PubSub: %s", self.email)
+            GoogleCloudPubSubPublisher.get_instance().publish(settings.NEW_USER_TOPIC, Message(data=self.email))
+
+    def _email_is_publishable(self) -> bool:
+        return all([self._state.adding, self.email, not self.email.endswith("@fundjournalism.org")])
```
x110dc, line 71: I'd rather not hardcode the `@fundjournalism.org` here. We can maybe make that a variable in the future, but for now let's just remove it and it won't be a big deal to get notifications for those addresses.

Makefile
```diff
@@ -38,9 +38,16 @@ run-dev:
...
+	@echo 'Running local development with Google Cloud Pub Sub Emulator'
+	docker-compose -f docker-compose.yml -f docker-compose-dev.yml up -d --remove-orphans google-cloud-pub-sub db
+	sleep 5 && curl -s -X PUT 'http://localhost:8085/v1/projects/revenue-engine/topics/new-nre-customer-test' # sleep for 5 seconds to allow google cloud to boot up
+	cd spa; export PORT=3000; npm run start:subdomains &
+	ENABLE_PUBSUB=true python manage.py runserver
+
 run-redis:
 	@echo 'Running local development with redis'
-	docker-compose -f docker-compose.yml -f docker-compose-dev.yml up -d --remove-orphans
+	docker-compose -f docker-compose.yml -f docker-compose-dev.yml up -d --remove-orphans redis
```
x110dc, line 50: To preserve the existing behavior this should be `--remove orphans redis db`

x110dc (changes requested):
A couple of minor change requests.

x110dc (approved):

PR #794, opened 2022-10-30
DEV-2699 | fix donor_selected_amount in Stripe metadata

benjaminEwhite:

#### What's this PR do?

Fixes a bug where the total amount (that is, donor selected amount + fee amount in case of agreeing to pay fees) was being recorded as value for `donor_selected_amount` whereas it should be the selected amount without fee tacked on.

Fixes Salesforce reporting

Make a contribution and agree to pay fees.

In Django admin, you should see that the contribution amount includes pay fees.

Look at the payment intent or subscription in Stripe. Its metadata should have a `donor_selected_amount` value, which should be the amount without fee included.

No

Yes. Existing ones have been updated to ensure this field has correct value.

N/A

No

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-2699

No

N/A

### DIFF
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -284,6 +284,7 @@ class BaseCreatePaymentSerializer(serializers.Serializer):
     captcha_token = serializers.CharField(max_length=2550, allow_blank=True, write_only=True)
     provider_client_secret_id = serializers.CharField(read_only=True)
     email_hash = serializers.CharField(read_only=True)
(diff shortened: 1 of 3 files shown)


njharman (approved):
lgtm, assuming the serializer was "incoming" from SPA to Django and the bug was in SPA / serializer not separating contributor's "share" into it's own field.

x110dc (approved):

PR #797, opened 2022-10-31
DEV-2519: Add Tax Id field to onboarding

GuiMend:

n/a

#### What's this PR do?
Add Tax Id field on onboarding modal + backend to support change

Renamed migration files in apps/organizations/migrations to solve migration conflicts that appeared during review of this PR

To be able to save the user's company tax id

- Create new user
- Verify email
- Check that onboarding modal has EIN field (tax id) and that it only accepts number
- Fill in info and use more than 9 numbers in EIN field (error should show EIN related error message)
- Save: should save user info without error and continue onboarding steps (stripe modal)

Tax Id is being saved:
image

Yes

Yes

Tax Id now is a field on the onboarding

No

No

https://news-revenue-hub.atlassian.net/browse/DEV-2519 [Frontend UI]
https://news-revenue-hub.atlassian.net/browse/DEV-2733 [Backend]

Migration (?)

No


apps/organizations/migrations/0009_revenueprogram_tax_id.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ("organizations", "0008_DEV-2178_refactor_plan_modeling_and_migrate_legacy_clients"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="revenueprogram",
+            name="tax_id",
+            field=models.PositiveIntegerField(
+                blank=True, null=True, validators=[django.core.validators.MaxValueValidator(999999999)]
```
nrh-cklimas, line 18: My Django knowledge is very low (and I'm also not an EIN/TIN expert either), but is there a way we can also validate that the number >= 111111111? e.g. it is a 9-digit, positive number?

spa/src/components/account/Profile/Profile.test.tsx
```diff
@@ -21,13 +21,13 @@ function tree() {
...
   const axiosMock = new MockAdapter(Axios);
-  let historyPushMock;
+  let historyPushMock: jest.Mock;
 
   beforeEach(() => {
     axiosMock.onPatch(`users/mock-user-id/${CUSTOMIZE_ACCOUNT_ENDPOINT}`).reply(204);
     historyPushMock = jest.fn();
-    useHistory.mockReturnValue({ push: historyPushMock });
-    useUser.mockReturnValue({ loading: false, refetch: jest.fn(), user: { id: 'mock-user-id' } });
+    (useHistory as jest.Mock).mockReturnValue({ push: historyPushMock });
```
nrh-cklimas, line 29: Just a style thing--I like doing `const useHistoryMock = useHistory as jest.Mock;` at the top of tests so I don't have to cast every time I use it. But I think it's used infrequently enough here to be OK.

spa/src/components/account/Profile/ProfileForm.styled.ts
```diff
@@ -1,6 +1,18 @@
+import { makeStyles } from '@material-ui/core/styles';
 import styled from 'styled-components';
 import InfoTooltip from './InfoTooltip';
 
+export const useStyles = makeStyles(() => ({
```
nrh-cklimas, line 5: Is this leftover from dev? I don't see it imported in ProfileForm. If I missed where it's used, is this change not possible with styled-components? I'd prefer to keep us on that only instead of mixing JSS and styled-components.

spa/src/components/account/Profile/ProfileForm.tsx
```diff
@@ -1,29 +1,44 @@
...
+  // companyTaxStatus needs to be "." so that it shows the default label "Select your status".
+  // Using empty string didn't work
+  companyTaxStatus: '.',
+  firstName: '',
+  jobTitle: '',
+  lastName: '',
+  taxId: ''
+};
+
+export type FormDataType = typeof defaultValues;
```
nrh-cklimas, line 24: Could we call this something like `ProfileFormFields` so that if you import it into some other module, the name is clear?

spa/src/components/account/Profile/__mocks__/ProfileForm.tsx
```diff
@@ -3,11 +3,17 @@ const mockFormData = {
...
 
-const ProfileForm = ({ disabled, onProfileSubmit }) => {
-  function handleSubmit(event) {
+type ProfileFormType = {
+  disabled: boolean;
+  onProfileSubmit: (form: typeof mockFormData) => void;
+};
+
+const ProfileForm = ({ disabled, onProfileSubmit }: ProfileFormType) => {
+  function handleSubmit(event: any) {
```
nrh-cklimas, line 16:
```suggestion
  function handleSubmit(event: FormEvent) {
```
This would need a `import { FormEvent } from 'react';` at the top too.

apps/organizations/models.py
```diff
@@ -215,6 +216,7 @@ class RevenueProgram(IndexedTimeStampedModel):
     )
     # TODO: [DEV-2403] non_profit should probably be moved to the payment provider?
     non_profit = models.BooleanField(default=True, verbose_name="Non-profit?")
+    tax_id = models.PositiveIntegerField(blank=True, null=True, validators=[MaxValueValidator(999999999)])
```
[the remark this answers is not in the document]
benjaminEwhite: Made this comment below, but circling back up here: think we should be validating that the tax ID is exactly 9 digits long. In that case, probably need a custom validator.

spa/src/components/account/Profile/ProfileForm.tsx
```diff
@@ -81,6 +96,24 @@ function ProfileForm({ disabled: disabledProp, onProfileSubmit }) {
           title="Your tax status determines the contribution fees charged through Stripe."
         />
       </TaxStatusContainer>
+      <Controller
+        name="taxId"
+        control={control}
+        render={({ field }) => (
+          <TextField
```
nrh-cklimas, line 103: This field doesn't have a max length on it (I think `inputProps={{ maxLength: 9 }}` would do it, but I didn't test this) and the error message from Django if you enter too many digits isn't great ("A valid integer is required").

It would be nice if the field filtered out dashes as typed instead of showing a validation error message when I click to submit.

spa/src/components/account/Profile/Profile.tsx
```diff
@@ -46,15 +47,16 @@ function Profile() {
       } else {
         dispatch({ type: FETCH_FAILURE, payload: data });
       }
-    } catch (e) {
+    } catch (e: any) {
```
benjaminEwhite, line 50: Could this be `(e: Error)`?

spa/src/components/account/Profile/ProfileForm.tsx
```diff
@@ -1,29 +1,44 @@
...
+};
+
+export type FormDataType = typeof defaultValues;
+export interface ProfileFormProps extends InferProps<typeof ProfileFormPropTypes> {
+  onProfileSubmit: (form: FormDataType) => void;
+}
+
+function ProfileForm({ disabled: disabledProp, onProfileSubmit }: ProfileFormProps) {
+  const classes = useStyles();
+  const { control, handleSubmit, watch } = useForm({ defaultValues });
```
benjaminEwhite, line 31: We should consider using Yup + RHF for schema validation on form inputs. No need for round-trip to server to find out that the value for tax ID is invalid (too many or not enough digits, for instance).

https://react-hook-form.com/get-started#schemavalidation

spa/src/components/account/Profile/ProfileForm.tsx
benjaminEwhite, line 103: Ideally, we'd do some sort of masking on this input such that as user types in numbers, the dash gets added after first two digits.

benjaminEwhite (changes requested):

spa/src/components/account/Profile/Profile.tsx
GuiMend: Typescript blocks typing in catch statements to `any` or `unknown`

apps/organizations/migrations/0009_revenueprogram_tax_id.py
GuiMend: Updated the field to have a better validator

apps/users/tests/test_views.py
```diff
@@ -783,6 +784,16 @@ def test_customize_account_sets_revenue_program_status_non_profit(self):
         revenue_program = RevenueProgram.objects.get(name=self.customize_account_request["organization_name"])
         assert revenue_program.non_profit
 
+    def test_customize_account_sets_revenue_program_tax_id(self):
```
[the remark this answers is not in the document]
GuiMend: Not sure how to do this one. Could you point me in the right direction?

spa/src/components/account/Profile/ProfileForm.tsx
GuiMend: I used the validation rules from react-hook-forms itself. Don't think it is necessary to have yup for this instance.

spa/src/components/account/Profile/ProfileForm.styled.ts
nrh-cklimas: I see. Can we replace with styled components, then?

spa/src/components/account/Profile/ProfileForm.tsx
nrh-cklimas: Masked input looks great now.

apps/users/tests/test_views.py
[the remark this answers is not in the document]
benjaminEwhite: See https://github.com/newsrevenuehub/rev-engine/compare/DEV-2519-capture-tax-id-in-onboarding...DEV-2519-capture-tax-id-in-onboarding-bw-suggestions

nrh-cklimas (approved):

benjaminEwhite (approved):

spa/src/components/account/Profile/Profile.tsx
```diff
@@ -30,14 +30,14 @@ function Profile() {
     dispatch({ type: FETCH_START });
 
     try {
-      const { data, status } = await axios.patch(`users/${user.id}/${CUSTOMIZE_ACCOUNT_ENDPOINT}`, {
+      const { data, status } = await axios.patch(`users/${user?.id}/${CUSTOMIZE_ACCOUNT_ENDPOINT}`, {
```
nrh-cklimas, line 33: Under what circumstances would `user` not be defined when this function gets called? I think, at a glance, that this is a "should never happen" scenario. If so, I would rather this throw an error saying so rather than the API request failing because it'll try to PATCH `users/undefined/customize_account/`.

nrh-cklimas (changes requested):

spa/src/components/account/Profile/Profile.tsx
GuiMend: @nrh-cklimas , you are correct. This scenario will never happen, because for the user to get to this place the `user` object will have an `id`. But TS makes me add the `?` because the typing has undefined as a possibility 
image

Do you want me to create something like:
```js
if (user.id) {
   // patch
} else {
  alert.error("User is undefined")
}
```

nrh-cklimas: Since this should never happen if things are working right, I'm thinking

```tsx
if (!user) {
  // Should never happen under normal circumstances.
  throw new Error('Asked to customize user, but user is undefined');
}

The user will see a "something went wrong" message and Sentry will log the problem.

nrh-cklimas (approved):

x110dc (approved):

benjaminEwhite:
Closing and reopening this PR in order to trigger a fresh review app build (see JIRA ticket for more context).

PR #823, opened 2022-11-10
Add SPA coverage report to PRs, fail if below 30%

nrh-cklimas:

n/a

#### What's this PR do?

Adds a comment to PRs reporting Jest coverage and fails the check if it goes below 30%.

Helps us improve our test coverage.

Look at coverage comment in this PR; I tested the coverage threshold in https://github.com/newsrevenuehub/rev-engine/pull/805 but I can show it failing in this PR too if need be.

No.

No, it's CI only.

n/a

Not sure the way I implemented this was current best practice. There might be better ways to do it in GitHub Actions. I tried a variety of prepackaged actions and couldn't get them to work.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2839

No.

No.


GuiMend (approved):
👍

spa/package.json
```diff
@@ -90,6 +90,11 @@
       "!**/*.stories.js",
       "!**/*.styled.js"
     ],
+    "coverageThreshold": {
+      "global": {
+        "lines": 30
```
benjaminEwhite, line 95: Given that we're already at ~38%, should we consider setting that as minimum, with a view of incrementally increasing over time?

Also, wondering how we should be thinking about statement and branch coverage vs line coverage. When i see this:

### Jest unit test coverage
| Lines                | Statements                |                 Branches |
|----------------------|---------------------------|--------------------------|
| 38.06% | 37.17% | 31.98% |

...it makes me wonder if it's good/normal/bad that line coverage is higher than statements/branches. 

If line coverage is the one that is actionable and matters, might make sense to only add that metric in the PR comment that gets attached (but that's not a strongly held opinion).

benjaminEwhite (commented):
Overall looks good, just had a couple of questions.

spa/package.json
nrh-cklimas: I picked 30% as a nice round number but I guess we could start with 35% instead? Don't feel strongly about it.

Lines are literal lines of code, I think, whereas statements are, well, statements. e.g.

```
callAFunction(
  true,
  false
);
```

Is 4 lines but 1 statement. The branches metric measures conditional statements specifically, so it's a subset of statements. I put them in without thinking too hard about it--I don't have strong opinions here either.

benjaminEwhite: Oh — one more question: is there some nifty command we can run to get a coverage report about which lines of code are not covered? We can do that with our Python coverage reports, which is really helpful for when you need to increase test coverage because you're below the threshold (which sometimes happens in unexpected ways after you merge in upstream changes...).

nrh-cklimas: 👍 updated the threshold in 223067149d2a55741d532f23cc5ee355cb87fa7d.

To get a better coverage report, you want `npm run test:coverage && open coverage/lcov-report/index.html`. It'd be awesome to include this level of detail in the PR comment, but when I tried this in the past, it runs afoul of the limit on comment length.

benjaminEwhite: One final request (and no need for me to re-review, as I'm going to go ahead and approve), but would you mind adding something to the `README.md` about that command?  Looks like we don't have an entry yet on Jest tests, so could add minimal section (maybe right before the Cypress section) that just describes command to run Jest tests, and then describes how to get coverage report.

benjaminEwhite (approved):
Made a final request to add an entry in readme about getting test coverage report, but no need for me to re-review that @nrh-cklimas and you can resolve that conversation on your own after adding.  

Approved...

x110dc (approved):

PR #829, opened 2022-11-14
DEV-2861 | Fix integrity error bug for sidebar elements

benjaminEwhite:

#### What's this PR do?

This fixes a bug that caused an integrity error when a user edited page elements on an existing page that already had an Image element in sidebar elements.

There is a [long explanatory comment in the code](https://github.com/newsrevenuehub/rev-engine/pull/829/files#diff-44087ac156d7cc64c8aaf6cf545f87cf3de6289c2cafdb02f7047ed160da226bR176-R197) that provides context for the cause of the bug.

In addition to fixing the bug, this PR adds a test for the specific edge case that caused the bug. If you keep that test in place but revert the change to the view code, you will get a test failure with the integrity error we saw "in the wild".

Makes page editor functional.

1. Edit a page, adding an image to sidebar elements and save.
2. Edit the page a second time, but this time edit a page element (say a rich text block or pay fees, or similar). Save.
3. There should be no error.

For comparison's sake, if you follow these same steps in staging or test ahead of merging this PR, after step 2 you should see an error in sentry / the dev tools console, and the page will not be updated.

No.

Yes.

N/A

There's probably a better way to be handling file uploads vis-a-vis our page serializer, but that will require some digging. I don't like that we have to look at both raw request data and converted request data (i.e., data that has been validated and converted to internal representation via the serializer), but there are oddities around file upload + nested serialization in DRF.

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-2861

No

N/A

### DIFF
--- a/apps/pages/tests/test_viewsets.py
+++ b/apps/pages/tests/test_viewsets.py
@@ -3,6 +3,7 @@ import json
 import os
 
 from django.conf import settings
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.utils import timezone
 
 from rest_framework import status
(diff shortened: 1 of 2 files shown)


x110dc (approved):
LGTM.

PR #830, opened 2022-11-14
DEV-2679: Replace react-select with MUI

nrh-cklimas:

n/a

#### What's this PR do?

Removes react-select in favor of MUI Autocomplete. Because the API of the component is different, some changes to the donation page logic were needed. tl;dr the country cannot be serialized from the form in the DOM and state needs to be tracked separately.

This also removes the ability for people to donate when setting their location as either Micronesia and the United States Outlying Islands, as they don't appear to have a FIPS code, which Stripe requires. The option never worked; if someone tried to make a contribution from these places, they would have seen an error message.

Smaller bundle size and removes a CSP warning we were getting.

1. Make a test contribution. Confirm that the country select looks and acts identically to one deployed to another tier, like dev.
2. Confirm that the test contribution's country is recorded correctly in Stripe.
3. Try making another test contribution without choosing a country, and confirm that the form will not let you progress, and that there is a validation error under the country field.
4. Try making another test contribution and allowing the Google Maps API to autocomplete the address being entered. Confirm that the correct country is selected, and that the country is recorded correctly in Stripe after making the contribution.

Maybe we want to document the countries we aren't allowing?

Yes.

n/a

I think we want to move the contribution form logic to react-form-hook but we're not ready to do so yet. There's still a mix of DOM interactions and React state being serialized.

n/a

https://news-revenue-hub.atlassian.net/browse/DEV-2679

No.

No.


spa/src/components/base/Select/SearchableSelect.test.tsx
```diff
@@ -26,6 +22,14 @@ describe('SearchableSelect', () => {
     expect(screen.getByRole('combobox')).toBeInTheDocument();
   });
 
+  it('renders options after the open button is clicked', () => {
```
benjaminEwhite, line 25: Isn't this just testing Mui Select?  — In which case maybe don't need test.

On other hand, if we are testing this, seems like would also want test that the input value resolves to the chosen option after selecting.

nrh-cklimas: If I remember right, I added this because halfway through implementation, I made a mistake with the component that caused it to never open. I'd rather go overboard with tests right now so I added what you suggested to 917ae37ed19981c363fc21b196a8bea8536086ca.

spa/cypress/e2e/01-donationPage.cy.js
```diff
@@ -344,7 +344,8 @@ function fillOutAddressSection() {
   cy.get('[data-testid*="mailing_city"]').type('Big City');
   cy.get('[data-testid*="mailing_state"]').type('NY');
   cy.get('[data-testid*="mailing_postal_code"]').type('100738');
-  cy.get('.country-select').click().find('.react-select__option').first().click();
+  cy.get('.country-select').click();
+  cy.get('[class*="MuiAutocomplete-popper"] li[data-option-index="0"]').click();
```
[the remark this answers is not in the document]
nrh-cklimas: Like this a lot better than what I came up with! 6253984dc19ac0952a5d671d835e16a05bbc09be

benjaminEwhite (approved):

spa/src/components/base/Select/SearchableSelect.stories.tsx
```diff
@@ -7,7 +7,12 @@ export default {
 } as ComponentMeta<typeof SearchableSelect>;
 
 const Template: ComponentStory<typeof SearchableSelect> = (args: any) => (
-  <SearchableSelect options={[{ label: 'Red' }, { label: 'Green' }, { label: 'Blue' }]} {...args} />
+  <SearchableSelect
+    label="Color"
+    getOptionLabel={({ label }: { label: string }) => label}
+    options={[{ label: 'Red' }, { label: 'Green' }, { label: 'Blue' }]}
```
GuiMend, line 13: If we add the `value` prop in the options, I think we will be able to select the option in the story. (Currently if you click an option, nothing happens)

spa/src/components/base/TextField/TextField.tsx
```diff
@@ -5,10 +5,10 @@ import styled from 'styled-components';
 
 const StyledMuiTextField = styled(MuiTextField)`
   && {
-    .MuiFormHelperText-root {
+    [class*='MuiFormHelperText-root'] {
```
GuiMend, line 8: Why make this change? Is there something wrong with one approach?

spa/src/components/base/Select/CountrySelect.tsx
```diff
@@ -1,12 +1,37 @@
...
+}
+
+export interface CountrySelectProps extends Omit<SearchableSelectProps<CountryOption>, 'options' | 'value'> {
+  value: string;
+}
+
+const options: CountryOption[] = countryCodes.countries
+  .filter(({ fips }) => fips !== '')
+  .map(({ fips, country }) => ({ label: country, fipsCode: fips }))
   .sort((a, b) => a.label.localeCompare(b.label));
```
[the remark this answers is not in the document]
nrh-cklimas: I poked around in Lodash source code and it appears it uses `>` and `<` to compare strings, not `localeCompare` as here. I think an advantage potentially here is that `localeCompare` may handle accents better? Tagging @benjaminEwhite since he added this initially.

spa/src/components/base/Select/CountrySelect.tsx
```diff
@@ -1,12 +1,37 @@
...
+  const { value: fipsValue, ...rest } = props;
+  const selected = useMemo(() => options.find(({ fipsCode }) => fipsValue === fipsCode), [fipsValue]);
+
+  // The cast of value to any is because we need to provide a non-undefined
+  // value to keep the component controlled, but the type definition doesn't
+  // allow this.
+
+  return (
+    <SearchableSelect
+      getOptionLabel={({ label }: CountryOption) => label}
```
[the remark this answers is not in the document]
nrh-cklimas: Not sure. I can omit this prop and TypeScript doesn't complain, but the Storybook story errors out when I click into the field to begin typing.

spa/src/components/base/Select/SearchableSelect.stories.tsx
nrh-cklimas: I'm confused. For me, this story keeps the selected value when I click on it. Same thing happens for me in [Chromatic](https://www.chromatic.com/component?appId=63654636d78401bbd268c64f&csfId=base-select-searchableselect&buildNumber=206&k=6373dc702eef353e987f62e2-1200-interactive-true&h=9&b=-1) (not sure if this link will work).

spa/src/components/base/TextField/TextField.tsx
nrh-cklimas: I could use your advice on this. The problem I've been seeing is that when I test in the app (I was using a contribution page in this instance), the nested class names change from `MuiFormHelperText-root` to something like `MuiFormHelperText-root-8`, where the number is unique in the page.

I haven't been able to find any docs on how to handle this case (best I could find was [this](https://v4.mui.com/guides/interoperability/#deeper-elements-2)), and I don't really like these loose selectors. I was thinking possibly there's a way to connect the generated class names from styled components to the `classNames` prop?

spa/src/components/base/Select/CountrySelect.test.tsx
```diff
@@ -31,14 +25,23 @@ describe('CountrySelect', () => {
...
-      ({ textContent }) => textContent
-    );
+    const options = Array.from(screen.getAllByRole('option')).map(({ textContent }) => textContent);
 
     expect(options).toEqual(['AAA', 'BBB', 'CCC']);
   });
 
+  it('selects the correct option based on the value prop', () => {
+    tree({ value: 'aa' });
+    expect(screen.getByRole('textbox')).toHaveValue('BBB'); // because it uses FIPS code as value, not country name
```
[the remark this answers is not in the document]
nrh-cklimas: I agree, but I'm not sure how to improve it. Do you have any suggestions?

spa/src/components/base/Select/SearchableSelect.stories.tsx
GuiMend: It's working for me too. Not sure what I saw different before... 🤔

spa/src/components/base/TextField/TextField.tsx
GuiMend: Yes, that happens. I believe it's something like css modules, where every compilation it returns an unique classname (in this case, `-8`). I wouldn't say it's loose selectors, as Mui applies the `.Mui[component]-[variation]` to the matching component > variant, and when compiled it returns an unique classname. 

It's kind of the way Mui works, I don't see it as a problem. In my opinion I would rather keep using it in the way of `.MuiFormHelperText-root` than `[class*='MuiFormHelperText-root']`, as in the documentation it's the suggested/recommended way of overwriting Mui styles. 

I'm not sure if this way of using `class*=` can somehow overwrite unwanted classes if we have nested components that use the same classnames.

spa/src/components/base/Select/CountrySelect.tsx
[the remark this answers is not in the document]
GuiMend: As it works with this we can leave it. At worst it's just redundant

spa/src/components/base/Select/CountrySelect.tsx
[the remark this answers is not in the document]
GuiMend: I'm ok either way, what you guys decide is good for me.

spa/src/components/base/Select/CountrySelect.test.tsx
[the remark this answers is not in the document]
GuiMend: Maybe just adding more comments?
```suggestion
    tree({ value: 'aa' }); // selectec value for Mui Autocomplete (uses FIPS code)
    expect(screen.getByRole('textbox')).toHaveValue('BBB'); // label that shows when selected (country name)
```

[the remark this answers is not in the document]
nrh-cklimas: Makes sense to me! a93942dc6a48b032fde8d3364f7a644608728b0a

spa/src/components/base/TextField/TextField.tsx
nrh-cklimas: Right, but the issue is that when the components are rendered with these scoped class names, the selectors no longer match.

Here's what the country select looks like as the PR is right now:

Screen Shot 2022-11-16 at 2 39 41 PM

There's a bunch of classes in play here, but an example is `MuiAutocomplete-inputRoot-177`, which this messy selector hits:

`[class*='MuiAutocomplete-inputRoot'][class*='MuiInput-root'] [class*='MuiAutocomplete-input']:first-child`

But if I change the selector to:

`.MuiAutocomplete-inputRoot .MuiInput-root .MuiAutocomplete-input:first-child`

It no longer applies to the DOM because of the class scoping. It's a little hard to see in the screenshot, but the height of the control is now wrong.

Screen Shot 2022-11-16 at 2 40 31 PM

So... I'd really like to use plain class selectors, but they don't seem to work for nested classes. I'm overall confused because this seems to be the approach the MUI docs suggest.

GuiMend: Why do we need all the `.MuiAutocomplete-inputRoot.MuiInput-root .MuiAutocomplete-input:first-child`, can't it only be `.MuiAutocomplete-input:first-child` or something simpler?

GuiMend: The style is not applying probably because you need to add that className into an inner component.
Ex: adding the style to `InputProps` instead of the parent component

spa/src/components/base/Select/CountrySelect.tsx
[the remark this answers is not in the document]
benjaminEwhite: @nrh-cklimas you're right about use of locale compare (i.e., deal with accents).

spa/src/components/base/TextField/TextField.tsx
nrh-cklimas: Thanks for all the help/feedback with this--7b63879225392908e2ceb78a164892b7b49cc492

GuiMend (approved):
👍

x110dc (approved):

PR #841, opened 2022-11-17
DEV-2750 | Stripe onboarding modal transition is delayed

benjaminEwhite:

#### What's this PR do?

This PR makes a minor change such that `Dashboard`'s displays the `GlobalLoading` component instead of the main body if `useConnectStripeAccount`'s return value for `isLoading` is `true`.  This solves for the issue in the ticket — namely, it ensures that the display of the 'Connect to Stripe' modal and toast is not delayed, and instead displays immediately on load of the main body of `Dashboard` (assuming that the account is indeed in need of Stripe connection).

Better UX for self-onboarded users.

Create a new account [here in the review app](https://dev-2750.revengine-review.org/sign-in/). 

After going through the account customization step, when taken to the `Dashboard`, you should immediately see the "Connect to Stripe" modal (assuming that you don't have a cookie set to hide it). 

As you go through subsequent Stripe connection steps (up until connecion success), the Stripe Connect toast should be immediately displayed. 

If you sign out and sign back in to the dashboard, it should also display immediately.

No.

Yes, it adds a very minimal test that shows that `Dashboard` displays the loading component if `useConnectStripeAccount` is loading the account link status from the server.  This is the first and only unit test for the `Dashboard` component. 

N/A

No.

N/A

[DEV-2750](https://news-revenue-hub.atlassian.net/browse/DEV-2750)

No

No

### DIFF
--- a/spa/src/components/dashboard/Dashboard.test.tsx
+++ b/spa/src/components/dashboard/Dashboard.test.tsx
@@ -1,0 +1,35 @@
+import { render, screen } from 'test-utils';
+
+import Dashboard from './Dashboard';
+import useConnectStripeAccount from 'hooks/useConnectStripeAccount';
+import useFeatureFlags from 'hooks/useFeatureFlags';
+import useUser from 'hooks/useUser';
+
+jest.mock('elements/GlobalLoading');
+jest.mock('hooks/useConnectStripeAccount');
+jest.mock('hooks/useFeatureFlags');
+jest.mock('hooks/useUser');
+
+describe('Dashboard', () => {
+  const useConnectStripeAccountMock = useConnectStripeAccount as jest.Mock;
+  const useFeatureFlagsMock = useFeatureFlags as jest.Mock;
+  const useUserMock = useUser as jest.Mock;
+
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


spa/src/components/dashboard/Dashboard.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+      isLoading: false,
+      isError: false
+    });
+    useUserMock.mockReturnValue({
+      isLoading: false,
+      isError: false,
+      refetch: jest.fn()
+    });
+  });
+  it('shows a loading status when Stripe account link status is loading', () => {
```
nrh-cklimas, line 30: Could we add a test for the opposite, to prove the component doesn't get stuck in loading? I think it could literally be the opposite of this test.

nrh-cklimas (changes requested):

spa/src/components/dashboard/Dashboard.test.tsx
benjaminEwhite: See 0c4e300

nrh-cklimas (approved):

x110dc (approved):

PR #843, opened 2022-11-18
DEV-2864: Add hook for page retrieval

nrh-cklimas:

1. #843 (this PR)
2. https://news-revenue-hub.atlassian.net/browse/DEV-2865
3. https://news-revenue-hub.atlassian.net/browse/DEV-2866
4. https://news-revenue-hub.atlassian.net/browse/DEV-2867 

#### What's this PR do?

This is the first of a few stacked PRs to refactor the data layer of the page editor. This introduces a hook, `useContributionPage`, which manages API interactions for contribution pages, and connects it to the PageEditor component. PageEditor is simplified as a result, but other components in the tree work the same way.

This removes the need to push the page ID into history state when editing a page; now, it always uses the RP and page slug to do the lookup.

Improves test coverage, migrates some code to TypeScript, simplifies routing.

I think we should QA/UAT this once all stacked PRs are merged here because a lot potentially could regress in this work. Briefly...

1. Create a page and confirm that it opens the editor as usual.
2. Edit a page, and confirm that changes in each tab are saved successfully.
3. Publish a page and confirm that it successfully goes live.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2864

No.

See other subtasks in https://news-revenue-hub.atlassian.net/browse/DEV-2765


GuiMend (commented):
While testing the edit functionality I've noticed that when editing, the changes show in the preview as expected, but I'm not able to save those changes. I keep getting an error (but nothing poped up in network or console 🤔)

image

GuiMend (commented):
Also, not sure it's specific from this PR, but the + button in on top of another one 
image

spa/src/components/pageEditor/PageEditor.js
```diff
@@ -70,329 +66,218 @@ const THUMBNAIL_KEYS = ['graphic_thumbnail', 'header_bg_image_thumbnail', 'heade
...
-        setPageContext(data);
-        setLoading(false);
-        handleEdit();
-      },
-      onFailure: handleGetPageFailure //() => setLoading(false)
-    });
-    // Don't include requestGetPage for now.
-  }, [pageId, parameters.revProgramSlug, parameters.pageSlug, handleGetPageFailure]);
+    if (isError && error.response?.data) {
+      alert.error(error.response.data.detail, { timeout: 0 });
```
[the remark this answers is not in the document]
nrh-cklimas: I [ported it over from existing code](https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/pageEditor/PageEditor.js#L133)--it makes the alert stay visible until the user closes it. Added a comment in 790748f20766a28318c40b225090ec71444aade7 to explain it better.

spa/src/components/pageEditor/PageEditor.js
```diff
@@ -70,329 +66,218 @@ const THUMBNAIL_KEYS = ['graphic_thumbnail', 'header_bg_image_thumbnail', 'heade
...
+      throw new Error("Can't delete this page");
+    }
+
+    async function finishDelete() {
+      try {
+        await deletePage();
+        history.push(CONTENT_SLUG);
+      } catch {
+        // Do nothing--the hook does needed error handling.
+      }
```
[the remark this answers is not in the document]
nrh-cklimas: The idea is that the deletePage function the hook gives the component shows an alert if the request fails, so the page doesn't need to do anything else. I added some clarifying comments in 790748f20766a28318c40b225090ec71444aade7 (I hope).

nrh-cklimas:
@GuiMend on your two comments:

- The error on the review app had to do with a problem saving the page screenshot. I added better logging in 36eb60b07ffb98a8914180a3cd125c3bef480913 and fixed the root cause in a301fd22a3065076e8e8a8001bc393561b0f92cd.
- Checked on dev and the + button overlaps the content for me there, so I think it's existing behavior. We're also getting rid of the + button in https://github.com/newsrevenuehub/rev-engine/pull/853, so I think it's not worth fixing here.

GuiMend (changes requested):
Code is looking good 👍 

But I noticed that if I save a page and the save is successfull, when clicking the "back" button on the header, the modal of "unsaved changes" pops up (and it shouldn't)

GuiMend (approved):
THink it's looking good 👍

spa/src/hooks/useUser.types.ts
```diff
@@ -1,31 +1,11 @@
...
-  label: string;
-  page_limit: number;
-  style_limit: number;
-  custom_thank_you_page_enabled: boolean;
-}
 export interface Organization {
   name: string;
   slug: string;
-  plan?: Plan;
+  plan?: EnginePlan;
```
benjaminEwhite, line 8: I know the optionality here is pre-existing, but I don't believe there's a possibility of orgs without plans, as orgs get created with a default value for plan (the free one).

spa/src/utilities/editPageGetSuccessMessage.test.ts
```diff
@@ -0,0 +1,82 @@
...
+const twoDaysAgo = add(now, { days: -2 });
+const yesterday = add(now, { days: -1 });
+
+describe('pageIsPublished', () => {
+  describe('When a date to compare is not passed', () => {
+    it('returns false if the page has no published_date', () => expect(pageIsPublished({})).toBe(false));
+
+    it('returns false if the page has a published_date after the current date/time', () => {
+      expect(pageIsPublished({ published_date: formatISO(oneMinuteLater) })).toBe(false);
+      expect(pageIsPublished({ published_date: formatISO(tomorrow) })).toBe(false);
```
benjaminEwhite, line 17: Why both expectations? Seems like the behavior is sufficiently proven just by the first expect.

spa/src/utilities/editPageGetSuccessMessage.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const now = new Date();
+const oneMinuteEarlier = add(now, { minutes: -1 });
+const oneMinuteLater = add(now, { minutes: 1 });
+const tomorrow = add(now, { days: 1 });
+const twoDaysAgo = add(now, { days: -2 });
+const yesterday = add(now, { days: -1 });
+
+describe('pageIsPublished', () => {
+  describe('When a date to compare is not passed', () => {
```
benjaminEwhite, line 12: This is kind of pedantic, but for this test block, to make test rock solid, could mock the default value returned by `new Date()`. According to something I saw on SO, there's this:

```
jest.setSystemTime(new Date(2020, 3, 1));
```

Probably would never happen, but if for some odd reason these tests ran really slowly, they could fail because the "relative now" for oneMinuteLater etc. defined as constant at top of module.

spa/src/hooks/useContributionPage/useContributionPage.ts
```diff
@@ -0,0 +1,92 @@
...
+  // slug.
+
+  const { data } = await axios.get<ContributionPage>(DRAFT_PAGE_DETAIL, {
+    params: {
+      revenue_program: revenueProgramSlug,
+      page: pageSlug
+    }
+  });
+  return data;
+}
```
benjaminEwhite, line 22: Just curious why going with lookup by slug vs id? Is it with a view of SEO/redirect related concerns?

spa/src/hooks/useContributionPage/useContributionPage.ts
```diff
@@ -0,0 +1,92 @@
...
+    },
+    {
+      onSuccess: () => queryClient.invalidateQueries({ queryKey: ['contributionPage', revenueProgramSlug, pageSlug] })
+    }
+  );
+
+  const updatePage = useCallback(
+    async (data: Partial<ContributionPage>, screenshotBaseName?: string, elementToScreenshot?: HTMLElement) => {
+      // No error handling here to allow callers to implement their own
+      // logic--they may not want an error notification at all.
```
benjaminEwhite, line 71: Why is delete mutation opinionated about logging error and alerting, while update is laissez-faire?

spa/src/hooks/useContributionPage/useContributionPage.ts
```diff
@@ -0,0 +1,92 @@
...
+    error,
+    isError,
+    isLoading,
+    refetch
+  } = useQuery(['contributionPage', revenueProgramSlug, pageSlug], () => fetchPage(revenueProgramSlug, pageSlug));
+
+  const deletePageMutation = useMutation(() => {
+    if (!page) {
+      throw new Error('Page is not yet defined');
+    }
```
benjaminEwhite, line 38: Can this even happen, given that deletePage is only returned if page exists?

spa/src/hooks/useContributionPage/useContributionPage.ts
```diff
@@ -0,0 +1,92 @@
...
+  const { data } = await axios.get<ContributionPage>(DRAFT_PAGE_DETAIL, {
+    params: {
+      revenue_program: revenueProgramSlug,
+      page: pageSlug
+    }
+  });
+  return data;
+}
+
+export function useContributionPage(revenueProgramSlug: string, pageSlug: string) {
```
benjaminEwhite, line 24: Also curious why no guarantee on type of return value from this function (mainly because I know for `useConnectStripeAccount` we're explicitly defining that).

spa/src/hooks/useContributionPage/useContributionPage.test.ts
```diff
@@ -0,0 +1,251 @@
...
+        params: { page: 'mock-page-slug', revenue_program: 'mock-rp-slug' },
+        url: 'pages/draft-detail/'
+      })
+    ]);
+    expect(result.current.page).toEqual(mockPage);
+  });
+
+  describe('While fetching the page', () => {
+    // These wait for the next update to allow component updates to happen after
+    // the fetch completes.
```
benjaminEwhite, line 45: I don't understand the waitForNextUpdate usage. I commented out each of the invocations of this function locally, and tests still passed. What's the benefit of using?

benjaminEwhite (changes requested):
stacked pr in plan is helpful https://github.com/newsrevenuehub/rev-engine/pull/847

spa/src/components/common/Button/GrabLink/GrabLink.js
```diff
@@ -1,20 +1,20 @@
 import { useState } from 'react';
 import PropTypes from 'prop-types';
 
-import { pageHasBeenPublished } from 'utilities/editPageGetSuccessMessage';
+import { pageIsPublished } from 'utilities/editPageGetSuccessMessage';
```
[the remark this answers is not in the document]
nrh-cklimas: I don't have a strong opinion on this--I think my thinking on it was to try to keep the diff on the PR as small as I could manage. I'm tackling the publish modal in a followup story so feels like that might be the place to rethink this approach.

spa/src/components/common/Button/GrabLink/GrabLink.js
```diff
@@ -1,20 +1,20 @@
...
-import { PagePropTypes } from 'constants/proptypes';
+import { PagePropTypes } from 'constants/propTypes';
 
 const GrabLink = ({ page, className }) => {
   const [copied, setCopied] = useState('');
   const [anchorEl, setAnchorEl] = useState(null);
   const open = Boolean(anchorEl);
   const id = open ? 'grab link popover' : undefined;
-  const isPublished = pageHasBeenPublished(page);
+  const isPublished = page && pageIsPublished(page);
```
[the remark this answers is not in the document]
nrh-cklimas: This is definitely a matter of opinion, but I'd prefer to force callers to be explicit about pages being null or undefined. A scenario that comes to mind is page deletion. Right now we have logic that says if a page is published, you have to OK a confirmation dialog to delete it. I think the way pageIsPublished is updated in this PR, TypeScript will flag any calls to it where page may not be defined (e.g. loaded) yet, and prevent a component from deleting a page without confirmation while it's loading.

It's a fairly hypothetical scenario though...

spa/src/components/pageEditor/ElementErrors.tsx
```diff
@@ -0,0 +1,35 @@
...
+  if (errors.length === 0) {
+    return null;
+  }
+
+  return (
+    <>
+      The following elements are required for your page to function properly:
+      <ul data-testid="missing-elements-alert">
+        {errors.map((error) => (
+          <li key={error.element}>{error.message}</li>
```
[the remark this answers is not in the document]
nrh-cklimas: I'll plead [copy/paste](https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/pageEditor/PageEditor.js#L458). I'm fuzzy on what error responses look like. It sounds like the safer bet is to key on index instead?

spa/src/hooks/useContributionPage/useContributionPage.ts
nrh-cklimas: This is to please TypeScript--it's not smart enough to realize that.

spa/src/hooks/useContributionPage/useContributionPage.test.ts
nrh-cklimas: I get these warnings in test logs if I take them out:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

Do you? I think the issue is that if I don't, the mocked fetch resolves after the test does, and the hook tries to update but the DOM it was mounted to is gone.

spa/src/components/common/Modal/AddPageModal/AddPageModal.tsx
```diff
@@ -26,7 +26,7 @@ const AddPageModal = ({ open, onClose, loading, revenuePrograms, onAddPage, oute
   });
 
   const onSubmit = (form: typeof formDefaultValues) => {
-    onAddPage(form.revenueProgram);
+    onAddPage(parseInt(form.revenueProgram));
```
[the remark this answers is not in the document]
nrh-cklimas: I think this is an artifact of us thinking RP IDs were strings in some parts of our code (which I think I mostly caught in this PR). It weirds me out a bit to use `NaN` as a default, how about something like this instead? Bonus: forces us to do more error checking instead of sending bad API requests if there's a bug in our form logic.

```tsx
const formDefaultValues: { revenueProgram: number | null } = {
  revenueProgram: null
};

const AddPageModal = ({ open, onClose, loading, revenuePrograms, onAddPage, outerError }: AddPageModalProps) => {
  const {
    control,
    handleSubmit,
    formState: { errors }
  } = useForm({
    defaultValues: formDefaultValues
  });

  const onSubmit = (form: typeof formDefaultValues) => {
    if (!form.revenueProgram) {
      throw new Error('Revenue program has not been set');
    }

    onAddPage(form.revenueProgram);
  };
```

spa/src/hooks/useContributionPage/useContributionPage.ts
nrh-cklimas: This is a really good question and I think I need to revisit it in the next stacked PR. The driver was our routing--obviously when people edit a page, the URL looks like `/edit/rpSlug/pageSlug`. I wanted to get rid of having to stuff the page ID into browser history and just be able to use the route params.

What I'm finding is, doing some prelim work on the publish modal, we probably need to also allow look up by ID, because the publish modal allows the user to change the page slug which makes the hook a lot harder to reason about. I'm tentatively thinking about giving the hook two signatures to allow usage in either case, though that does mean that it will cache less than it could--it will have a cache entry for the page by its slug, and one by its ID (potentially). I don't think having two separate hooks makes sense.

What do you think?

spa/src/hooks/useContributionPage/useContributionPage.ts
nrh-cklimas: Really good question. I started out following the pattern we've using with other hooks, where a component makes a call to the hook to make an API call and the hook handles errors--the component . But I discovered we're relying on the API to do validation for us, so we have two types of error states with updatePage()...

1. Something our fault--API is unreachable, DB problems, etc
2. The user didn't fill in a form field right

I think the ideal end state is the SPA stops relying on the API to do basic validation for it, so the hook would one day only see exceptional errors. But in the meantime, updating needs nuance around errors. I do generally think hooks should handle error handling for components where possible, so that we don't have to repeat ourselves.

So, I could...

1. Write a longer comment :)
2. Make at least updatePage and deletePage consistent with each other and not do any error handling, but make them inconsistent with other hooks
3. Remove error handling from all our react-query hooks and make consumers handle it
4. Something else?

spa/src/hooks/useContributionPage/pageUpdateToFormData.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Takes image blobs that have been set in elements and moves them into form
+ * data. It doesn't modify the elements.
+ */
+export function putImagesInFormData(elements: ContributionPageElement[], formData: FormData) {
+  for (const element of elements) {
+    if (element.type === 'DImage' && element.content instanceof File) {
+      formData.append(element.uuid, element.content, element.content.name);
+    }
+  }
+}
```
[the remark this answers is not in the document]
nrh-cklimas: They are equivalent, but TS understands the `if` as narrowing the type of `element.content` but isn't smart enough to reason about a filter operation (I believe). The fix would look like this...

```tsx
  elements
    .filter(({ type, content }) => type === 'DImage' && content instanceof File)
    .forEach(({ uuid, content }) => formData.append(uuid, content as File, (content as File).name));
```

I have a slight pref for the original because it doesn't require a manual cast, which means better typechecking, but it's not a strongly-held opinion.

spa/src/hooks/useContributionPage/useContributionPage.types.ts
```diff
@@ -0,0 +1,342 @@
...
+  /**
+   * User-visible name of the revenue program.
+   */
+  name: string;
+  /**
+   * Has the revenue program been set up with Stripe and verified by them?
+   */
+  payment_provider_stripe_verified: boolean;
+  /**
+   * Slug for the revenue program used in URLs.
```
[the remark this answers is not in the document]
nrh-cklimas: Updated in 8b6ab65eb452720b80a0335737f2c74d4a155cad

spa/src/hooks/useContributionPage/useContributionPage.types.ts
```diff
@@ -0,0 +1,342 @@
...
+   */
+  id: number;
+  /**
+   * Account ID of the Stripe account.
+   */
+  stripe_account_id: string | null;
+
+  // Unsure of what these two attributes are for.
+  stripe_oauth_refresh_token: string;
+  stripe_product_id: string | null;
```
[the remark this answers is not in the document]
nrh-cklimas: Updated in 8b6ab65eb452720b80a0335737f2c74d4a155cad

spa/src/utilities/editPageGetSuccessMessage.test.ts
nrh-cklimas: :+1: 0c4703b163f4f95e09e5ec7e178fd22ac23495cd

spa/src/utilities/editPageGetSuccessMessage.test.ts
nrh-cklimas: I was iterating on this test--originally did it with just one day later then thought, "need to test that it doesn't look at just the date". Simplified in 0c4703b163f4f95e09e5ec7e178fd22ac23495cd

spa/src/hooks/useUser.types.ts
nrh-cklimas: Updated in 29e95f3147c5de747acb166ce9e2166e19dc2bea

spa/src/hooks/useContributionPage/useContributionPage.ts
nrh-cklimas: Just being lazy :)

27035eb60d00788a47c274912d0bfc4415091c7f

spa/src/components/common/Modal/AddPageModal/AddPageModal.tsx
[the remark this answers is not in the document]
benjaminEwhite: Yeah, I like this and support this change.

spa/src/hooks/useContributionPage/useContributionPage.ts
benjaminEwhite: That makes sense, and seems fine to revisit later.

spa/src/hooks/useContributionPage/useContributionPage.ts
benjaminEwhite: As a fan of incrementalism, IMO longer comment seems like the way to go.

spa/src/hooks/useContributionPage/useContributionPage.test.ts
benjaminEwhite: Not sure if I got those logs or not, but reasoning makes sense. Maybe just add a comment?

benjaminEwhite (approved):
@nrh-cklimas changes look great. I resolved most of the conversations except the ones where there's a next step for you. 

I'm going ahead and approving with assumption you'll handle those loose ends and can then resolve conversations on your own. No need for re-review from me.

spa/src/components/common/Modal/AddPageModal/AddPageModal.tsx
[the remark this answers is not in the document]
nrh-cklimas: Argh, I take it back! MUI complains in unit tests:

```
console.warn
      Material-UI: You have provided an out-of-range value `null` for the select (name="revenueProgram") component.
      Consider providing a value that matches one of the available options or ''.
      The available values are `1`, `2`, `3`.

      at node_modules/@material-ui/core/Select/SelectInput.js:366:17
      at invokePassiveEffectCreate (node_modules/react-dom/cjs/react-dom.development.js:23487:20)
      at HTMLUnknownElement.callCallback (node_modules/react-dom/cjs/react-dom.development.js:3945:14)
      at HTMLUnknownElement.callTheUserObjectsOperation (node_modules/jsdom/lib/jsdom/living/generated/EventListener.js:26:30)
      at innerInvokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:338:25)
      at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:274:3)
      at HTMLUnknownElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:221:9)
      at HTMLUnknownElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:94:17)
```

I think we have to leave it as-is.

spa/src/hooks/useContributionPage/useContributionPage.test.ts
nrh-cklimas: 9688091f12ac5828bed1f6263413acdde364cd56

spa/src/hooks/useContributionPage/useContributionPage.ts
nrh-cklimas: 9688091f12ac5828bed1f6263413acdde364cd56

spa/src/hooks/useContributionPage/useContributionPage.types.ts
```diff
@@ -0,0 +1,342 @@
...
+  revenue_program: RevenueProgram;
+  /**
+   * FIPS code of the country that the revenue program belongs to.
+   */
+  revenue_program_country: string;
+  /**
+   * Is the revenue program nonprofit?
+   */
+  revenue_program_is_nonprofit: boolean;
+
```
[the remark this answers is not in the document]
nrh-cklimas: https://news-revenue-hub.atlassian.net/browse/DEV-2953

spa/src/hooks/useContributionPage/useContributionPage.types.ts
```diff
@@ -0,0 +1,342 @@
...
+    symbol: string;
+  };
+  /**
+   * Payment provider configured for this page.
+   */
+  payment_provider: PaymentProvider;
+  /**
+   * Account ID of the Stripe account associated with the page.
+   */
+  stripe_account_id: string;
```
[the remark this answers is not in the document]
nrh-cklimas: https://news-revenue-hub.atlassian.net/browse/DEV-2953

spa/src/hooks/useContributionPage/useContributionPage.types.ts
```diff
@@ -0,0 +1,342 @@
...
+export interface PaymentProvider {
+  /**
+   * Timestamp when the payment provider was created.
+   */
+  created: string;
+  /**
+   * Currency used by the payment provider, e.g. `USD`.
+   */
+  currency: string;
+  default_payment_provider: string;
```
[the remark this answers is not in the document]
nrh-cklimas: https://news-revenue-hub.atlassian.net/browse/DEV-2953

spa/src/hooks/useContributionPage/useContributionPage.types.ts
```diff
@@ -0,0 +1,342 @@
...
+   * Internal ID of the payment provider.
+   */
+  id: number;
+  /**
+   * Account ID of the Stripe account.
+   */
+  stripe_account_id: string | null;
+
+  // Unsure of what these two attributes are for.
+  stripe_oauth_refresh_token: string;
```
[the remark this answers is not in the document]
nrh-cklimas: https://news-revenue-hub.atlassian.net/browse/DEV-2953

PR #870, opened 2022-11-29
DEV-2293: Allow GTM in CSP

nrh-cklimas:

n/a

#### What's this PR do?

Allows Google Tag Manager in our CSP configuration.

We use GTM for tracking, and this gets us closer to being able to turn on CSP enforcement.

1. Go to the donation thank-you page for Billy Penn in the review app.
2. Confirm that there is no CSP-related error in dev tools console, and the network tab in dev tools shows Google Tag Manager loading successfully.

No.

No, but I'd really like to. I'm not sure how to write a test for this kind of change, but it feels like it would be very valuable. I did some Googling around for ideas but didn't have much success.

n/a

Lack of tests

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2293

I don't think so but I am not 100% sure.

No.

### DIFF
--- a/revengine/settings/base.py
+++ b/revengine/settings/base.py
@@ -325,6 +325,7 @@ CSP_SCRIPT_SRC = (
     "'self'",
     "https://js.stripe.com",
     "https://risk.clearbit.com",
+    "https://www.googletagmanager.com",
     "https://www.google-analytics.com",
     "https://maps.googleapis.com",
     "https://www.google.com/recaptcha/",


revengine/settings/base.py
```diff
@@ -325,6 +325,7 @@
     "'self'",
     "https://js.stripe.com",
     "https://risk.clearbit.com",
+    "https://www.googletagmanager.com",
```
benjaminEwhite, line 328: Wild!

benjaminEwhite (approved):
Code change is fine.

As for your comment on testing, i threw together one way we could test, but not sure how valuable it is.  Inside revengine.tests.test_views, we could add a test along these lines

```


def test_csp_headers(client):

    setting_to_header_name = {
        "CSP_DEFAULT_SRC": "default-src",
        "CSP_SCRIPT_SRC": "script-src",
        "CSP_STYLE_SRC": "style-src",
        "CSP_IMG_SRC": "img-src",
        "CSP_FONT_SRC": "font-src",
        "CSP_FRAME_SRC": "frame-src",
        "CSP_CONNECT_SRC": "connect-src",
        "CSP_OBJECT_SRC": "object-src",
    }
    response = client.get("")
    csps = {}
    # CSP header splits rule for each content source by ";".
    # The first item per source is the source name (for instance, "style-src")
    # and then each item for the policy is split by space.
    for x in response.headers["Content-Security-Policy"].split(";"):
        split = [y.strip() for y in x.strip().split(" ")]
        csps[split[0]] = split[1:]
    for k, v in setting_to_header_name.items():
        assert set(getattr(settings, k)) == set(x for x in csps[v] if "nonce" not in x)
    for src in settings.CSP_INCLUDE_NONCE_IN:
        assert next(x for x in csps[src] if x.startswith("'nonce-"))
```

On one hand, this is kind of testing against 3rd party package (django-csp), in which case not super valuable.

On other hand, it's a guard against inadvertently changing the CSP settings in settings, which is possibly valuable.

@x110dc -- any thoughts on this approach to testing?

nrh-cklimas:
@benjaminEwhite (and @x110dc) I don't have a firm idea on this, but the code snippet to get the generated policy seems useful. It would be nice if there was a magical Python library that emulated a browser and validated whether a request to a particular resource would be allowed or not, but I wasn't able to find much on PyPI. I don't think Cypress could be used for this (yet) because it runs directly against the SPA. My gut wants to suggest that this could be a snapshot test, e.g. we want to make sure the policy only changes when we expect it to but completely parsing the policy is unnecessary, but I don't think we have snapshot pytest tests set up.

x110dc (approved):
I agree that we shouldn't need to test much as this is more of a configuration change than a code change and the third party library presumably has its own tests. We could add the test Ben describes to guard against inadvertent changes to the config. Could also test for the presence of something like `https://js.stripe.com` in the CSP headers. 

I'm okay with it as-is, but okay with adding that test(s) if you'd like.

PR #876, opened 2022-11-30
DEV-2897: Limit Chromatic runs based on branch name and code changed

nrh-cklimas:

n/a

#### What's this PR do?

- Blocks Chromatic CI from running on the branches listed in the YAML below. I kept `main` in for now with the idea that it is a baseline for comparison. We can revisit that later if we find ourselves bumping against the snapshot limit.

- Doesn't run Chromatic in CI for PRs that don't modify code in `spa/`.

Faster CI, prevents useless snapshots.

n/a I think since it only runs in CI. You can see Chromatic not running in this PR, but we need to check once merged that Chromatic does run on SPA commits. Alternatively, we could do a test commit on this PR that changes SPA code to show it does the right thing.

No.

No, CI only.

Not needed.

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2897

No.

No.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -13,11 +13,31 @@ on:
       - staging
       - test
 
-# Copied from https://www.chromatic.com/docs/github-actions
-
 # List of jobs
 jobs:
+  # Only run if SPA code has changed
+  check-changes:
+    runs-on: ubuntu-latest
+    # Required permissions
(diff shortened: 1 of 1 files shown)


benjaminEwhite (approved):
Seems fine. I guess we can't really test if it works until this has been merged, right?

x110dc (approved):

PR #895, opened 2022-12-13
DEV-2296: Fix CSP errors with rich text editor

nrh-cklimas:

#### What's this PR do?

Fixes CSP problems with our rich text editor. It sets text alignment and font size using inline styles (e.g. `<p style="text-align:center">`) which our CSP policy blocks. The approach here is to convert these styles to CSS classes which we can style accordingly.

Implementation details:

- Moved the rich text editor to a standalone component since it seems like we'll probably use it outside of the rich text editor block one day.
- Added a new component to display the content created in a rich text editor, applying styles to the CSS classes set by the editor.
- Replaced the `draftjs-to-html` package with `draft-js-export-html` because it didn't have the flexibility needed to replace inline styles with classes.
- Migrated the DRichText component to TypeScript.

Gets us closer to being able to enforce CSP in production.

Testing the code change:

1. Edit a contribution page and put a rich text editor block in the page content if it isn't there already.
2. Put every possible style into the block :-/
  a. Bold
  b. Italics
  c. Underline
  d. Strikethrough
  e. H2
  f. H3
  g. Blockquote
  h. All font size variations :-/: 13, 14, 15,16, 24
  i. Bulleted list (with indent)
  j. Numbered list (with indent)
  k. Left-aligned text
  l. Center-aligned text
  m. Right-aligned text
  n. Links
3. Save changes to the block.
4. Confirm the page preview reflects all style changes correctly.
5. Save changes to the page.
6. Confirm the page as published reflects all these styles too.

🚨 This PR potentially changes how existing pages will display (but shouldn't). I feel pretty strongly we should deploy this to staging and review all our existing contribution pages to make sure they look right.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2296

No.

Although this fixes content going forward, this will not fix CSP violations for existing content. I created https://news-revenue-hub.atlassian.net/browse/DEV-3033 to capture that work.


spa/cypress/e2e/01-donationPage.cy.js
```diff
@@ -11,8 +11,6 @@ import { CONTRIBUTION_INTERVALS } from '../../src/constants/contributionInterval
 import * as freqUtils from 'utilities/parseFrequency';
 import calculateStripeFee from 'utilities/calculateStripeFee';
 import { DEFAULT_BACK_BUTTON_TEXT } from 'components/common/Button/BackButton/BackButton';
-import { CANCEL_PAYMENT_FAILURE_MESSAGE } from 'components/donationPage/DonationPage';
```
nrh-cklimas, line 14: This is here because Cypress was giving Webpack compilation errors on this suite otherwise. I think something in this PR changed these two components such that they now include modules that Cypress doesn't like, but I admit I don't know exactly what.

spa/cypress/e2e/04-donations.cy.js
```diff
@@ -145,11 +145,11 @@ describe('Donations list', () => {
     it('should display the second page of donations when click on next page', () => {
       cy.wait('@getDonations');
       cy.get('li > button[aria-label="Go to page 2"]').click();
-      cy.wait('@getDonations').then((intercept) => {
-        console.log({ intercept });
-        cy.getByTestId('donations-table')
-          .find('tbody tr[data-testid="donation-row"]')
-          .should('have.length', intercept.response.body.results.length);
+      cy.wait('@getDonations');
```
nrh-cklimas, line 148: I changed this because I saw Cypress failures on this test. I don't think this PR affected this part of the app, however.

spa/src/components/common/Modal/AddPageModal/AddPageModal.test.tsx
```diff
@@ -54,7 +54,8 @@ describe('AddPageModal', () => {
     userEvent.click(screen.getByRole('button', { name: /Revenue Program/i }));
     userEvent.click(screen.getByRole('option', { name: revenuePrograms[0].name }));
     userEvent.click(screen.getByRole('button', { name: 'Create' }));
-    await waitFor(() => expect(onAddPage).toHaveBeenCalledWith(revenuePrograms[0].id));
```
nrh-cklimas, line 57: This also wasn't part of the PR work but I saw failures in CI only on this test. Making this change fixed this issue.

spa/cypress/e2e/01-donationPage.cy.js
```diff
@@ -711,7 +708,7 @@ describe('User flow: canceling contribution', () => {
     cy.wait('@create-subscription-payment');
     cy.findByRole('button', { name: DEFAULT_BACK_BUTTON_TEXT }).click();
     cy.wait('@cancel-payment');
-    cy.contains(CANCEL_PAYMENT_FAILURE_MESSAGE);
+    cy.contains("Something went wrong, but don't worry, you haven't been charged. Try refreshing.");
```
GuiMend, line 711: In other places we have the `'` instead of `"`. Is the linter running in these files? I would expect the eslint test to fail in this PR because of these differences

spa/cypress/e2e/04-donations.cy.js
GuiMend: We have to wait for 2 `@getDonations`? Probably the page is calling an additional query that could be removed. I remember facing a similar issue that I had to add 2 waits and when I looked into the tested code I found it was calling the query twice (which was fixed and didn't need the 2 waits anymore)

spa/src/components/base/RichTextEditor/RichTextDisplay.tsx
```diff
@@ -0,0 +1,78 @@
+import PropTypes, { InferProps } from 'prop-types';
+import styled from 'styled-components';
+
+const RichTextDisplayPropTypes = {
+  /**
+   * HTML to display. **This will be rendered in the DOM as-is with no cleaning
+   * or security checks.**
```
GuiMend, line 7: Shouldn't we clean it? 🤔 
Doesn't seem like a good idea not to.

spa/src/components/base/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,48 @@
...
+  blockType: {
+    inDropdown: false,
+    options: ['Normal', 'H2', 'H3', 'H4', 'Blockquote']
+  },
+  link: {
+    defaultTargetOption: '_blank'
+  }
+};
+
+// We can't apply styles directly to <Editor> directly. The class names we're
```
GuiMend, line 24:
```suggestion
// We can't apply styles directly to <Editor>. The class names we're
```

spa/src/components/common/Modal/AddPageModal/AddPageModal.test.tsx
```diff
@@ -54,7 +54,8 @@ describe('AddPageModal', () => {
...
+    await waitFor(() => expect(onAddPage).toHaveBeenCalled());
+    expect(onAddPage).toHaveBeenCalledWith(revenuePrograms[0].id);
```
GuiMend, line 58: Does this mean it's being called 2 times?

spa/src/components/donationPage/pageContent/DRichText.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { axe } from 'jest-axe';
+import { render, screen } from 'test-utils';
+import DRichText, { DRichTextProps } from './DRichText';
+
+describe('DRichText', () => {
+  function tree(props?: Partial<DRichTextProps>) {
+    // This renders to a <li> which needs a container for AX purposes.
```
GuiMend, line 7: I'm not against using it this way, but wouldn't this be "cheating" the axe test? I think the axe test is supposed to guarantee that the component is accessible in it's own bubble, not dependant on external components (because they may or may not exist)

spa/src/components/donationPage/pageContent/DRichText.tsx
```diff
@@ -0,0 +1,36 @@
...
+    </DElement>
+  );
+}
+
+DRichText.propTypes = DRichTextPropTypes;
+DRichText.type = 'DRichText';
+DRichText.displayName = 'Rich Text';
+DRichText.description = 'Add arbitrary rich text to your contribution page';
+DRichText.required = false;
+DRichText.unique = false;
```
GuiMend, line 34: What are these for? Are they default props?

spa/src/components/pageEditor/elementEditors/richText/RichTextElementEditor.tsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  return (
+    <Root>
+      <RichTextEditor editorState={editorState} onEditorStateChange={handleEditorStateChange} />
+    </Root>
+  );
+}
+
+(RichTextElementEditor as any).for = 'DRichText';
```
GuiMend, line 26: Also wondering what this does

spa/cypress/e2e/01-donationPage.cy.js
nrh-cklimas: This is [prettier doing its thing](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOEA2cB0mIDmAFAOQAqAFgJZoAEFAhnQ7QDqsm1VRpUAmcLjBIBKANwgANCAgAHGFXTJQDAE6qIAdwAKahGmQgGmTQwCeB6QCNVDMAGs4MAMoMAtnAAy3OMgBmxmhw1rYOTs6ydtwEyDCqAK7BIHBuVnB8AnyeDFAE8QwEcABiEKpuDDAKuYYM8TAQUiAUMG6YAOrU8GiRYHDO+lQKAG6DZoZgaJYg3EGqMNq2BOX+gUkAVmgAHs7R2ACK8RDwK5hB0pGqs4ZWDGmYjbKq3DBt-DAUyAAcAAznGkFtWyyQyPOCzIa+aQAR0O8AWcgMKCYAFooHB0ulGqo4DCqNiFgVlkgAqckkE3FRYgkybs4Acjr4kHFEtIYLdXnx3sgAEys2xUTDRADCEDcROSaAArI14kEyLdESSziAhokAJJQASwZxgJ7yACCmucMDM2BOQQAvhagA). I think eslint does run on the Cypress directory. I tried committing a change with bad indentation and it didn't accept it.

spa/src/components/common/Modal/AddPageModal/AddPageModal.test.tsx
nrh-cklimas: No, in general Jest asserts don't affect each other, so this waits for a call to occur, then tests what the call is. I don't know why this fixed it, but without it I was seeing test failures in CI (but not locally).

spa/src/components/donationPage/pageContent/DRichText.test.tsx
nrh-cklimas: I think most likely, this component and others shouldn't be a `<li>`. Contribution page content right now is structured as a `<ul>` with each element being an `<li>`, which to me doesn't really make sense. But making this change would be pretty big (would affect every content element) so I'd like to not do it in this PR.

spa/src/components/donationPage/pageContent/DRichText.tsx
nrh-cklimas: They are configuration for the page editor. See other components under `components/donationPage/pageContent/`.

spa/src/components/pageEditor/elementEditors/richText/RichTextElementEditor.tsx
nrh-cklimas: Same explanation as the other static attributes her too.

GuiMend (approved):
👍

spa/src/components/base/RichTextEditor/RichTextDisplay.tsx
nrh-cklimas: This seems like a pretty big expansion of scope to me, but do you have a library you'd suggest using? We don't do this currently in main [here](https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/donationPage/pageContent/DRichText.js#L8). I think enforcing the CSP policy will do the best job of addressing this, because it will block inline styles/JS that could be malicious.

GuiMend: We don't have to do it in this PR as we don't have it in main. But in a previous project that I had to render html, I used this `dompurify` library and all we need to do to clean the HTML is to run it through `DOMPurify.sanitize(html)` instead of using the raw `html`

nrh-cklimas: Updated in 03797d50f9d0b3b0afa85b29ac9f07149edb7805

x110dc (approved):

PR #900, opened 2022-12-14
DEV-2791: Show user email & id in sentry

GuiMend:

n/a

#### What's this PR do?
Adds user email and id to sentry so that errors are identifiable with the user that got the error.

This will be possibly be useful for a few different scenarios: 
- if we have an identifier for a contributor in Sentry we can use that to look in the logs/Stripe/Salesforce to get more detail about what happened
- if we have an ID of a Org Admin we can contact them about the issue to get more detail or assist them

With user id/email we can identify the above information.

- Log into review app and try to break the app so that a sentry error is shown with the user's email/id
- Sentry error: https://sentry.io/organizations/news-revenue-hub/issues/3808783797/?environment=dev-2791&project=6046263&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h

image

no

n/a

n/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-2791

no

no


spa/src/hooks/useUser.ts
```diff
@@ -44,8 +45,12 @@ function useUser(): UserHookResult {
     // When user logs in, the authentication endpoint returns user data and it gets
     // stored in localstorage. We should update the localstorage data each time user
     // is refetched so it isn't stale.
-    onSuccess: (data) => localStorage.setItem(LS_USER, JSON.stringify(data)),
+    onSuccess: (data: User) => {
+      Sentry.setUser({ email: data.email, id: data.id });
```
nrh-cklimas, line 49: Do we want to record the IP address too? https://docs.sentry.io/platforms/javascript/enriching-events/identify-user/#ip_address I know it's not in AC but it seems like we may as well while we are here.

GuiMend: Hey @nrh-cklimas , if you check sentry right now, it already captures the user's IP.

nrh-cklimas: I was looking at this:

> If the field is omitted, the default value is null. However, due to backwards compatibility concerns, certain platforms (in particular JavaScript) have a different default value for "{{auto}}". SDKs and other clients should not rely on this behavior and should set IP addresses or "{{auto}}" explicitly.

It's hard for me to fully understand what they mean by this, but it sounds like we shouldn't depend on them recording IP addresses by default?

nrh-cklimas (changes requested):

nrh-cklimas (approved):

spa/src/hooks/useUser.ts
GuiMend, line 49: Adding ip address:
```suggestion
      Sentry.setUser({ email: data.email, id: data.id, ip_address: "{{auto}}" });
```

x110dc (approved):

PR #910, opened 2022-12-20
DEV-3010 | Contributions are not appearing in contributions tab after being accepted in test

benjaminEwhite:

This PR solves for race conditions in our Stripe webhook receivers that were discovered in working on DEV-2342. It will be merged into DEV-2342, which will then go through a hopefully final round of QA before being merged into main.

#### What's this PR do?

This PR solves for race conditions in our Stripe webhook receivers that were discovered in working on DEV-2342. These race conditions were the root cause of data integrity problems we've been seeing with contributions not appearing where we think they should or missing data that should be there. 

More discussion and examples of race conditions can be found referenced in the JIRA ticket. Additionally, a comprehensive write up of the normal flows through our webhook receivers can now be found here: [Contributions Types, Stripe Entities, and Contribution Life Cycle](https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2323644441/Contributions+Types+Stripe+Entities+and+Contribution+Life+Cycle)

Here are the key changes:

- Configures `ContributionAdmin` to inherit from `CompareVersionAdmin` which proved very helpful when trying to run down cases of race conditions occurring.
- Adds `Contribution.having_org_viewable_status` which exclude contributions with  statuses that org users should not see.
- Fixes error in `complete_one_time_payment` where we should have been but were not updating contribution status.
- Uses `ContributionFactory` + traits in a number of places were "rolling our own" individual contributions within tests.
- Updates contribution queryset for contributions endpoint to exclude with respect to status (via `having_org_viewable_status`) whereas before, we were inferring status from presence/absence of `provider_payment_method_details`
-  When saving in Stripe webhook receivers, we now use `update_fields` to only save fields touched by the running method

- Fix contributions tab so that all relevant successful contributions appear there
- Prevent future data integrity problems from arising due to race conditions involving Stripe webhooks 

Unfortunately, this is not entirely straightforward to manually test, in part because we're testing for the absence of something that only intermittently happens (the race codition). 

For the narrow purpose for which this ticket was initially created, we can test by going through the checkout flow triggering flagged contributions, then accepting, then ensuring the contributions appear in contributions tab after being accepted. You might want to do this 5 or so times to convince yourself that the race conditions are no longer happening.

No

Yes

N/A

No

The investigation into webhook flow, which was necessary to understand bug in this ticket, is documented in this doc:

- [Contributions Types, Stripe Entities, and Contribution Life Cycle](https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2323644441/Contributions+Types+Stripe+Entities+and+Contribution+Life+Cycle)

[DEV-3010](https://news-revenue-hub.atlassian.net/browse/DEV-3010)

No

N/A

[DEV-3010]: https://news-revenue-hub.atlassian.net/browse/DEV-3010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/payment_managers.py
```diff
@@ -111,11 +111,14 @@ def complete_one_time_payment(self, reject=False):
                     stripe_account=revenue_program.payment_provider.stripe_account_id,
                     cancellation_reason="fraudulent",
                 )
+                # we rely on Stripe webhook for payment_intent.canceled that updates status on contribution
```
njharman, line 114: Good to add comment why there's no contribution.save() here.  Does this mean payment_intent.canceled results in Contribution.cancel() being called?

[huh, my muscle memory does two els. "While both canceled and cancelled are acceptable for the past tense of cancel, the version with one L is more common in American English, while the version with two L's is more common in British English."

apps/contributions/webhooks.py
```diff
@@ -84,14 +84,14 @@ def handle_payment_intent_canceled(self):
             contribution.payment_provider_data = self.event
             logger.info("Contribution %s canceled.", contribution)
 
-        contribution.save()
+        contribution.save(update_fields=["status", "payment_provider_data", "modified"])
```
njharman, line 87: Do the self.contribution.save() calls in ./payment_managers.py require similar update_fields? [Sorry don't understand flow enough to answer myself]

apps/contributions/webhooks.py
```diff
@@ -103,7 +103,7 @@ def handle_payment_intent_succeeded(self):
             self.obj_data["created"], tz=datetime.timezone.utc
         )
         contribution.status = ContributionStatus.PAID
-        contribution.save()
+        contribution.save(update_fields=["status", "last_payment_date", "payment_provider_data", "modified"])
```
njharman, line 106: provider_payment_id and others set above, but not updated? If unneeded, then delete above. If side effects, bleck and deserves comment.

apps/contributions/tests/test_webhooks.py
```diff
@@ -68,7 +68,15 @@ def test_when_contribution_found(self, payment_intent_succeeded, monkeypatch, cl
                 payment_provider_data=None,
                 provider_payment_id=payment_intent_succeeded["data"]["object"]["id"],
             )
+            spy = mocker.spy(Contribution, "save")
             response = client.post(reverse("stripe-webhooks"), data=payment_intent_succeeded, **header)
+
+        # the next two assertions are to ensure we're only allowing webhook to update a subset of fields
```
njharman, line 74: +1

njharman (changes requested):
Clarifying questions and possibly one small issue.

apps/contributions/models.py
```diff
@@ -426,7 +435,7 @@ def cancel(self):
                 self.provider_subscription_id,
                 stripe_account=self.donation_page.revenue_program.stripe_account_id,
             )
-        elif self.status == ContributionStatus.FLAGGED:
+        elif self.status == ContributionStatus.FLAGGED and self.provider_payment_method_id:
```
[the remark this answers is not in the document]
benjaminEwhite: This is defensive coding, in light of what I learned about lifecycle of contributions. While it shouldn't normally happen, there's a logical possibility whereby a contribution could have a flagged status but not have a `provider_payment_method_id`. If we don't have `provider_payment_method_id`, there's no action to take. But if it is there, we tell Stripe to detach the payment method from the customer, which guarantees that it cannot be used  for future charges for that customer. You're right that this is a "forget about it method" in that we just tell Stripe to update its data model, but there's no downstream action for us.

`provider_payment_method_id` only will be present if one of the following events has completed and been processed on our end via webhooks: 

- `payment_intent.succeeded` (happens in normal checkout flow for one-time and recurring)
- `customer_subscription.updated` (happens in normal checkout flow for recurring contributions and in other cases, and sets pm id subject to certain conditions)
-  `payment_method.attached"` (happens in checkout flow involving a recurring contribution, whether flagged or accepted -- crucially, this is the way a flagged-then-accepted contribution can have everything work out how we need it to in our domain model and in Stripe's )

apps/contributions/webhooks.py
benjaminEwhite: Good question. I don't think those need to be updated. There are two methods in that file that call `save`

- `complete_recurring_payment`
- `complete_one_time_payment`

These only get called when a flagged contribution is accepted via the Django admin or when the`auto_accept_flagged_contributions` task runs. Neither of those things will happen right around the time all of Stripe's webhooks fire having to with checkout flow.

apps/contributions/webhooks.py
benjaminEwhite: Missed those, and added to update_fields (+ updating tests) in this commit: 1d69d4d28

apps/contributions/payment_managers.py
njharman: I guess I should have been more clear, I didn't see where / how payment_intent.canceled updates status

x110dc (approved):

PR #919, opened 2022-12-30
DEV-2441: Display Org Name in Menu

GuiMend:

n/a

#### What's this PR do?
Add Org Name to Menu. Currently, if user has multiple Orgs, it will display the first org information.

Show the user their org name. As for now, there isn't a place where the organization is shown, so it can get confusing for the user.

- Log in with a new user or create an account and make sure there isn't an Org name showing in the menu
- Log in with a user that has an organization and make sure it's name shows up in the menu.

yes

yes

Organization name now appears in menu.

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-2441

no

no

[DEV-2441]: https://news-revenue-hub.atlassian.net/browse/DEV-2441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spa/src/components/common/OrganizationMenu/OrganizationMenu.tsx
```diff
@@ -0,0 +1,33 @@
+import PropTypes, { InferProps } from 'prop-types';
+import { faNewspaper } from '@fortawesome/free-solid-svg-icons';
+
+import { Title, Flex, OrgWrapper, Divider, OrgIcon, IconWrapper } from './OrganizationMenu.styled';
+
+export type OrganizationMenuProps = InferProps<typeof OrganizationMenuPropTypes>;
+
+const OrganizationMenu = ({ title, className, hideBottomDivider }: OrganizationMenuProps) => (
+  <Flex className={className!}>
```
benjaminEwhite, line 9: I've never seen post-fixed `!` -- can you explain what this is doing?

spa/src/components/dashboard/sidebar/DashboardSidebar.js
```diff
@@ -5,18 +5,25 @@ import ContributionSectionNav from './navs/ContributionSectionNav';
...
 import DashboardSidebarFooter from './DashboardSidebarFooter';
 import useFeatureFlags from 'hooks/useFeatureFlags';
+import OrganizationMenu from 'components/common/OrganizationMenu';
 
 function DashboardSidebar() {
+  const { user } = useUser();
   const { flags } = useFeatureFlags();
   const hasContributionsSectionAccess = hasContributionsDashboardAccessToUser(flags);
   const hasContentSectionAccess = flagIsActiveForUser(CONTENT_SECTION_ACCESS_FLAG_NAME, flags);
+  // TODO: temporarily use FIRST organization
```
benjaminEwhite, line 20: Can you provide more detail on what the TODO is here and create/link a JIRA ticket, please?

spa/src/components/common/OrganizationMenu/OrganizationMenu.tsx
```diff
@@ -0,0 +1,33 @@
+import PropTypes, { InferProps } from 'prop-types';
+import { faNewspaper } from '@fortawesome/free-solid-svg-icons';
+
+import { Title, Flex, OrgWrapper, Divider, OrgIcon, IconWrapper } from './OrganizationMenu.styled';
+
+export type OrganizationMenuProps = InferProps<typeof OrganizationMenuPropTypes>;
+
+const OrganizationMenu = ({ title, className, hideBottomDivider }: OrganizationMenuProps) => (
```
benjaminEwhite, line 8: Overall, this component seems fine, but I'm just wondering about the component name. `OrganizationMenu` makes me think that this component contains a menu within it, but IIUC, this component just displays an icon and name.

spa/src/components/common/OrganizationMenu/OrganizationMenu.tsx
GuiMend: This is a TS syntax to force TS to handle the previous value as non-null, non-undefined.

In this case, if TS is: `className: string | null | undefined`, className! would force TS to see it as: `className: string` 
We do this in this scenario because nothing happens if we pass in `null` or `undefined` to the className prop, but as the TS of that field only accepts `string`, it serves as a bypass.

spa/src/components/dashboard/sidebar/DashboardSidebar.js
GuiMend: What Mariclare mentioned to me was that the bussiness logic around this should be:
TL;DR
```
Multiple orgs -> hide
One org (no matter how many RP) -> show org name
```

Long version:
> because we don’t have a selection feature we should use org name (not RP), which solves for users with multiple RPs. for now, since we don’t have a way for users with multiple ORGs (e.g. Hub Staff) to identify themselves (or self-onboard), we can omit the org display for those users.

benjaminEwhite: Ok. Think either the TODO here needs a JIRA ticket with that work identified, or just remove this comment.

benjaminEwhite (approved):
Looks like Jest and Cypress tests are failing.  Code looks good. Just one comment to either add JIRA ticket number or remove TODO code comment.  

I'm going to go ahead and approve with understanding that tests will pass and above issue is resolved. No need for re-review.

spa/src/components/dashboard/sidebar/DashboardSidebar.js
GuiMend: Removed the comment already 👍

PR #927, opened 2023-01-04
DEV-3073 | Fix donor selected formatted amount and sync from Stripe management command

benjaminEwhite:

#### What's this PR do?

This PR addresses a variety of problems that together were causing data integrity problems and out of sync data between our system and Stripe's.

This PR started out with a new management command (`sync_missing_contribution_data_from_stripe.py`) that was narrowly meant to fix some data integrity issues that were identified in CSV exports.  In QAing the management commands and doing subsequent exports, an underlying problem crystalized:  We had numerous view functions (accessed through the rest API layer and the webhook layer) that were rather "loosey goosey" in save discipline -- this is to say, they did not make use of the `update_fields` option to limit saved fields to only those explicitly updated. 

This became a problem because when a user goes through checkout, it prompts entities to be created directly in our system via the API and then additional entities to be created after webhook events fire from Stripe.

There is no gaurantee of webhook delivery order, and before the changes in this PR, this meant that some views saved a contribution >= 2 times, while happening at the same time as webhooks targeting the same contribution, with all of these functions saving "blindly" (that is to say without using `update_fields`).

Additionally, this complicated, non-deterministic sequencing was very difficult to audit because we weren't making as good of use  of our django-reversion integration as we could have.

To solve for the above, in addition to furnishing a management command that can be used to fix existing data integrity problems, it also significantly refactors all code that creates and updates contributions, ensuring the following:

1. Within a given view function, the contribution is guaranteed to be saved at most 1 time within the request session.
2. Any time a contribution is updated, it will have a django reversion revision created so the change can be audited.
3. All saves of contributions set `update_fields` to decrease the chances of stale contribution state being saved.
4. Adds tests for all code affected by above ensuring that save is called once, that the right update fields are part of the save, and that Django reversion does the right thing.

Here's a more granular breakdown of some but not all of the changes in this PR:

- Adds a management command (`sync_missing_contribution_data_from_stripe`) that
  - Updates contributions that are incorrectly "stuck" in PROCESSING and sets `last_payment_date` appropriately.
  - Updates missing values for `provider_payment_method_id`, `provider_payment_method_details`, and `contribution_metadata` 
- Updates data source for some of the columns in contribution CSV exports:
    - `billing_details` is derived from `provider_payment_method_details`
     of contributor email in the CSV export to come directly from contribution object instead of Stripe data (as specified in AC for the ticket).
    - `formatted_donor_selected_amount` from `contribution_metadata` instead of `contribution.amount`. 
- Adds the following model properties on `Contribution`:
    - `stripe_setup_intent` which can be used to retrieve the setup intent, if any.
    - `stripe_payment_intent` which can be used to retrieve the payment intent, if any.
    - `stripe_subscription` which can be used to retrieve the subscription, if any.
- Adds several methods on `Contribution` model that get used in `sync_missing_contribution_data_from_stripe` management command:
    - `fix_contributions_stuck_in_processing`
    - `fix_missing_provider_payment_method_id`
    - `fix_missing_payment_method_details_data`
    - `fix_missing_contribution_metadata`
- Refactors the `BaseCreatePaymentSerializer.create_contribution` method into a `build_contribution` method which does not save internally.
- Creates a new task `email_contribution_csv_export_to_user` which is used to handle most of the work to do CSV exports.
- Replaces unittest testcase tests with pytest ones in several places
- Moves some fixtures that were in contributions/tests/test_views into `conftest.py` because they're useful elsewhere.
- Refactors our stripe payment manager tests (TestStripePaymentManager) to use pytest. Changes in this PR caused tests there to break and it was easiest to refactor the tests rather than try to understand old brittle test.

Recover from data integrity issues and allow users to get needed data in CSV exports 

A dev would need to run the management command in a given environment after identifying contributions they expect to update (ones with incorrect "processing" status or missing provider payment method details data), and then confirm that contributions are updated accordingly.

We could achieve this by SSHing into a review app and handpicking a complete contribution and setting its status to "PROCESSING" and/or setting its `provider_payment_method_details` to None.

No

Yes

Probably should be comms with impacted users once released letting them know they're good to do exports.

No.

No.

[DEV-3073](https://news-revenue-hub.atlassian.net/browse/DEV-3073)

No

Once merged into main, run management command.

[DEV-3010]: https://news-revenue-hub.atlassian.net/browse/DEV-3010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-3073]: https://news-revenue-hub.atlassian.net/browse/DEV-3073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/models.py
```diff
@@ -424,3 +424,114 @@ def stripe_metadata(contributor, validated_data, referer):
...
+                ContributionStatus.PAID,
+                ContributionStatus.FLAGGED,
+                ContributionStatus.REJECTED,
+                ContributionStatus.CANCELED,
+            ],
+            provider_payment_method_id__isnull=False,
+            provider_payment_method_details__isnull=True,
+        )
+        one_times_updated = 0
+        for contribution in queryset.filter(interval=ContributionInterval.ONE_TIME):
```
njharman, line 511: Not worth change here, probably.  But if we're writing these two queries (one times) and (recurrings) all over. Might consider creating custom Contribution manager with methods like .onetime() and .recurring() that return those querysets.

https://docs.djangoproject.com/en/4.1/topics/db/managers/#adding-extra-manager-methods

njharman (changes requested):
Looks good, as per usual. Only a few minor suggestions. Sigh, you never code anything wrong enough for a meaty review ;)

I do see ticket comment about adjusting AC to match the limited scope of this ticket. In cases like that, it would be helpful (for review) to list the AC this PR purports to fulfill.  I don't think it does all of them, but, I believe that is intended.  Not 100% sure. Has the AC edit happened?

apps/contributions/payment_managers.py
```diff
@@ -81,12 +81,12 @@ def get_serializer_class(self, data=None, contribution=None):
             return StripeOneTimePaymentSerializer
         return StripeRecurringPaymentSerializer
 
-    def attach_payment_method_to_customer(self, stripe_customer_id, org_strip_account, payment_method_id=None):
+    def attach_payment_method_to_customer(self, stripe_customer_id, org_stripe_account, payment_method_id=None):
```
njharman, line 84: This doesn't seem to be covered. But, it seems wrong to "punish" the guy fixing typo with having to also fix the lack of test coverage.

apps/contributions/models.py
```diff
@@ -424,3 +424,114 @@ def stripe_metadata(contributor, validated_data, referer):
...
         }
+
+    @property
+    def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
+        if not all(
+            [
+                pi_id := self.provider_payment_id,
+                acct_id := self.donation_page.revenue_program.payment_provider.stripe_account_id,
+            ]
+        ):
```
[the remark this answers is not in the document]
benjaminEwhite: 54a597b

apps/contributions/models.py
```diff
@@ -424,3 +424,114 @@ def stripe_metadata(contributor, validated_data, referer):
...
+        )
+        one_times_updated = 0
+        for contribution in queryset.filter(interval=ContributionInterval.ONE_TIME):
+            if (pi := contribution.stripe_payment_intent) and pi.status == "succeeded":
+                logger.info("Contribution with ID %s has a stale status of PROCESSING", contribution.id)
+                if not dry_run:
+                    logger.info("Setting status on contribution with ID %s to PAID", contribution.id)
+                    contribution.status = ContributionStatus.PAID
+                    contribution.save()
+                    one_times_updated += 1
```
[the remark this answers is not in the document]
benjaminEwhite: 2db9a23b2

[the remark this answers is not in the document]
benjaminEwhite: 2db9a23b2

apps/contributions/models.py
```diff
@@ -424,3 +424,114 @@ def stripe_metadata(contributor, validated_data, referer):
...
+    def fix_missing_payment_method_detail_details_data(dry_run: bool = False) -> None:
+        """Retrieve provider_payment_method_details from Stripe if it's None and it appears that this should not be the case.
+
+        For the eligible subset of contributions, we retrieve the payment data from Stripe and set `provider_payment_method_details`
+        on the NRE contribution.
+
+        For discussion of need for this, see discussion of Stripe webhook reciever race conditions in this JIRA ticket:
+        https://news-revenue-hub.atlassian.net/browse/DEV-3010"""
+        queryset = Contribution.objects.exclude(provider_payment_method_id="").filter(
+            # For optimal data integrity, this function should be run only after `fix_contributions_stuck_in_processing`
```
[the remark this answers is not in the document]
benjaminEwhite: 019b9631136

apps/contributions/models.py
benjaminEwhite: 49d78c7617

apps/contributions/payment_managers.py
benjaminEwhite: While I'm happy to clean up the road, I don't want to pay the toll.

apps/contributions/models.py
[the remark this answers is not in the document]
njharman: nice

benjaminEwhite:

apps/contributions/serializers.py
```diff
@@ -221,10 +217,6 @@ def to_internal_value(self, data):
             data["amount"] = self.convert_amount_to_cents(data["amount"])
         return super().to_internal_value(data)
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.fields["amount"].error_messages["invalid"] = "Enter a valid amount"
-
```
benjaminEwhite, line 227: This is weird and has no related test or apparent use case, and tests don't break without it, so removed.

apps/contributions/models.py
```diff
@@ -555,3 +589,369 @@ def stripe_metadata(contributor, validated_data, referer):
...
         }
+
+    @property
+    def stripe_setup_intent(self) -> stripe.SetupIntent | None:
+        if not (
+            (si_id := self.provider_setup_intent_id)
+            and (acct_id := self.donation_page.revenue_program.payment_provider.stripe_account_id)
+        ):
+            return None
+        else:
```
x110dc, line 600: Is this `else` necessary?

apps/contributions/models.py
```diff
@@ -555,3 +589,369 @@ def stripe_metadata(contributor, validated_data, referer):
...
+                return None
+
+    @property
+    def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
+        if not (
+            (pi_id := self.provider_payment_id)
+            and (acct_id := self.donation_page.revenue_program.payment_provider.stripe_account_id)
+        ):
+            return None
+        else:
```
x110dc, line 622: Is this `else` necessary?

apps/contributions/models.py
```diff
@@ -555,3 +589,369 @@ def stripe_metadata(contributor, validated_data, referer):
...
+    @property
+    def stripe_subscription(self) -> stripe.Subscription | None:
+        if not all(
+            [
+                sub_id := self.provider_subscription_id,
+                acct_id := self.donation_page.revenue_program.payment_provider.stripe_account_id,
+            ]
+        ):
+            return None
+        else:
```
x110dc, line 646: Is this `else` necessary?

apps/contributions/serializers.py
```diff
@@ -480,30 +472,69 @@ def create(self, validated_data):
         that the PaymentElement should not be loaded.
 
         """
+        logger.info("`CreateOneTimePaymentSerializer.create` called with validated data: %s", validated_data)
         contributor, _ = Contributor.objects.get_or_create(email=validated_data["email"])
         bad_actor_response = self.get_bad_actor_score(validated_data)
-        contribution = self.create_contribution(contributor, validated_data, bad_actor_response)
+        contribution = self.build_contribution(contributor, validated_data, bad_actor_response)
         if contribution.status == ContributionStatus.REJECTED:
+            logger.info("`CreateOneTimePaymentSerializer.create` is saving a new contribution with REJECTED status")
```
x110dc, line 480: I appreciate the verbosity; but we can log the module, function name and line number by changing the logging format. That's probably the preferred way to capture this.

apps/contributions/views.py
```diff
@@ -309,44 +307,13 @@ def email_contributions(self, request):
...
-                data={"status": "failed", "detail": "Something went wrong generating CSV export"},
-                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            )
-        except Exception:
-            logger.exception("Unexpected error.")
-            return Response(
-                data={"status": "failed", "detail": "Something went wrong"},
-                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            )
+        queryset = self.get_queryset() if request.user.is_staff else self.get_queryset()
```
x110dc, line 310: It's the same result for both conditions, no?

apps/contributions/webhooks.py
```diff
@@ -83,36 +85,49 @@ def handle_payment_intent_canceled(self):
...
             contribution.payment_provider_data = self.event
             logger.info("Contribution %s canceled.", contribution)
-
-        contribution.save(update_fields=["status", "payment_provider_data", "modified"])
-        logger.info("Contribution %s canceled.", contribution)
+        with reversion.create_revision():
+            contribution.save(update_fields={"status", "payment_provider_data", "modified"})
+            logger.info("Contribution %s canceled.", contribution)
+            reversion.set_comment(
+                f"`StripeWebhookProcessor.handle_payment_intent_canceled` webhook handler ran for contribution with ID {contribution.id}"
```
x110dc, line 92: I don't think these log statements need to change as part of this PR, but maybe we open a ticket and add TODOs with that ticket ID?

x110dc (changes requested):
I requested a few small changes and had a few questions. Nothing big I don't think.

apps/contributions/serializers.py
benjaminEwhite: I agree. Leo and I were discussing this last week and created [DEV-3520](https://news-revenue-hub.atlassian.net/browse/DEV-3520) to adopt function name prefixing in our logging formatters and to standardize throughout the app.

apps/contributions/webhooks.py
benjaminEwhite: [DEV-3520](https://news-revenue-hub.atlassian.net/browse/DEV-3520)

apps/contributions/models.py
benjaminEwhite: 98fb73bfd6c

apps/contributions/models.py
benjaminEwhite: 98fb73bfd6c

apps/contributions/models.py
benjaminEwhite: 98fb73bfd6c

apps/contributions/views.py
benjaminEwhite: bc903576e8cf51

x110dc (approved):

PR #936, opened 2023-01-11
Dev 2745 user last login admin page

harsha09:
#### What's this PR do?
Adds `last_login` field to the users table in Django Admin.

Admin to know the context of user in on-boarding process.

Login to Django Admin -> Go to `Users` section -> `Last Login` column should be displayed.
Click on a user in user list -> `Last Login` field displayed which is read-only.

Changes only in Django Admin.

Added.

NA

NA

Not yet

[DEV-2745]

Adding one environment variable `UPDATE_LAST_LOGIN` to update the user last_login in `SIMPLE_JWT` auth. By default it is set to `True`.

NA

[DEV-2745]: https://news-revenue-hub.atlassian.net/browse/DEV-2745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/users/admin.py
```diff
@@ -26,6 +26,7 @@ def role_assignment(self, obj):
             None,
             {
                 "fields": (
+                    "last_login",
                     "email",
```
leo-dcfa, line 30: I know this is a small change but a unit test wouldn't hurt 
See example here:

https://github.com/newsrevenuehub/rev-engine/blob/main/apps/users/tests/test_admin.py#L33

I would add a unit test for both detail as well as list display

leo-dcfa (changes requested):
Small change requested

apps/users/admin.py
harsha09: addressed using 472dec2941a358700f53016787113e81ac0c5de5

leo-dcfa (approved):

leo-dcfa (approved):
Test performed:

Created new user leo.alves+dev-2745@fundjournalism.org
Logged in through sign-in page (not Django Admin)
Last login was updated

x110dc (approved):

PR #929, opened 2023-01-06
DEV-2518: add tax_id field to revenue program response

leo-dcfa:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

This PR has another stacked PR as part of it - Please merge #939 before merging the current PR to `main` 

#### What's this PR do?

- Add `tax_id` to revenue program response from rev-engine
- Makes `tax_id` available in Django admin
- Added new permissions `IsRpAdmin`, `IsOrgAdmin`, `IsGetRequest` and `IsPatchRequest`
- Added patch method to allow patching the tax_id in a revenue program
- Modified Revenue Program view to return responses under following conditions:

1. User is authenticated and
2. User is a super user OR
3. User is an org admin and it's a patch or get request OR
4. User is a rp admin and it's a get request

#### Why are we doing this? How does it help us?

As part of [DEV-3067], we are making org settings modifiable by the Org Admin. [DEV-2518] is about making the tax_id available for edit for the Org Admin and visible for the RP Admin

#### How should this be manually tested? Please include detailed step-by-step instructions.

Please note the test below below covers patching and retrieving a revenue program with a super user. I've added multiple tests in the unit tests to cover different types of permissions. The test below verifies patching and retrieving revenue program + asserting that tax_id is included in response

- Start the app with `make run-dev`
- Navigate to Django Admin and open any revenue program
- Copy the id for the revenue program
- Run ` curl -c cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -X POST -d '{"email":"{your username}", "password": "{your password}"}' http://localhost:8000/api/v1/token/`
- As part of the response from the above, a CSRF token is returned - you will use the token in
- Run ` curl -b cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -H "X-CSRFToken: {CSRF TOKEN from response above}"  -X PATCH "http://localhost:8000/api/v1/revenue-programs/{revenue program id from above}/" --data '{"tax_id": "223456789"}'\ | jq .` and assert the `tax_id` was updated
- Run ` curl -b cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -H "X-CSRFToken: {CSRF TOKEN from response above}" -X GET "http://localhost:8000/api/v1/revenue-programs/18/" \ |jq .`
- Assert that `tax_id` was included in the response

#### Did this PR make changes to the user interface that may require changes to the user-facing docs?

N/A

[...]

PR #939, opened 2023-01-12
Dev 2718 org admins can patch org name

leo-dcfa:

This PR is stacked on top of another PR - Please merge this PR before merging #929 to main

#### What's this PR do?

- Adds ability to patch org name to the org admin
- Modifies `OrganizationViewSet` model such `authenticated_classes` to allow responses under the following conditions:

1. User is authenticated and
2. User is super user or
3. User is Org Admin and is a patch request or
4. User has a role and it's a get request

This is part of larger epic [DEV-3067] - Org Settings V1 - where we are allowing org admins to access and modify certain settings for their orgs.

- Start the app with `make run-dev`
- Navigate to Django Admin and open any revenue program
- Copy the id for the revenue program
- Run ` curl -c cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -X POST -d '{"email":"{your username}", "password": "{your password}"}' http://localhost:8000/api/v1/token/`
- As part of the response from the above, a CSRF token is returned - you will use the token in the next step
- curl -b cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -H "X-CSRFToken: {csrf token from response above}" -X PATCH "http://localhost:8000/api/v1/organizations/25/" -d '{ "name": "{some new name}"} '| jq .` and assert the response has the update name

N/A

Yes

N/A

There is a bit of duplicate code on methods `test_permissive_users` and `test_restricted_users`; I opted for a bit of duplication over a complex set of if/else statements that obscured what was being tested.

N/A

[DEV-2718]

N/A

N/A

[DEV-3067]: https://news-revenue-hub.atlassian.net/browse/DEV-3067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-2718]: https://news-revenue-hub.atlassian.net/browse/DEV-2718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/organizations/models.py
```diff
@@ -151,6 +151,14 @@ def filter_queryset_by_role_assignment(cls, role_assignment, queryset):
...
             raise UnexpectedRoleType(f"{role_assignment.role_type} is not a valid value")
 
+    def user_has_ownership_via_role(self, role_assignment):
+        return all(
+            [
+                role_assignment.role_type == Roles.ORG_ADMIN.value,
+                role_assignment.organization == self,
+            ]
+        )
+
```
benjaminEwhite, line 161: I think we should converge on the approach I've started to try to standardize in PR for DEV-3149. That involves using a model manager, and by convention defining a common method called `filtered_by_role_assignment`, which takes a roleassignment instance as its argument and returns a filtered queryset.

You can see how we're using this method in a view for revenue programs here: https://github.com/newsrevenuehub/rev-engine/blob/DEV-3149-standardize-approach-for-testing-permissions-layer-of-api-with-organizations-app-as-first-instance/apps/organizations/views.py#L66-L70

apps/organizations/tests/test_org_views.py
```diff
@@ -64,11 +58,12 @@ def test_unauthed_cannot_access(self):
         self.assert_unauthed_cannot_put(self.detail_url)
 
     def test_expected_user_types_can_only_read(self):
-        for user in self.expected_user_types:
+        read_only_users = [self.hub_user, self.rp_user]
```
benjaminEwhite, line 61: After DEV-3149 gets merged into DEV-2518, we'll need to revisit these tests and make sure everything's squared away.

apps/organizations/views.py
```diff
@@ -32,22 +32,40 @@
...
 
     def get_queryset(self):
         # this is supplied by FilterQuerySetByUserMixin
         return self.filter_queryset_for_user(self.request.user, self.model.objects.all())
 
+    def patch(self, request, pk):
+        organization = get_object_or_404(Organization, pk=pk)
+        if not request.user.is_superuser and not organization.user_has_ownership_via_role(request.user.roleassignment):
+            return Response(status=status.HTTP_404_NOT_FOUND)
+        patch_serializer = serializers.OrganizationPatchSerializer(organization, data=request.data, partial=True)
```
benjaminEwhite, line 60: After DEV-3149 is merged into DEV-2518, we should be able to simplify this quite a bit. Need to revisit at that point.

benjaminEwhite (commented):
As mentioned in line, need to wait on DEV-3149 to be merged before this can move forward. It establishes a paradigm around testing and model managers that we should use moving forward instead of extending some in retrospect idiosyncratic patterns I created several months ago.

apps/organizations/tests/test_org_views.py
```diff
@@ -78,8 +77,50 @@ def test_unpermitted_methods(self, method, url_name, has_data, api_client, user,
...
+        ),
+    )
+    @pytest_cases.parametrize(
+        "user",
+        (
+            pytest_cases.fixture_ref("org_user_free_plan"),
+            pytest_cases.fixture_ref("superuser"),
+        ),
+    )
+    def test_permissive_users(self, method, url_name, has_data, api_client, user, organization):
```
leo-dcfa, line 100: @benjaminEwhite did my best to incorporate these cases in the method above but the if/elses become so complex that it didn't make much sense anymore. So I separated this test into another method and kept some duplication for the sake of clarity. What do you think?

apps/organizations/views.py
```diff
@@ -50,6 +57,19 @@ def get_queryset(self):
         else:
             return Organization.objects.none()
 
+    def patch(self, request, pk):
+        organization = get_object_or_404(Organization, pk=pk)
+        if not request.user.is_superuser and not self.model.objects.filtered_by_role_assignment(request.user.roleassignment):
```
leo-dcfa, line 62: thanks for the updates, now using `filtered_by_role_assignment` as indicated @benjaminEwhite

apps/organizations/models.py
leo-dcfa: See implemented code and let me know if it's not ok

benjaminEwhite:
Closing then tearing down then reopening in order to trigger a fresh review app, as required by recently merged migrations.

apps/organizations/tests/test_org_views.py
```diff
@@ -194,21 +155,102 @@ def test_list_when_expected_user(self, user, api_client, mocker):
...
+            ),
+            (
+                pytest_cases.fixture_ref("invalid_patch_data_unexpected_fields"),
+                status.HTTP_200_OK,
+                {},
+                True,
+            ),
+        ),
+    )
+    def test_patch_when_expected_user(
```
benjaminEwhite, line 216: @leo-dcfa ideally tests for patching org name should be added in paramterize call above for this test function. If the test implementation would have to change a lot, can test for this else where, but ideally, AC4

> Display Name is not modifiable by Revenue Program Admins (only by Org Admins)

would be tested for as one case among the several other parametrized ones here.

apps/organizations/views.py
```diff
@@ -31,27 +33,43 @@
 logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
```
leo-dcfa, line 35: To review: changes to OrganizationViewSet + patch method

apps/organizations/tests/test_org_views.py
```diff
@@ -29,61 +29,22 @@
...
-            pytest_cases.fixture_ref("superuser"),
-            pytest_cases.fixture_ref("hub_admin_user"),
-            pytest_cases.fixture_ref("user_no_role_assignment"),
-            pytest_cases.fixture_ref("contributor_user"),
-            None,
-        ),
-    )
-    def test_unpermitted_methods(self, method, url_name, has_data, api_client, user, organization):
-        """Test each combo of method and user to show that server responds with appropriate status code.
+@pytest.fixture
```
leo-dcfa, line 32: to review: TestOrganizationViewSet and other

apps/organizations/tests/test_org_views.py
leo-dcfa: @benjaminEwhite confused about a couple of things:
- `test_patch_when_expected_user` exists twice, is this intended?
- you added a test for `invalid_patch_data_unexpected_fields` expecting a 200 - what are we testing here?
- Using this test function to validate negative cases doesn't seem right... because of the combination of two `parametrize` decorators, we would need a bunch of if/else to cover all the cases...

apps/organizations/views.py
```diff
@@ -31,27 +33,43 @@
...
-            return Organization.objects.filtered_by_role_assignment(ra)
-        else:
-            return Organization.objects.none()
+    def get_queryset(self) -> models.QuerySet:
+        return self.filter_queryset_for_superuser_or_ra()
 
+    def patch(self, request, pk):
+        organization = get_object_or_404(Organization, pk=pk)
+        if not request.user.is_superuser and organization not in self.get_queryset():
+            return Response(status=status.HTTP_404_NOT_FOUND)
```
benjaminEwhite, line 61:
```suggestion
    def patch(self, request, pk):
        organization = get_object_or_404(self.get_queryset(), pk=pk)
```

apps/organizations/tests/test_org_views.py
benjaminEwhite: @leo-dcfa `test_patch_when_expected_user` exists once on `TestOrganizationViewSet` and once on `TestRevenueProgramViewSet`.

Re: 200 and unexpected fields, that's just to codify the default behavior of our API -- namely, if you throw unexpected fields over the fence in patch data, things still work out okay. You can also choose to reject the entire request, but that would require higher LOE.

Re: 3rd comment IMO there's not an excessive amount of conditional nesting here.  There are other places where I've broken out thins out into more granular, specific tests when complexity in test set up / conditionality etc. is too much. That said, I like having a single test function that can succinctly tell the entire story of an endpoint  + method + set of users.

Would be totally fine to break this out into two separate tests -- one for happy path, one for validation errors. I can see why that might be easier to maintain.

benjaminEwhite: At the risk of being pedantic -- one thing we'll need to preserve if refactor this test is need to continue testing patch of owned vs. unowned resources for non-superuser.  Part of the reason I like the current implementation is because it tests many related behaviors in a small amount of code, but the downside is that it's easier to miss the point of each step in the test. But overall, I like being able to have a docstring like this one:

```
"""Show that expected users are able to patch (only) permitted RPs, with valid data"""
```

If your test actually does all that that docstring says, it's very clear how it aligns to business value.

(And again, to be clear, that has to be balanced vs. maintainability).

benjaminEwhite (changes requested):
@leo-dcfa this is close (and again, my apologies for dumping this huge diff on your lap!).

Left one suggestion to simplify the `patch` method.

Also left some thoughts on the test of patch endpoint.

apps/organizations/tests/test_org_views.py
```diff
@@ -252,6 +252,34 @@ def test_patch_when_expected_user(
...
+        response = api_client.patch(reverse("organization-detail", args=(organization.id,)), data={})
+        # if unauthed, get 401
+        if not user:
+            assert response.status_code == status.HTTP_401_UNAUTHORIZED
+        # if unexpected role assignment role type
+        else:
+            assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_patch_different_org(self, org_user_free_plan, api_client):
+        """Show that user cannot patch another org they are not part of"""
```
benjaminEwhite, line 277: Can be more specific in this comment — only org admins can access this patch endpoint.

benjaminEwhite (changes requested):
Trivial request for some explanation in a doc string. Other than that, there's a conversation from last review I think you may have missed: https://github.com/newsrevenuehub/rev-engine/pull/939/files#r1100519800

apps/organizations/tests/test_org_views.py
leo-dcfa: Sorry for missing this comment yesterday! But I think the new tests do cover the extra cases, namely:

- test_patch_when_unexpected_user: covers negative cases
- test_patch_different_org: covers security aspect

test_patch_when_expected_user covers positive cases/happy path. From that perspective I think we covered everything. I did chose to not have a method that covers all tests as I believe this is easier to reason about (+ I could not get it to work 😓 )

benjaminEwhite (approved):

apps/organizations/views.py
```diff
@@ -31,27 +33,41 @@
 logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
 
 
-class OrganizationViewSet(viewsets.ReadOnlyModelViewSet):
+class OrganizationViewSet(
```
leo-dcfa, line 36: To review: `OrganizationViewSet`:

- permission classe updates
- patch method

apps/organizations/tests/test_org_views.py
```diff
@@ -29,61 +29,22 @@
...
-            pytest_cases.fixture_ref("hub_admin_user"),
-            pytest_cases.fixture_ref("user_no_role_assignment"),
-            pytest_cases.fixture_ref("contributor_user"),
-            None,
-        ),
-    )
-    def test_unpermitted_methods(self, method, url_name, has_data, api_client, user, organization):
-        """Test each combo of method and user to show that server responds with appropriate status code.
+@pytest.fixture
+def org_valid_patch_data():
```
leo-dcfa, line 33: To review: `OrganizationViewSetTest` method

- `test_patch_when_expected_user`
- `test_patch_when_unexpected_user`
- `test_patch_different_org`

x110dc (approved):

benjaminEwhite:
After merging in upstream changes, need to close, then tear down, then reopen to deal with otherwise conflicting migrations that will block releases on review app.

benjaminEwhite:
In hindsight, since QA for this was local, probably could have not rebuilt review app...

PR #950, opened 2023-01-19
Dev 3131 handle fiscal sponsorship tax status

leo-dcfa:

This is the first branch of stacked PRs:

1. BE this: https://github.com/newsrevenuehub/rev-engine/pull/950
2. FE changes https://github.com/newsrevenuehub/rev-engine/pull/963

#### What's this PR do?

- Added fields `fiscal_sponsor_name` and `fiscal_status` to `RevenueProgram`
- Introduced new enum to record fiscal status choices
- Make `non_profit` field a property that returns true if `fiscal_status` is either nonprofit or fiscally sponsored
- Added migration that adds new fields, backfills them based on the current value of `non_profit` and finishes by removing the field
- Exposed new fields through Django Admin
- Took opportunity to modify some assertions to use pytest `assert` and `pytest.raises` instead in `test_views.py` in the user app
- Add ability to modify `fiscal_sponsor_name` and `fiscal_status` through `customize_account` api
- Added validation on both serializer as well as model to make sure that `fiscal_sponsor_name` can only be populated if a Revenue Program is fiscally sponsored
- Update .pre-commit config for isort - there were some failures trying to push code

On the Revenue Program object in our data model, we feature several fields that help organizations define their tax status. These settings are used to determine how fees should be calculated on the donation page and during payment processing, as well as how the Org is described in disclaimer language in our transactional receipts and payment reminder emails. 

##### Django Admin

- Go to Django Admin -> `Revenue Programs`
- Find a for profit Revenue Program
- Add a fiscal_sponsor_name + set the fiscal status to `fiscally sponsored` and save the record
- Assert that you are able to save
- Remove `fiscal_sponsor_name` and try to save again
- Assert that an error message is displayed - `"Please enter the fiscal sponsor name."`
- Add a fiscal sponsor name and set the fiscal status to `Nonprofit`
- Assert that an error message is displayed - `Only fiscally sponsored Revenue Programs can have a fiscal sponsor name.`
- Add a fiscal sponsor name and set the fiscal status to `For profit`
- Assert that an error message is displayed - `Only fiscally sponsored Revenue Programs can have a fiscal sponsor name.`

##### Customize Account API

There are more tests on unit tests, this is the happy path

- Run migrations with `python manage.py migrate`
- Run app with `make run-dev`
- Go to Django Admin and login
- Go to `Users` and find your user record; take note of user id
- Authenticate to API with `curl -c cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -X POST -d '{"email":"{your email}", "password": "{your password}"}' http://localhost:8000/api/v1/token/`
- Take note of the CSRF token in the response
- Call customize account with ` curl -b cookiejar -v -H "Content-Type: application/json" -H "Accept: application/json" -H "X-CSRFToken:{your token}" -X PATCH "http://localhost:8000/api/v1/users/135/customize_account/" --data ' '{"first_name": "first_name", "last_name": "last_name", "fiscal_status": "fiscally sponsored", "fiscal_sponsor_name": "News Revenue", "organization_name": "Azul Labs" }'\ | jq . 
- The above will create a new Revenue Program called `Azul Labs`; Navigate to Django Admin and go to Revenue Programs
- Find the Revenue Program for Azul Labs
- Assert that fiscal_sponsor_name is set to ATO

N/A

Yes

N/A

We need to tackle tests for customize_account at some stage, to move to pytest/parametrize. I decided not to tackle it on this ticket since we shifted the solution already which led to some refactoring.

N/A

[DEV-3131]

N/A

N/A

[DEV-3131]: https://news-revenue-hub.atlassian.net/browse/DEV-3131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/organizations/models.py
```diff
@@ -251,6 +251,7 @@ class RevenueProgram(IndexedTimeStampedModel):
     tax_id = models.CharField(blank=True, null=True, max_length=9, validators=[MinLengthValidator(9)])
     payment_provider = models.ForeignKey("organizations.PaymentProvider", null=True, on_delete=models.SET_NULL)
     domain_apple_verified_date = models.DateTimeField(blank=True, null=True)
+    fiscal_sponsor_name = models.CharField(max_length=63, null=True, blank=True)
```
benjaminEwhite, line 254: 63 is sort of a "magic number" here, so let's re-use the existing constant for that value `RP_SLUG_MAX_LENGTH` here, or else define a different constant (which could assign the value for `RP_SLUG_MAX_LENGTH` as its own value).

apps/organizations/models.py
```diff
@@ -328,6 +329,13 @@ def clean(self):
         if self.twitter_handle and self.twitter_handle[0] == "@":
             self.twitter_handle = self.twitter_handle.replace("@", "")
 
+        self.clean_non_profit()
+
+    def clean_non_profit(self):
+        # Ensure the presence of a Fiscal sponsor automatically treats the RevenueProgram as nonprofit
```
benjaminEwhite, line 335:
```suggestion
        """Ensure the presence of a fiscal sponsor automatically treats the RevenueProgram as nonprofit"""
```

apps/organizations/models.py
```diff
@@ -328,6 +329,13 @@ def clean(self):
...
+        if self.fiscal_sponsor_name is not None and self.fiscal_sponsor_name != "":
```
benjaminEwhite, line 336: Seems like it would be enough just to test that self.fiscal_sponsor_name is truthy.

```suggestion
        if self.fiscal_sponsor_name:
```

apps/organizations/tests/test_org_models.py
```diff
@@ -157,17 +158,35 @@ def test_slug_validated_against_denylist(self):
...
 
     def test_admin_benefit_options(self):
-        self.assertTrue(isinstance(self.instance.admin_benefit_options, list))
+        assert isinstance(self.instance.admin_benefit_options, list)
 
     def test_admin_benefitlevel_options(self):
-        self.assertTrue(isinstance(self.instance.admin_benefitlevel_options, list))
+        assert isinstance(self.instance.admin_benefitlevel_options, list)
+
+    def test_fiscal_sponsor_name_sets_non_profit(self):
```
benjaminEwhite, line 173: I'd rather see this test broken out into a module level pytest function just so we can make use of `pytest.mark.parametrize` instead  of this subTest.

you can have a look at how that's being done in this PR, among other places. https://github.com/newsrevenuehub/rev-engine/pull/952

apps/organizations/tests/test_org_models.py
```diff
@@ -157,17 +158,35 @@ def test_slug_validated_against_denylist(self):
...
+        fake = Faker()
+        for fiscal_sponsor_name, expected in [("", False), (None, False), ("NRH", True)]:
```
benjaminEwhite, line 175: These are the values to parametrize.

benjaminEwhite (changes requested):

apps/organizations/tests/test_org_models.py
leo-dcfa: I was split between doing that or not, initially tried to use `pytest.mark.parametrize` and discovered it doesn't work when a class extends `UnitTest` - will do it

GuiMend (dismissed):
@leo-dcfa , I believe this PR is also missing to add `fiscal_sponsor_name` in 
`patch: users/[userId]/customize_account`.
```
apps > users > serializer > CustomizeAccountSerializer
apps > users > views > UserViewset > customize_account
```

apps/organizations/tests/test_admin.py
```diff
@@ -124,3 +124,13 @@ def test_show_expected_fields_on_organization_pages(admin_client):
     assert soup.find("input", {"name": "show_connected_to_slack"}) is not None
     assert soup.find("input", {"name": "show_connected_to_salesforce"}) is not None
     assert soup.find("input", {"name": "show_connected_to_mailchimp"}) is not None
+
+
+def test_show_expected_fields_on_rp_pages(admin_client):
```
benjaminEwhite, line 129: This test name either needs to be narrower (`test_rp_page_admin_has_fiscal_sponsor_name`) or else should assert what it claims to assert in test name, and show that all expected fields appear.

apps/organizations/tests/test_org_models.py
```diff
@@ -66,7 +67,7 @@ def test_org_deletion_cascades_when_no_contributions_downstream(self):
         self.assertFalse(RoleAssignment.objects.filter(id=ra_id).exists())
 
 
-class RevenueProgramTest(TestCase):
+class RevenueProgramTest:
```
benjaminEwhite, line 70: Believe this test class won't be picked up by pytest's test runner.
```suggestion
class TestRevenueProgram:
```

apps/organizations/tests/test_org_models.py
```diff
@@ -66,7 +67,7 @@ def test_org_deletion_cascades_when_no_contributions_downstream(self):
...
     def setUp(self):
```
benjaminEwhite, line 71: This method should be deleted entirely. `setUp` is a unittest specific method. Direction we should be trying to go in is data setup for tests in (parametrized) fixtures

apps/organizations/tests/test_org_models.py
```diff
@@ -157,17 +158,33 @@ def test_slug_validated_against_denylist(self):
...
     def test_admin_benefit_options(self):
-        self.assertTrue(isinstance(self.instance.admin_benefit_options, list))
+        assert isinstance(self.instance.admin_benefit_options, list)
 
     def test_admin_benefitlevel_options(self):
-        self.assertTrue(isinstance(self.instance.admin_benefitlevel_options, list))
+        assert isinstance(self.instance.admin_benefitlevel_options, list)
+
+    @pytest.mark.django_db
+    @pytest.mark.parametrize("fiscal_sponsor_name, expected", (("", False), (None, False), ("NRH", True)))
```
benjaminEwhite, line 174: This is nitpicking, but i've been omitting space between commas in first arg to `.parametrize`. Though spaces are fine, i just followed convention in pytest docs (no space), and would prefer to keep this consistent throughout.  If there's a strong argument for adding space, that's something we can consider too, but probably would want to handle in a separate ticket. 

```suggestion
    @pytest.mark.parametrize("fiscal_sponsor_name,expected", (("", False), (None, False), ("NRH", True)))
```

apps/users/serializers.py
```diff
@@ -151,6 +151,7 @@ class CustomizeAccountSerializer(UserSerializer):
     """Special custom serializer to validate data received from customize_account method"""
 
     first_name = serializers.CharField(write_only=True, required=True)
+    fiscal_sponsor_name = serializers.CharField(write_only=True, required=False, default=None)
```
benjaminEwhite, line 154: Do we want to consider a max length here?

apps/users/tests/test_views.py
```diff
@@ -319,6 +319,7 @@ def setUp(self):
             "organization_name": "Test",
             "organization_tax_status": "nonprofit",
             "organization_tax_id": "987654321",
+            "fiscal_sponsor_name": "",
```
benjaminEwhite, line 322: Out of scope for this PR, but sure would like to refactor this test suite to only use pytest + parametrized fixtures at some point.

benjaminEwhite (changes requested):
Left several comments/requests.

High-level, I wanted to respond to the following point you made, @leo-dcfa:

> I went for the implementation choice that we infer a RevenueProgram is considered a nonprofit if there is an associated fiscal sponsor name. Removing the fiscal sponsor name will not remove the status of nonprofit automatically - is this something to be considered?

I do think this is important to consider both from an implementation and testing standpoint. If we ship as is, we're all but guaranteed to have data integrity issues where we won't be able to naively trust that when the API reports an RP as non-profit, that it is indeed non-profit. 

So one solution is to revisit the clean step and to handle the case of going from having fiscal sponsor to not having one. 

That all seems pretty "side effecty" to me. Haven't fully thought it through, but we might consider doing something like having a derived model property called called from a custom serializer field that derives the value for `is_nonprofit`  _at the API layer_ from the state of the `fiscal_sponsor_name` and `non_profit` fields on the model. IMO, it's cleaner to handle the interplay between fiscal sponosor name and non profit boolean in that layer of app.

leo-dcfa:
I see your point about having too much side effects in the transition between nonprofit and for-profit when a fiscal sponsor name is added. I did consider going down the route of using the view to derive if a record is nonprofit but that introduced another challenge: potential data integrity issues on Django Admin side. If on the Django Admin side we were to add a fiscal sponsor name and forget to make an RP for nonprofit, the UI would present a record that is technically incorrect. And if the org admin updates the fiscal sponsor name, then the RP would become nonprofit. 

A few alternatives:
- On the clean method, if the `fiscal_sponsor_name` has a value and `non_profit` is checked, we raise a validation error: fiscally sponsored RPs cannot be for profit. We remove the side effects but enforce integrity
- Introduce a new choice field `fiscal_status`: `for profit`, `nonprofit` and `fiscally sponsored`; make `is_nonprofit` a `@property` that `return True if self.fiscal_status in ("nonprofit", "fiscally sponsored")`. I am a bigger fan of this approach, but requires a bit more work

apps/organizations/tests/test_admin.py
leo-dcfa: that makes sense, fixed

benjaminEwhite:
@leo-dcfa I'm a big fan of the second solution. IMO it's worth the incremental additional effort spent (assuming that it is indeed incrementally more difficult) as we'll avoid an entire landscape of data integrity bugs that may otherwise end up consuming future dev time.

apps/users/serializers.py
leo-dcfa: good idea

apps/users/tests/test_views.py
leo-dcfa: agreed, I changed the assertions to pytest style to pave the way for that

apps/organizations/tests/test_org_models.py
leo-dcfa: reverted this change for now

apps/users/serializers.py
```diff
@@ -151,21 +151,37 @@ class CustomizeAccountSerializer(UserSerializer):
...
     last_name = serializers.CharField(write_only=True, required=True)
     job_title = serializers.CharField(write_only=True, required=False, default=None)
     organization_name = serializers.CharField(write_only=True, required=True)
     organization_tax_id = serializers.CharField(
         write_only=True, required=False, validators=[tax_id_validator], default=None
     )
-    organization_tax_status = serializers.ChoiceField(choices=["for-profit", "nonprofit"], required=True)
+
+    def validate_fiscal_status(self, value):
+        fiscal_sponsor_name = self.initial_data.get("fiscal_sponsor_name")
```
benjaminEwhite, line 168: Wondering what would happen here if there's a first patch that sets both fiscal_sponsor_name and fiscal_status, and then a subsequent patch that only sets fiscal_status. IIUC, in that case, `self.initial_data.get("fiscal_sponsor_name")` will be None, even though the RP instance. in question that value might be set. I haven't fully run down how this plays out, but can you look into it and see if there's anything to this?

benjaminEwhite: Further thought: may be that this should be a follow up ticket (non-blocking). Looking at the `customize_account` view function, looks like it doesn't have any guards against endpoint being hit up multiple times with same data, in which case on subsequent requests, there will be uncaught server errors. Probably want to get/or create or else validation behavior that catches this case, since it appears that this endpoint is only meant to be used once.

benjaminEwhite (commented):
Overall looking good. A couple of questions.

apps/organizations/tests/test_org_models.py
```diff
@@ -90,53 +92,53 @@ def test_slug_immutable(self):
...
         self.instance.refresh_from_db()
-        self.assertNotIn(slugify("A new Name"), self.instance.slug)
+        assert slugify("A new Name") not in self.instance.slug
 
     def test_slug_larger_than_100(self):
         fake = Faker()
         Faker.seed(0)
         long_slug_rp = factories.RevenueProgramFactory(name=f"{' '.join(fake.words(nb=30))}")
-        self.assertLessEqual(len(long_slug_rp.slug), 100)
+        assert len(long_slug_rp.slug) < 100
```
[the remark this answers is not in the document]
leo-dcfa: good point!

apps/users/serializers.py
leo-dcfa: https://news-revenue-hub.atlassian.net/browse/DEV-3236 great point, added here. Maybe we could use this as an opportunity to remove `UnitTest` from `customize_account`

GuiMend (commented):
LGTM @leo-dcfa 👍

GuiMend:
I've updated this PR's description to mention the stacked PRs.

benjaminEwhite:
Closing then tearing down then reopening to trigger fresh review app build so we don't get error triggered by renamed migration.

GuiMend:
Just a heads-up @leo-dcfa , https://github.com/newsrevenuehub/rev-engine/pull/963 was merged into this branch 👍

leo-dcfa:
thanks @GuiMend - cc @x110dc

benjaminEwhite:

x110dc (approved):

PR #951, opened 2023-01-20
DEV-3083: Create Button Border Preview component

GuiMend:

n/a

#### What's this PR do?
- Creates component to show border radius changes 
- Use the component in the current Create/Edit style modal

The current display of border radius changes in the style modal is confusing, so this component clarifies it.

- Check storybook
- Check Create/Edit Style Modal and make sure the border radius change is reflected in the preview component next to it
- Also test the preview in the live style editor

yes

yes

Border radius preview is now available in the Create/Edit Style modal to facilitate seeing what changes looks like.

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-3083

no

no


spa/src/components/common/ButtonBorderPreview/ButtonBorderPreview.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { render, screen } from 'test-utils';
+
+import ButtonBorderPreview, { ButtonBorderPreviewProps } from './ButtonBorderPreview';
+
+describe('ButtonBorderPreview', () => {
+  function tree(props?: Partial<ButtonBorderPreviewProps>) {
+    return render(<ButtonBorderPreview {...props} />);
+  }
+
+  it('should apply default border radius', () => {
```
nrh-cklimas, line 11:
```suggestion
  it('should default the border radius to 0', () => {
```

spa/src/components/common/ButtonBorderPreview/ButtonBorderPreview.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Flex, Dot, Position, Button, ArrowLeft, ArrowDown } from './ButtonBorderPreview.styled';
+
+export type ButtonBorderPreviewProps = InferProps<typeof ButtonBorderPreviewPropTypes>;
+
+const ButtonBorderPreview = ({ className, borderRadius = 0 }: ButtonBorderPreviewProps) => {
+  const dotPositions: Position[] = ['top-left', 'top-right', 'bottom-left', 'bottom-right'];
+
+  return (
+    <Flex className={className!}>
+      {dotPositions.map((position) => (
```
nrh-cklimas, line 13: An alternate approach to consider here--but not a must. We could create an SVG file for the dots like you did the arrows, I made something quick locally to test my idea:

```svg
<svg width="10" height="10" viewBox="0 0 10 10" fill="none" xmlns="http://www.w3.org/2000/svg">
<circle cx="5" cy="5" r="4.5" fill="white" stroke="#157CB2"/>
</svg>
```

Then we could use multiple backgrounds on `Flex` instead of doing this loop (and having associated types + styles) like so:

```ts
import BorderDot from 'assets/icons/button-border-dot.svg';

export const Flex = styled.div`
  background: url(${BorderDot}), url(${BorderDot}), url(${BorderDot}), url(${BorderDot});
  background-position: top left, bottom left, top right, bottom right;
  background-repeat: no-repeat;
  // rest of definitions
`;
```

Another idea that hit me was you could try to do it all in one background SVG using [border-image-slice](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-slice), but that's probably overkill.

nrh-cklimas (dismissed):

spa/src/components/common/ButtonBorderPreview/ButtonBorderPreview.tsx
GuiMend: Thanks for the suggestion @nrh-cklimas, but if it's ok with you I'd leave it as is because it was a pain to align all the components

spa/src/components/stylesEditor/StylesEditor.js
```diff
@@ -297,7 +298,7 @@ function StylesEditor({ styles, setStyles, handleKeepChanges, handleDiscardChang
               min={1}
               max={12}
             />
-            <ButtonBorderPreview borderRadius={getBaseFromRadii(styles.radii)} />
+            <ButtonBorderPreview borderRadius={getBaseFromRadii(styles.radii) * 2} />
```
nrh-cklimas, line 301: Style point--I'd suggest putting the logic into `getBaseFromRadii` instead of here so the calculation is in one place.

nrh-cklimas (approved):
Changes since my last review seem good--have one suggestion that doesnt need a re-review.

spa/src/components/stylesEditor/StylesEditor.js
GuiMend: Can't do it. This function is used elsewhere and can't have this `*2`. Also don't think it's necessary to create a new func just to use here.

x110dc (approved):

PR #974, opened 2023-02-01
Dev 3070 add event indicates page published

leo-dcfa:

N/A

#### What's this PR do?

- Refactored Google Cloud PubSub into a separate Django app
- Adapted usage of Google Cloud PubSub emulator for M1 Macbook
- Introduced Django Signals and using it in two places: user signals and page signals. 
- Removed coupling between Google Cloud PubSub message publishing from User model

- Introducing Django Signals means we can decouple operations independently from operations on Models
- We would like to notify whenever a page is first published (different ticket will tackle the notification to slack - [DEV-3069]
- Refactoring user model so that the publishing operation is decoupled from save operations leads us to a cleaner model

##### Test donation page message

- Add the following variables to your `.envrc`
```text
export NEW_USER_TOPIC=new-nre-customer-test
export GOOGLE_CLOUD_PROJECT=revenue-engine
export PUBSUB_PROJECT_ID=revenue-engine
export GOOGLE_APPLICATION_CREDENTIALS=fake-service-account.json
export PUBSUB_EMULATOR=[::1]:8085
export PUBSUB_EMULATOR_HOST=[::1]:8085
export PAGE_PUBLISHED_TOPIC=contribution-page-published
```
- Run `make run-gcloud-pub-sub`
- Start the dev server + google cloud pub sub emulator with make run-gcloud-pub-sub
- On a separate terminal window run  `python manage.py shell` and paste the following script:

```python
# this is a helpful script to connect to topic from the Django shell
import os

from google.cloud import pubsub_v1


topic_name = "projects/{project_id}/topics/{topic}".format(
    project_id=os.getenv("GOOGLE_CLOUD_PROJECT"), topic=os.getenv("PAGE_PUBLISHED_TOPIC")
)

print(f"Subscribing to topic {topic_name}")

subscription_name = "projects/{project_id}/subscriptions/{sub}".format(
    project_id=os.getenv("GOOGLE_CLOUD_PROJECT"),
    sub="TEST",
)

print(f"Will attempt to delete existing subscription with name {subscription_name}")

try:
    pubsub_v1.SubscriberClient().delete_subscription(request={"subscription": subscription_name})
except:  # noqa
    print("Subscription did not exist prior; will continue")


def callback(message):
    print(message.data)
    message.ack()


with pubsub_v1.SubscriberClient() as subscriber:
    subscriber.create_subscription(name=subscription_name, topic=topic_name)
    print(f"New subscription created; listening messages in topic {topic_name}")
    future = subscriber.subscribe(subscription_name, callback)
    future.result()
```

The above will create a subscription to your local Google Cloud PubSub emulator
- Sign in and go to `/pages`
- Choose any unpublished page and publish it
- Go back to your terminal and assert that a message was delivered
##### Test user email message

- Add the following variables to your `.envrc`
```text
export NEW_USER_TOPIC=new-nre-customer-test
export GOOGLE_CLOUD_PROJECT=revenue-engine
export PUBSUB_PROJECT_ID=revenue-engine
export GOOGLE_APPLICATION_CREDENTIALS=fake-service-account.json
export PUBSUB_EMULATOR=[::1]:8085
export PUBSUB_EMULATOR_HOST=[::1]:8085
```
- Start the dev server + google cloud pub sub emulator with `make run-gcloud-pub-sub`
- On a separate terminal window run ` python manage.py shell`
- On the same window where the shell is running, copy and the following script and hit enter
```python
# this is a helpful script to connect to topic from the Django shell
import os

from google.cloud import pubsub_v1


topic_name = "projects/{project_id}/topics/{topic}".format(
    project_id=os.getenv("GOOGLE_CLOUD_PROJECT"), topic=os.getenv("NEW_USER_TOPIC")
)

print(f"Subscribing to topic {topic_name}")

subscription_name = "projects/{project_id}/subscriptions/{sub}".format(
    project_id=os.getenv("GOOGLE_CLOUD_PROJECT"),
    sub="TEST",
)

print(f"Will attempt to delete existing subscription with name {subscription_name}")

try:
    pubsub_v1.SubscriberClient().delete_subscription(request={"subscription": subscription_name})
except:  # noqa
    print("Subscription did not exist prior; will continue")


def callback(message):
    print(message.data)
    message.ack()


with pubsub_v1.SubscriberClient() as subscriber:
    subscriber.create_subscription(name=subscription_name, topic=topic_name)
    print(f"New subscription created; listening messages in topic {topic_name}")
    future = subscriber.subscribe(subscription_name, callback)
    future.result()
```
- Go to [create-account page](http://localhost:3000/create-account/) and create a new account (please note as part of tests I had to temporarily disable bad-actor api, user views.py, line 272)
- Go back to the terminal window where you ran the subscribe script and validate that the email address you entered has been printed out to the terminal 

N/A

Yes

N/A

No

N/A

[DEV-3070]

- We need to configure Google Cloud PubSub and create a new topic for publishing the page messages
- Rev Engine needs to have the environment variable `PAGE_PUBLISHED_TOPIC` set to new topic

- [DEV-3069]

[DEV-3069]: https://news-revenue-hub.atlassian.net/browse/DEV-3069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-3070]: https://news-revenue-hub.atlassian.net/browse/DEV-3070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/common/tests/test_utils.py
```diff
@@ -263,29 +264,11 @@ def test_ip_in_cf_connecting_header():
...
-
-def test_google_cloud_pub_sub_is_configured_project_missing(monkeypatch):
-    monkeypatch.setattr("django.conf.settings.ENABLE_PUBSUB", True)
-    monkeypatch.setattr("django.conf.settings.NEW_USER_TOPIC", "topic")
-    monkeypatch.setattr("django.conf.settings.GOOGLE_CLOUD_PROJECT", "")
-    assert not google_cloud_pub_sub_is_configured()
+@pytest.mark.parametrize(
+    "enable_pubsub, gcloud_project, expected",
+    ((True, "project", True), (False, "project", False), (True, None, False), (False, None, False), (True, "", False)),
+)
```
benjaminEwhite, line 270: 🥇

apps/pages/signals.py
```diff
@@ -0,0 +1,30 @@
...
+from apps.common.utils import google_cloud_pub_sub_is_configured
+from apps.google_cloud.pubsub import Message, Publisher
+from apps.pages.models import DonationPage
+
+
+logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
+
+
+@receiver(pre_save, sender=DonationPage)
+def donation_page_pre_save(sender, instance, update_fields, **kwargs) -> None:
```
benjaminEwhite, line 17: I'm a little concerned about this being a pre-save signal in that we're signaling to outside world that a page is published, when the save itself could fail, in which case the page would not in fact be published. Tracking down and managing side effects in a pub/sub system will be complex.

Any reason we can't have this be a post save?

apps/users/tests/test_signals.py
```diff
@@ -0,0 +1,33 @@
...
+
+import pytest
+
+from apps.google_cloud.pubsub import Message
+from apps.users.signals import user_post_save_handler
+from apps.users.tests.factories import create_test_user
+
+
+@pytest.mark.parametrize(
+    "gcloud_configured, created, user_topic",
```
benjaminEwhite, line 11: Would prefer to consistently use "no space between param name" style as that's what's in pytest docs, and it' what the preponderance of existing usage is (and would like try to be uniform).

google_cloud_pub_sub/Dockerfile
```diff
@@ -1,11 +1,4 @@
-FROM google/cloud-sdk:406.0.0-alpine
-
-RUN apk --no-cache update && \
-    apk --no-cache add openjdk8-jre && \
-    rm -rf /var/cache/apk/*
-
-RUN gcloud components install --quiet beta pubsub-emulator && \
-    gcloud components update
+FROM google/cloud-sdk:414.0.0-alpine
```
benjaminEwhite, line 1: Just curious why we don't need the `RUN` commands. Does the updated cloud sdk version give us that for free?

benjaminEwhite (changes requested):
Overall this is a great direction. Really happy to see a bunch of side effects in save methods going away. Believe we should be able to emulate this approach to solve for [DEV-3026 | remove buried Stripe retrieve side effect from `Contribution.save`](https://news-revenue-hub.atlassian.net/browse/DEV-3026).

Left a few comments/requests for change, the biggest of which is probably the one about implications of page pub being a pre- rather than post-save signal.

apps/common/tests/test_utils.py
leo-dcfa: 🚀

apps/pages/signals.py
leo-dcfa: I had the same predicament. We need to know if the page is being published for the first time and in the post_save signal, we cannot determine it; at least I couldn't come up with a way of doing it. Do you have any suggestions?

apps/users/tests/test_signals.py
leo-dcfa: I'm sorry - it's not the first time you asked me to do this; will try to remember next time, force of habit

google_cloud_pub_sub/Dockerfile
leo-dcfa: Exactly, the updated version just worked

apps/users/tests/test_signals.py
```diff
@@ -0,0 +1,33 @@
...
+    monkeypatch.setattr("django.conf.settings.NEW_USER_TOPIC", user_topic)
+    monkeypatch.setattr("django.conf.settings.ENABLE_PUBSUB", "True")
+    instance = MagicMock()
+    publisher.get_instance.return_value = instance
+    user = create_test_user()
+    user_post_save_handler(sender=MagicMock(), instance=user, created=created)
+    if not gcloud_configured or not created or not user_topic:
+        assert not publisher.called
+    elif gcloud_configured:
+        instance.publish.assert_called_with(user_topic, Message(data=user.email))
```
benjaminEwhite, line 33: See approach in equivalent test for pages. Think we need something similar here.

benjaminEwhite (changes requested):

benjaminEwhite:

apps/users/signals.py
```diff
@@ -0,0 +1,26 @@
...
+from apps.common.utils import google_cloud_pub_sub_is_configured
+from apps.google_cloud.pubsub import Message, Publisher
+from apps.users.models import User
+
+
+logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
+
+
+@receiver(post_save, sender=User)
+def user_post_save_handler(sender, instance, created, **kwargs) -> None:
```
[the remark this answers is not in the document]
leo-dcfa: This is part of the overall goal to switch to signals, I thought it was an easy win

apps/users/models.py
```diff
@@ -60,16 +58,6 @@ def get_role_type(self):
...
 
-    def save(self, *args, **kwargs):
-        email_publishable = self._email_is_publishable()
-        super().save(*args, **kwargs)
-        if google_cloud_pub_sub_is_configured() and email_publishable:
-            logger.debug("Will publish email to Google Cloud PubSub: %s", self.email)
-            GoogleCloudPubSubPublisher.get_instance().publish(settings.NEW_USER_TOPIC, Message(data=self.email))
-
-    def _email_is_publishable(self) -> bool:
-        return self._state.adding and self.email
```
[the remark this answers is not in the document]
leo-dcfa: At this point I would leave it here... when we decided to switch to signals + make google cloud a separate django app I thought it was an easy win. If you don't mind I prefer if we include it instead of trying to separate the changes

apps/pages/models.py
```diff
@@ -189,8 +194,12 @@ def clean_fields(self, **kwargs):
         super().clean_fields(**kwargs)
 
     def save(self, *args, **kwargs):
+        should_send_first_publication_signal = self.first_publication()
         self.set_default_logo()
         super().save(*args, **kwargs)
+        if should_send_first_publication_signal:
+            logger.info("DonationPage.save - sending signal page_published for page %s", self.id)
+            signals.page_published.send(sender=self.__class__, instance=self)
```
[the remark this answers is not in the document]
leo-dcfa: I think this suggestion wo't work. We need to save the value of should_send_first_publication_signal prior to calling the save method; that way we can check the db to see if the record was published for the first time

apps/users/signals.py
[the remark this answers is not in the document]
leo-dcfa: updated AC

apps/users/tests/test_signals.py
leo-dcfa: updated AC

apps/pages/models.py
```diff
@@ -215,6 +224,15 @@ def make_template_from_page(self, template_data={}, from_admin=False):
...
         return Template.objects.create(**merged_template)
 
+    def first_publication(self) -> bool:
+        if not self.published_date:
+            return False
+        if not self.id:
+            return True
+        else:
+            existing_page = DonationPage.objects.get(pk=self.pk)
+            return True if not existing_page.published_date else False
```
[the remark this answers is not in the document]
leo-dcfa: This method is called prior to saving the record (calling super().save()). That way we can check the value in db vs in flight record

apps/pages/tests/test_signals.py
```diff
@@ -0,0 +1,45 @@
...
+    [
+        (False, None),
+        (True, None),
+        (False, PAGE_TOPIC),
+        (True, PAGE_TOPIC),
+    ],
+)
+@patch("apps.pages.signals.Publisher")
+@pytest.mark.django_db
+def test_donation_page_page_published_handler(publisher, gcloud_configured, topic, monkeypatch):
```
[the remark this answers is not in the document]
leo-dcfa: I thought that gets covered on `test_first_published_pub_sub_behavior_when_pubsub_configured`, no? Especially after your suggestion, unless I missed something

leo-dcfa:
I did rename `should_send_first_publication_signal` method as suggested but kept the flow to call the method before we save so that we can compare in flight record with db record prior to saving

apps/pages/models.py
```diff
@@ -168,6 +174,11 @@ def __str__(self):
     def is_live(self):
         return bool(self.published_date and self.published_date <= timezone.now())
 
+    @property
+    def page_url(self) -> str:
+        http_scheme = "https://"
+        return f"{http_scheme}{self.revenue_program.slug}.{settings.SITE_URL.partition(http_scheme)[2]}/{self.slug}"
```
benjaminEwhite, line 180: Huh! i never knew about `.partition`!

apps/pages/models.py
```diff
@@ -209,6 +224,15 @@ def make_template_from_page(self, template_data={}, from_admin=False):
...
         return Template.objects.create(**merged_template)
 
+    def should_send_first_publication_signal(self) -> bool:
+        if not self.published_date:
+            return False
+        if not self.id:
+            return True
+        else:
+            existing_page = DonationPage.objects.get(pk=self.pk)
+            return True if not existing_page.published_date else False
```
benjaminEwhite, line 234: Is line 230 right? Shouldn't it be:

```
if not self.id and self.published_date:
```

?

apps/pages/tests/test_models.py
```diff
@@ -142,6 +145,78 @@ def test_cannot_delete_when_related_contributions(self):
...
+                        "page_id": page.pk,
+                        "url": page.page_url,
+                        "publication_date": str(page.published_date),
+                        "revenue_program_id": page.revenue_program.pk,
+                        "revenue_program_name": page.revenue_program.name,
+                        "revenue_program_slug": page.revenue_program.slug,
+                    }
+                )
+            ),
+        )
```
benjaminEwhite, line 216: Another advantage of moving message generation to a page model method is that test assertion around it here can be much cleaner (namely we could just do `publisher_spy.publish.assert_called_once_with(topic_name, Message(json.dumps(page.get_pubsub_page_published_message_data()`.

apps/pages/tests/test_signals.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            expected_payload = {
+                "page_id": page_instance.pk,
+                "url": page_instance.page_url,
+                "publication_date": str(page_instance.published_date),
+                "revenue_program_id": page_instance.revenue_program.pk,
+                "revenue_program_name": page_instance.revenue_program.name,
+                "revenue_program_slug": page_instance.revenue_program.slug,
+            }
+            publisher_instance.publish.assert_called_with(topic, Message(data=json.dumps(expected_payload)))
```
benjaminEwhite, line 45: Ditto here. Moving generation of this payload to a model method allow this test to be more concise/cleaner.

apps/users/signals.py
benjaminEwhite, line 16: This should have a doc string that explains basic business logic of this (namely that on new user creation, we publish a message).

benjaminEwhite (changes requested):
A few small requests left in line.

Other than that, there are a couple of unresolved conversations:

- https://github.com/newsrevenuehub/rev-engine/pull/974/files#r1101718624
- https://github.com/newsrevenuehub/rev-engine/pull/974/files#r1101736520

apps/pages/models.py
```diff
@@ -209,6 +224,15 @@ def make_template_from_page(self, template_data={}, from_admin=False):
         merged_template = {**page, **template}
         return Template.objects.create(**merged_template)
 
+    def should_send_first_publication_signal(self) -> bool:
+        if not self.published_date:
+            return False
+        if not self.id:
```
benjaminEwhite, line 230: This works, but either an `elif` or `if not self.id and self.published_date` just so we're iron clad.

apps/pages/models.py
leo-dcfa: I also discovered it as part of this change, pretty cool right?

apps/pages/models.py
```diff
@@ -183,8 +194,12 @@ def clean_fields(self, **kwargs):
         super().clean_fields(**kwargs)
 
     def save(self, *args, **kwargs):
+        should_send_first_publication_signal = self.should_send_first_publication_signal()
```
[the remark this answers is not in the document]
leo-dcfa: good point

apps/pages/models.py
leo-dcfa: updated it

apps/pages/signals.py
```diff
@@ -2,38 +2,35 @@
...
+            "donation_page_page_published_handler: Existing donation page published for the first time: %s", instance
+        )
+        message_data = {
+            "page_id": instance.pk,
+            "url": instance.page_url,
+            "publication_date": str(instance.published_date),
+            "revenue_program_id": instance.revenue_program.pk,
+            "revenue_program_name": instance.revenue_program.name,
+            "revenue_program_slug": instance.revenue_program.slug,
+        }
```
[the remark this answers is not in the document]
leo-dcfa: not sure it's a good idea to live it in the page model. I think the model stays cleaner, we pass on the instance to the signal and the signal worries about what it wants. Cleaner model, focused on business logic and not background processing

apps/pages/tests/test_models.py
```diff
@@ -148,6 +149,49 @@ def test_page_url(self):
...
 
 
+@pytest.fixture
+def donation_page_no_published_date():
+    return DonationPageFactory(published_date=None)
+
+
+@pytest.fixture
+def donation_with_published_date():
+    return DonationPageFactory(published_date=datetime.datetime.now())
```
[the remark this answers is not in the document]
leo-dcfa: Is it ok if we apply that change later? Not a big deal, just don't want to block this ticket further because of DEV-2718 and DEV-2518 which have been blocked for a while

apps/users/signals.py
leo-dcfa: done

apps/pages/tests/test_models.py
leo-dcfa: See comment here https://github.com/newsrevenuehub/rev-engine/pull/974/files#r1103724985

apps/pages/tests/test_signals.py
leo-dcfa: Same, comment here See comment here https://github.com/newsrevenuehub/rev-engine/pull/974/files#r1103724985

apps/pages/signals.py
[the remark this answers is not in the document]
leo-dcfa: I don't think there is a right or wrong answer about this one actually, in terms of implementation there is a balance between choosing how much we do on each side. If we consider `signals` as a flavour of events in event driven architectures, it's the debate between thin events or fat events. I went for the approach where the signal has more data and figures out what it needs on the handler size. the Django post_save and pre_save handlers pass the full instance to signal receivers. We could make a thin signal that has the very minimal for publishing data, which is another valid approach. 

I think before we have a clear direction between which we should adopt, I suggest we go with the current approach and refactor later if we think we prefer something different.

apps/pages/models.py
```diff
@@ -183,8 +194,15 @@ def clean_fields(self, **kwargs):
         super().clean_fields(**kwargs)
 
     def save(self, *args, **kwargs):
+        """should_send_first_publication_signal has to be called prior to saving the record
+        to allow us to compare in flight record with record in database
+        """
```
benjaminEwhite, line 199:
```suggestion
        # should_send_first_publication_signal has to be called prior to saving the record
        # to allow us to compare in flight record with record in database
```

benjaminEwhite (approved):

x110dc (changes requested):
Had one question.

apps/pages/signals.py
```diff
@@ -0,0 +1,42 @@
...
+        Publisher.get_instance().publish(settings.PAGE_PUBLISHED_TOPIC, Message(data=json.dumps(message_data)))
+    logger.warning(
```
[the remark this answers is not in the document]
leo-dcfa: Good catch! Fixed it

x110dc (approved):

njharman (approved):
Reviewed and approved. Wrote several typo / consideration comments. Nothing requiring rereview. You should just resolve them (ideally after reading them).

apps/google_cloud/pubsub.py
```diff
@@ -34,10 +34,10 @@ def publish(self, topic, message: Message):
...
     def get_instance(cls):
-        """Returns an instance of GoogleCloudPubSubPublisher;
+        """Returns an instance of Publisher;
         Singleton pattern was chosen here since it is not necessary to instantiate/authenticate multiple times with
         GoogleCloud, and it will only do it the first time the application requires it. Additionally, memory is saved by
         sharing the same instance across RevEngine"""
         if not cls.__instance:
-            cls.__instance = GoogleCloudPubSubPublisher()
+            cls.__instance = Publisher()
         return cls.__instance
```
njharman, line 43: This not covered by unittests

x110dc (approved):

PR #978, opened 2023-02-02
DEV-2857: Add validation to URL setup fields

nrh-cklimas:

n/a

#### What's this PR do?

Adds front-end validation of URL fields in the page editor setup tab.

Improved UX; prevents users from losing changes on save--e.g. previous to this, if you tried to save a change with an invalid URL, Django would reject it and all changes would be lost.

For each URL field in the Setup tab:

1. Edit it to text that isn't a valid URL (single space, URL without https:// in front of it, etc.)
2. Confirm that the field shows an error message and the Update button on the tab is disabled.
3. Edit the field to a valid URL and confirm you are able to save changes.
4. Confirm that deleting all text from the field is allowed.

We might want to capture the validation messages in some way.

Yes.

n/a

I thought about using an external module to test for a valid URL, but it's such a small function it didn't feel worth it.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-2857

No.

No.


spa/cypress/fixtures/pages/live-page-1.json
```diff
@@ -162,6 +162,8 @@
     "code": "CAD",
     "symbol": "🍁"
   },
+  "header_link": "",
```
nrh-cklimas, line 165: Needed because otherwise, Cypress tests would set these values to `undefined` which `isValidUrl` rejects. These should always be strings IRL, but they might be empty.

GuiMend (changes requested):
Not sure it's working entirely @nrh-cklimas:

image

It does work when I mess up the protocol part:
image

GuiMend (changes requested):
Spaces in between the URL now throws error 👍 

image

But it's still allowing some weird URLs like `[any char in the place of http]:[any string even without dots in the middle]`. Is that acceptable?

GuiMend (approved):
One small comment. No need for re-review

x110dc (approved):

PR #993, opened 2023-02-07
DEV-2719: Include "Organization" link in avatar menu and link to Org Settings Page

GuiMend:

This is the second PR of stacked PRs:

1. Org settings page: https://github.com/newsrevenuehub/rev-engine/pull/994
2. Link into AvatarMenu (this PR)
3. Enhancement (Warn message) - https://github.com/newsrevenuehub/rev-engine/pull/996
4. Handle Fiscal Sponsor in Org Settings - https://github.com/newsrevenuehub/rev-engine/pull/997

#### What's this PR do?
- Adds a new Material UI icon library that is more complete than the other we have and the long-term objective is to remove the older lib and swap all icons to this new lib. Talked about this with @nrh-cklimas.
- Update TS, jest & package.json to work with svgr icons
- Added Organization menu item to AvatarMenu
- Create a new protected router wrapper to handle AC 4. `users with access to multiple orgs, and users with access to multiple orgs cannot navigate to these pages`

- Make it possible for users to navigate to Organization page
- Standardize icon library

Test all 3 scenarios:
1) - Log in with HUB admin
    - Check that AvatarMenu does not have Organization button
    - Access "settings/organization" in the URL and make sure it works

2) - Log in with user with access to multiple Orgs
    - Check that AvatarMenu does not have Organization button
    - Access "settings/organization" in the URL and make sure it redirects to "/pages"

3) - Log in with user with access to 1 Org
    - Check that AvatarMenu has Organization button and that it works (redirects to "settings/organization")
    - Access "settings/organization" in the URL and make sure it works

yes

yes

Users now have access to settings/organization page

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-2719

no

no


spa/__mocks__/svgrMock.js
```diff
@@ -0,0 +1 @@
+module.exports = { ReactComponent: 'icon-mock' };
```
nrh-cklimas, line 1:
```suggestion
export const ReactComponent = 'icon-mock';
```

spa/package.json
```diff
@@ -101,8 +102,9 @@
       "\\.pnp\\.[^\\/]+$"
     ],
     "moduleNameMapper": {
-      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
-      "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js"
+      "\\.(jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
+      "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js",
+      "\\.svg": "<rootDir>/__mocks__/svgrMock.js"
```
nrh-cklimas, line 107:
```suggestion
      "\\.svg$": "<rootDir>/__mocks__/svgrMock.js"
```
Nitpicking

spa/src/components/settings/SettingsProtectedRoute.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const SettingsProtectedRoute = ({ children, ...props }: SettingsProtectedRouteProps) => {
+  const { user, isLoading } = useUser();
+  const userHasSingleOrg = user?.organizations?.length === 1;
+
+  if (!isLoading && !userHasSingleOrg) {
+    return <Redirect to={CONTENT_SLUG} />;
+  }
+
+  return <SentryRoute {...props}>{children}</SentryRoute>;
```
nrh-cklimas, line 18: This is a judgment call so feel free to disregard, but I feel like it would be easier to test if this just returns children instead of wrapping with a SentryRoute. My thinking is...

1) We can put tests around Sentry logging in one place
2) When updating routes, we don't have to remember that we don't need to wrap this component in SentryRoute

If we do have a SentryRoute here, we should have a unit test to make sure it's present.

spa/src/components/settings/SettingsProtectedRoute.tsx
```diff
@@ -0,0 +1,25 @@
+import { Redirect, RouteProps } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { SentryRoute } from 'hooks/useSentry';
+
+import useUser from 'hooks/useUser';
+import { CONTENT_SLUG } from 'routes';
+
+export interface SettingsProtectedRouteProps extends RouteProps {}
+
+const SettingsProtectedRoute = ({ children, ...props }: SettingsProtectedRouteProps) => {
```
nrh-cklimas, line 10: How about a name like `SingleOrgUserOnlyRoute`? I could see us potentially using this outside the organizations menu, and we may end up protecting settings routes with different criteria down the road.

spa/src/components/settings/SettingsProtectedRoute.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const { user, isLoading } = useUser();
+  const userHasSingleOrg = user?.organizations?.length === 1;
+
+  if (!isLoading && !userHasSingleOrg) {
+    return <Redirect to={CONTENT_SLUG} />;
+  }
+
```
nrh-cklimas, line 17: We should have a check for `isLoading` and either return nothing or a GlobalLoading. While we don't know whether the user has one org, we shouldn't show them UI for single-org users.

nrh-cklimas (changes requested):

benjaminEwhite:
Closing, tearing down, then reopening to trigger new review app build, as part of [DEV-3243 | fix hookdeck webhooks](https://news-revenue-hub.atlassian.net/browse/DEV-3243).

spa/src/components/settings/SettingsProtectedRoute.tsx
GuiMend: That was my first idea for this component as well but than I saw `ProtectedRoute` and made it similar to that. But I'll update it with your suggestion 👍 

```js
function ProtectedRoute({ contributor, ...props }) {
  if (isAuthenticated(contributor)) {
    return <SentryRoute {...props} />;
  }
  return <Redirect to={SIGN_IN} />;
}
```

spa/src/components/settings/SettingsProtectedRoute.tsx
GuiMend: Makes sense. Will update it 👍

spa/src/components/settings/SettingsProtectedRoute.tsx
nrh-cklimas: Change looks good, but there should be a unit test around it too (unless I'm missing it).

GuiMend: It's there but I renamed the test to be clear.

nrh-cklimas: I see, thanks.

nrh-cklimas (approved):

spa/src/components/routes/AddSlashToRoutes.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * Wraps a set of routes with a redirect to force a trailing slash. The route
+ * paths themselves *must* have a trailing slash.
+ */
+export const AddSlashToRoutes = ({ children }: AddSlashToRoutesProps) => {
+  const location = useLocation();
+  if (!location) {
+    throw new Error('AddToSlash must be a child of a Router');
+  }
+
+  // Only redirect if pathname doesn't end with `/` and doesn't start with `//`
```
nrh-cklimas, line 16: Could you add a quick comment explaining how the edge case can happen? This change looks good (and yay migrating to TypeScript) but I don't get the circumstances when this could happen and why we shouldn't do anything in this case.

GuiMend: Sure, will add that. But to explain it here: this edge case might happen if the user types in a url that starts with `//` and we shouldn't do nothing because:

1. If the URL doesn't match anything, we already have a redirect to `/pages/`
2. If we try to redirect with pathname containing 2 trailing slashes it thinks that it's because the pathname is part of `https://`. Example: if you type `www.[localhost].com:3000//settings` the function `AddSlashToRoutes` will try to redirect to `http://settings`, which will cause the sentry issue that popped up in Agustin's tests and will cause the app to crash (white screen).

x110dc (approved):

PR #1002, opened 2023-02-10
DEV-3135: Add state label configurability

nrh-cklimas:

n/a

#### What's this PR do?

Updates the contributor address block editor so that it can take on additional labels.

Better UX for non-USians.

1. Edit a contribution page and the Contribution Address block.
2. Check the Province checkbox and save changes.
3. Confirm that the field now shows as "State/Province" in the published page and that state is saved properly in a contribution.
4. Edit the page/block.
5. Check the Region checkbox and save changes.
6. Confirm that that the field now shows as "State/Province/Region" in the published page and that state is saved properly.
7. Create a new contribution page.
8. Confirm that the page only shows "State" as the label in the contributor address block.

We probably want to document this new functionality.

Yes.

Mention this to international orgs, perhaps?

Hit a very strange test failure that only happened in CI that doesn't appear related to this PR. It might just be the phase of the moon but I got it to work in c1fc4caa3bfa6da4e5e322ea83fdd0a121c575c8. I don't understand why it failed.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3135

No.

No.

### DIFF
--- a/spa/src/hooks/useContributionPage/useContributionPage.types.ts
+++ b/spa/src/hooks/useContributionPage/useContributionPage.types.ts
@@ -114,6 +114,18 @@ export interface AmountElement extends ContributionPageElement {
   };
 }
 
+export type DonorAddressElementAdditionalStateFieldLabel = 'province' | 'region';
+
+export interface DonorAddressElement extends ContributionPageElement {
+  content: {
+    /**
+     * Additional labels to show on the State field. These are *not* displayed
+     * as-is, and order is not significant.
+     */
+    additionalStateFieldLabels?: DonorAddressElementAdditionalStateFieldLabel[];
+  };
+}
+
 export interface DonorInfoElement extends ContributionPageElement {
   content: {
     /**
--- a/spa/cypress/e2e/02-donationPageEdit.cy.js
+++ b/spa/cypress/e2e/02-donationPageEdit.cy.js
@@ -277,7 +277,7 @@ describe('Contribution page edit', () => {
   });
 
   describe('Contributor address editor', () => {
-    it('should render the DonorAmountEditor', () => {
+    it('should render the DonorAddressEditor', () => {
       cy.editElement('DDonorAddress');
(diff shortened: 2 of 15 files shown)


spa/src/hooks/useContributionPage/useContributionPage.types.ts
```diff
@@ -96,6 +97,53 @@ export interface ContributionPageElement {
...
+  content: {
+    /**
+     * Show a "Province" label on the State field?
+     */
+    showProvinceOnStateField?: boolean;
+    /**
+     * Show a "Region" label on the State field?
+     */
+    showRegionOnStateField?: boolean;
+  };
```
benjaminEwhite, line 110: If using this approach with booleans, seems liked we'd want one for "state" label as well.

We could also just have the backend supply the right labels, in which case, the interface would expect an array of 1 or more enumerated string values.

nrh-cklimas: I think we'd at least want booleans here for localization reasons, so the UI could translate values as need be. (If we already had localization set up we could at least use localization keys, but we haven't thought that far yet.)

I'm not sure if we envision ever allowing people to disable the State label, but I'll ask for some cliarification.

GuiMend (commented):
Think the changes seem in the right direction. Could you update this PR with the base branch? It's showing additional changes that were added in the previous branch.

spa/src/hooks/useContributionPage/useContributionPage.types.ts
GuiMend: I think this is a good enough approach for what we need, but as you raised localization @nrh-cklimas, would we have a word for all 3 options in all supported languages? Probably good to handle the scenario where one of those (even when selected) could be undefined/null - or maybe handle this when time comes.

GuiMend (commented):
I'm good with this approach 👍

GuiMend (approved):
👍

x110dc (approved):

PR #1014, opened 2023-02-21
DEV-3304 | Fix bug that causes multiple receipt emails for recurring contributions

benjaminEwhite:

#### What's this PR do?

This PR fixes a bug whereby recurring contributions were getting a receipt email (when enabled by org) for each distinct recurring charge for the subscription.

This was happening because of a fundamental misunderstanding in our code as originally implemented re: the number of `payment_intent.succeeded` events that a single recurring contribution causes. Our code assumed there was only ever one such event for a given recurring contribution, but in fact, there's a distinct event for each individual charge.

I reached out to a Stripe dev to get their advice about best way to determine if a charge is the first on a subscription or not, and they suggested tracking that state in our db.  It turns out this is possible to deduce from the already extant `Contribution.status` field.  If a `payment_intent.succeeded` event fires for a contribution whose status is already paid, we can deduce that this is not the first charge, and can refrain from sending receipt email. On other hand, if the contribution's status is not PAID, we can deduce that this is the first invoice, in which case, we call the `handle_thank_you_email` method.

- Adds distinct fixtures for Stripe payment_intent.succeeded webhooks for single payments vs. subscriptions.
- Updates Stripe webhook receiver code to handle `payment_intent.succeeded` events differently depending on contribution interval.
- Updates tests accordingly

Fix bug that causes users to get confusing receipt emails.

To manually test, this we need to prove that a 2nd (or 3rd or 4th etc) payment for a recurring subscription does not trigger a thank you email, while the first payment does. 

The manual testing strategy requires using Stripe's [test clock feature](https://stripe.com/docs/billing/testing/test-clocks) in order to trigger the invoice.upcoming webhook. Here's what you'll need to do.

1. Decide which RP/page you'll contribute to, then go to the relevant org's change page in Django admin and ensure that the org is set up to send NRE receipt emails.
2.  Make an annual recurring contribution for the selected page (ensuring it will be approved). Confirm you get a receipt email.
3.  Go to the Django admin, find your new contribution, and grab its ID. You'll need this momentarily.
4. Next go to [test-clocks section of the Stripe dashboard](https://dashboard.stripe.com/acct_1JhHQsPbhdfvD1kV/test/subscriptions) for the connected account that you contributed to in step 1 (note that it's crucial you do this from connected account, otherwise things won't work out). There, click the "new simulation" button. Add a unique name. You can go with the default clock time, and save.
5. You'll be taken to the detail view for your test clock. Click the "Add" button to add a customer. The values for the data you provide here aren't that important (the email doesn't actually need to match the email you used in the contribution).
6. After adding the customer, click the "Add" button a second time and choose the "Add subscription" option. In that interface, choose the customer you just created. Choose one of the $20 products. You can go with default values for everything else, except for Payment, where you should click "add payment method" and use the 424242.... fake cc info.
7. After adding the subscription, you'll be taken back to the clock detail view. Click on the subscription to get taken to its detail view and grab the subscription ID from the browser address bar (the URL will look something like https://dashboard.stripe.com/test/subscriptions/sub_1M4plABMAMOLTEakrms2cOlo — you want the sub_... value). Leave this browser tab open, as you'll need to come back to it in a moment.
8.  Next, we need to add that subscription ID as the provider_subscription_id value for the contribution you created in step 7. To do that, you'll need to use the Heroku CLI to log in to the review app. From the command line, do  heroku run bash --app rev-engine-dev-2832-del-u6muvc. Then once you're logged in, do python manage.py shell. From there do the following:

    ```
    from apps.contributions.models import Contribution
    contribution = Contribution.objects.get(id=<id from step 2>)   # substitute in right value
    contribution.provider_subscription_id = "<subscription id>"  # substitute in right value
    contribution.save()
    exit()
    ```
    While you're here, you should also copy down the current value for `modified` and `last_payment_date` on the contribution. You can now close this command line terminal or otherwise exit from the heroku bash shell.

9. Now back in the browser tab for the Stripe subscription from step 6, click the "Advance time" button. Click on the date input and then click the +1 year option. Wait for the test clock to finish processing.
10. Wait for up to ~5 minutes or so for the webhook to process. Check your inbox. You should not see a new receipt emaill.
11. Back in your Heroku shell, re-retrieve the contribution from the db. Confirm that its `last_payment_date` and `modified` values are > what they previously were.
12. When you're done with all of the above, you can delete the test clock from Stripe.

No

Yes

N/A

No.

N/A

[DEV-3304](https://news-revenue-hub.atlassian.net/browse/DEV-3304)

No

No dev next steps. Could be an ops task to reach out to affected contributors, but that is being discussed in JIRA ticket.

[DEV-3304]: https://news-revenue-hub.atlassian.net/browse/DEV-3304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


benjaminEwhite:
Closing, tearing down, then rebuilding because adding a new Stripe webhook to listen for, and need review app rebuilt with that webhook.

benjaminEwhite:
Tearing down and rebuilding again because the name for webhook was incorrect in base.py

PR #1020, opened 2023-02-28
DEV-3254 |  Backend support for styled emails

benjaminEwhite:

#### What's this PR do?

This PR adds a ~default_style~ `transactional_email_style` property to our `RevenueProgram` model that can be used in transactional emails for configuring style characteristics.

The `RevenueProgram.transactional_email_style` returns an instance of a new `TransactionalEmailStyle` dataclass, which has the following properties:

- logo_url
- header_color
- header_font
- body_font
- button_color

If an rp's org is on the free plan, or if if the rp does not have a default donation page, a `HubDefaultEmailStyle` (which is an instance of `TransactionalEmailStyle`) will be used.

- Adds `style` as a context variable for our contributor-facing transactional emails (a separate ticket, DEV-2999 will start using these values to deliver styled emails). This value is produced by `asdict`ing a `TransactionalEmailStyle` instance.
- Adds a `TransactionalEmailStyle` datacalass and `HubDefaultEmailStyle` which is a preset instance of the class.
- Deletes the file `apps.organizations.tests.test_org_models.py` and moves all those tests into pre-existing `apps.organizations.tests.test_models.py`. The genesis of the two separate test files is unclear. I brought this into scope in part because in trying to run-down a bug in in-progress work, I spent a fair amount of time chasing a red herring, thinking that a failing test was caused by some kind of race condition related to pytest vs. non-pytest tests between these two modules.  Aside from this, this was low effort, and inches us closer to a "pytest only" world.
- Fixes a problem with our StyleFactory where they `Style.styles` attribute was hardcoded as a mutable dictionary instead of as lazy function that returns a new dictionary on each run.

Styled emails will increase trust

There is no good way to manually test this. 

As a dev, you could SSH into the review app, retrieve an RP that has a default donation page, and verify that the values returned in `rp.transactional_email_style` are correct.

No

Yes

N/A

No

N/A

- [DEV-3254](https://news-revenue-hub.atlassian.net/browse/DEV-3254)
- [DEV-2999](https://news-revenue-hub.atlassian.net/browse/DEV-2999)

No

[DEV-2999](https://news-revenue-hub.atlassian.net/browse/DEV-2999)

[DEV-3254]: https://news-revenue-hub.atlassian.net/browse/DEV-3254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/organizations/models.py
```diff
@@ -359,6 +375,19 @@ def stripe_account_id(self):
     def non_profit(self):
         return self.fiscal_status in (FiscalStatusChoices.FISCALLY_SPONSORED, FiscalStatusChoices.NONPROFIT)
 
+    @property
+    def default_style(self) -> DefaultStyle:
+        if not (page := self.default_donation_page):
+            return DefaultStyle()
+        _style = AttrDict(page.styles.styles if page.styles else {})
+        return DefaultStyle(
+            logo_url=urljoin(settings.SITE_URL, page.header_logo.url) if page.header_logo else None,
```
njharman, line 384: Huh! TIL that the single line if else does short circuit logic, in the logical way and not left to right.  I had to check it didn't like "A or B" where B not evaled if A True. 

[One reason I still enjoy doing reviews, even when I don't wanna do nuttin]

njharman (approved):
Approved ship it!

apps/organizations/models.py
njharman, line 384: I don't quite remember but weren't DP put under a branded URL?  Should that be used instead of SITE_URL.  ... I guess image will be available under either URL, right? So it doesn't matter?

benjaminEwhite: @njharman this comment ended up being super helpful, but not for exact reason you intended. I went down path of trying to include subdomain, and that led to me to realize/recall that in deployed environments, our static assets get deployed from Google storage (via white noise). What I had here would not have worked. 

See 566a75690c6f7

x110dc (approved):

PR #1023, opened 2023-03-01
DEV-3324: fix typescript eslint

nrh-cklimas:

n/a

#### What's this PR do?

- Adds the TypeScript eslint plugin and uses it on TypeScript files only.
- Merges the eslint config used by pre-commit by what was set inside `spa/`. Before, there were two separate config files: `.eslintrc.yaml` at the project root and `spa/package.json`. The config is now at `spa/.eslintrc.json`. This has the upside now that VS Code etc will use the correct linter config while editing.
- Adds a npm task for linting so that you can check as you go (`npm run lint`).

There are three commits total:

1. Update config.
2. Fix linter issues that the config update flagged. I didn't touch anything related to React hook dependencies. We should fix these, but changing them blindly can cause bugs.
3. Rewrite a test that got flagged by the linter so that it uses Jest. This test has not been running recently because it was named `.spec.js` instead of `.test.js`.

Improves code quality.

This should not have affected anything in the bundled app, but a general QA pass to make sure things look generally OK would be helpful.

No.

n/a

n/a

See note about React hook deps above

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3345

No.

No.


spa/src/utilities/isStringInString.ts
```diff
@@ -1,4 +1,4 @@
 export function isStringInStringCaseInsensitive(base: string, search: string) {
-  const regex = /[.,\/#!$%\^&\*;:@{}[\]=\-_`~()\s]/g;
+  const regex = /[.,/#!$%^&*;:@{}[\]=\-_`~()\s]/g;
```
GuiMend, line 2: I don't think changing the regex is a good idea. Think it broke it:
image

GuiMend:
One observation, the ticket number is **DEV-3345**, not **3324**. Not sure if this difference can cause confusion down the line

spa/src/utilities/isStringInString.ts
nrh-cklimas: Try switching regex101 to use the ECMAScript flavor on the left and it should be OK with it.

The linter error here was https://eslint.org/docs/latest/rules/no-useless-escape.

GuiMend (approved):

nrh-cklimas:
> One observation, the ticket number is **DEV-3345**, not **3324**. Not sure if this difference can cause confusion down the line

Good call--I think 3234 was the original story I was working on when I noticed the problem. I'll rename the branch and re-open the PR.

PR #1036, opened 2023-03-07
DEV-3302 | introduce google cloud secrets management and retire mailchimp access token from revenue program model

benjaminEwhite:

This is 1 of 5 stacked PRs

1. #1036 [This PR]
2. #1106 
3. #1060 
4. #1124
5. #1125 

QA will be done on the final stacked PR as that is the first one that the entire flow will be easily verifiable from front end.

#### What's this PR do?

This PR establishes a pattern for managing Django model fields whose values are stored via Google Cloud Secret manager.  It adds`apps.common.secrets.GoogleCloudSecretProvider` which is a [descriptor]() that can be used to store sensitive model field values through a secure mechanism. 

For a given revenue program, `rp.mailchimp_access_token` will return this value as generated by Mailchimp if the environment enables mailchimp, and a value has been saved to Google Cloud Secret Manager.  

Also note that ...

```
del rp.mailchimp_access_token
```

...deletes the access token from Google Cloud Secret Manager (not only from memory of current process).  

And....

```
rp.mailchimp_access_token = "foo"
```

...updates the access token value in Google Cloud, _even if without `.save()` or `.update()` being called on the model.  

Additionally, this PR:

- Removes the temporary model field `RevenueProgram.mailchimp_access_token`
- Adds a post-delete signal for revenue program that deletes mailchimp access token secret, if any.
- Adds a new settings var for `ENABLE_GOOGLE_CLOUD_SECRET_MANAGER` which will determine if a given environment uses Google Cloud secret manager.

It allows us to securely store sensitive data.

1. Sign up for an account in org dashboard.
6. In Django admin in separate tab, upgrade the corresponding org to core plan.
7. Find the id of the user created in step 1, and add it to the list of user ids for mailchimp access flag in Django admin.
8. Switch back to the org dashboard tab. Go to Settings. Click on the toggle to turn on the integration for Mailchimp.
9. Go through Mailchimp oauth flow.
10. You'll be routed back to the SPA. Wait a few seconds, then refresh the settings tab. The Mailchimp integration toggle should appear as "On". If it's not, wait a few more seconds and try refreshing again.

Given the changes in this PR, seeing the mailchimp integration as connected from settings tab is evidence that the Mailchimp access token secret has been securely stored.

No

Yes

N/A

No

No

[DEV-3302](https://news-revenue-hub.atlassian.net/browse/DEV-3302)

Yes. For any environment where we want this to work, we'll need to add env var for `ENABLE_GOOGLE_CLOUD_SECRET_MANAGER`, settings its value to the string value `true`.

N/A

[DEV-3302]: https://news-revenue-hub.atlassian.net/browse/DEV-3302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/common/google_cloud_secrets.py
```diff
@@ -0,0 +1,109 @@
+import logging
+
```
leo-dcfa, line 2: I think this file belongs inside the google cloud app, instead of commons. And should be called secrets_manager, to keep consistent with other file in app called `pubsub` representing another service provided by gcloud

apps/organizations/models.py
```diff
@@ -359,6 +371,48 @@ def stripe_account_id(self):
     def non_profit(self):
         return self.fiscal_status in (FiscalStatusChoices.FISCALLY_SPONSORED, FiscalStatusChoices.NONPROFIT)
 
+    @property
+    def mailchimp_access_token_secret_name(self) -> str:
```
leo-dcfa, line 375: In addition to the access token secret name, how can I retrieve server from switchboard api?

apps/common/google_cloud_secrets.py
```diff
@@ -0,0 +1,109 @@
...
+from google.cloud import secretmanager
+from google.cloud.secretmanager_v1.types import Secret, SecretVersion
+
+
+logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
+
+
+class GoogleCloudSecretManagerException(Exception):
+    """"""
+
```
leo-dcfa, line 15: Not sure if it's helpful, but I took a different approach in switchboard api where I composed a secret manager client https://github.com/newsrevenuehub/switchboard-api/pull/625/files#diff-3a0d4be3a54cf0bbe0a1fc9568965b284f2ca624aa64cec0a42f5ea09cb4614dR1

This is again similar to the approach here https://github.com/newsrevenuehub/rev-engine/blob/main/apps/google_cloud/pubsub.py

benjaminEwhite: I tend to prefer a more functional vs. OOP approach. That probably comes down to individual dev preferences, but in this case, I find I can follow code flow more easily because there's less indirection.

apps/organizations/models.py
benjaminEwhite: I believe that will just be on the serialized revenue program.

apps/common/google_cloud_secrets.py
benjaminEwhite: cf06b294f58

apps/google_cloud/secrets_manager.py
```diff
@@ -0,0 +1,109 @@
...
+        return secret
+    except PermissionDenied:
+        logger.warning("`get_secret` cannot access secret %s", secret_detail)
+        raise GoogleCloudSecretManagerException(f"Secret {secret_detail} is not accessible")
+    except NotFound:
+        logger.warning("`get_secret` cannot find a secret with secret_id %s in project %s", secret_id, project_id)
+        raise GoogleCloudSecretManagerException(f"Secret with secret_id {secret_id} not found in project {project_id}")
+
+
+def create_secret(secret_id: str, project_id: str = settings.GOOGLE_CLOUD_PROJECT) -> Secret:
```
njharman, line 34: This (usage) is not very clear.  I think I would have hidden/isolated the "version concept" in these functions. In general I would have hidden platform (google) implementation details from users of these functions. Ideally they would be swappable with another provider's functions.

1. create_secret() is creating a contianer for sercret versions?  
2. and use create_secret_version to actually store my secret which is "payload"?

apps/organizations/models.py
```diff
@@ -359,6 +364,35 @@ def stripe_account_id(self):
...
+    def mailchimp_access_token(self) -> str | None:
+        """The value for the RP's Mailchimp access token stored in Google Cloud Secrets Manager, if any.
+
+        Note that this property will return `None` if there are errors retrieving the secret or if the RP
+        is not configured to use Mailchimp.
+        """
+        if not settings.ENABLE_GOOGLE_CLOUD_SECRET_MANAGER:
+            return None
+        try:
+            return get_secret_version(self.mailchimp_access_token_secret_name).payload.data.decode("utf-8")
```
njharman, line 382: get_secret_versin should return string.  Do the decoding.  Caller of function should not (have to) know the internal data structure of Secret.

apps/organizations/models.py
```diff
@@ -359,6 +364,35 @@ def stripe_account_id(self):
...
+        except GoogleCloudSecretManagerException:
+            logger.debug(
```
njharman, line 384: [edit: I guess it's logged in get_secret_version().... nm?] Should this be debug? and not warning or error.  [It's configured but failed to work sounds like something we'd want to hear about]

apps/organizations/models.py
```diff
@@ -359,6 +364,35 @@ def stripe_account_id(self):
...
+                "`RevenueProgram.mailchimp_access_token` failed to fetch access token from Google Cloud Secrets for RP with ID %s",
+                self.id,
+            )
+            return None
+        except Exception:
```
njharman, line 389: What exception are we expecting here?  network failures? I'm at default mildly adverse to catching Exception.

apps/organizations/signals.py
```diff
@@ -0,0 +1,27 @@
...
+from apps.organizations.models import RevenueProgram
+
+
+logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
+
+
+@receiver(post_delete, sender=RevenueProgram)
+def delete_rp_mailchimp_access_token_secret(sender, instance, *args, **kwargs) -> None:
+    """When an RP is deleted, we delete the mailchimp_access_token_secret, if there is one"""
+    if instance.mailchimp_access_token:
```
njharman, line 17: (In secrets_manager)Should check for and log error if delete requested and settings.ENABLE_GOOGLE_CLOUD_SECRET_MANAGER is False. 

Or perhaps do nothing/log debug if it's expected only dev/test env won't have that set and we don't want spurious exceptions logged.

apps/organizations/signals.py
```diff
@@ -0,0 +1,27 @@
...
+        logger.info("Deleting mailchimp_access_token_secret for RevenueProgram %s", instance)
+        try:
+            delete_secret(
+                project_id=settings.GOOGLE_CLOUD_PROJECT, secret_id=instance.mailchimp_access_token_secret_name
+            )
+        except GoogleCloudSecretManagerException:
+            logger.exception(
+                "Failed to delete mailchimp_access_token_secret for RevenueProgram with ID %s", instance.id
```
njharman, line 25: Maybe we should define a code practice on how we log identifying information. I need consistency! :)

first logger its _instance.__str___, in second its _instance.id_

apps/organizations/models.py
```diff
@@ -359,6 +364,35 @@ def stripe_account_id(self):
...
+        Note that this property will return `None` if there are errors retrieving the secret or if the RP
+        is not configured to use Mailchimp.
+        """
+        if not settings.ENABLE_GOOGLE_CLOUD_SECRET_MANAGER:
+            return None
+        try:
+            return get_secret_version(self.mailchimp_access_token_secret_name).payload.data.decode("utf-8")
+        except GoogleCloudSecretManagerException:
+            logger.debug(
+                "`RevenueProgram.mailchimp_access_token` failed to fetch access token from Google Cloud Secrets for RP with ID %s",
```
njharman, line 385: Should GoogleCloudSecretManagerException as e: and include in log (it will be the first param, the explanation note given when exception raised.

apps/google_cloud/secrets_manager.py
```diff
@@ -0,0 +1,109 @@
...
+
+class GoogleCloudSecretManagerException(Exception):
+    """"""
+
+
+def get_secret_version(
+    secret_id: str, version_id: str = "latest", project_id: str = settings.GOOGLE_CLOUD_PROJECT
+) -> Secret:
+    client = secretmanager.SecretManagerServiceClient()
+    secret_detail = f"projects/{project_id}/secrets/{secret_id}/versions/{version_id}"
```
njharman, line 21: What's the relationship with this and client.secret_path() I see used below? I mean can secret_path() or some other client method be used to create this? This "magic" literal stands out...

apps/google_cloud/secrets_manager.py
```diff
@@ -0,0 +1,109 @@
...
+            "`create_secret` cannot create a secret with secret_id %s because one already exists for project %s",
+            secret_id,
+            project_id,
+        )
+        raise GoogleCloudSecretManagerException(
+            f"Secret with secret_id {secret_id} already exists for project {project_id}"
+        )
+
+
+def create_secret_version(
```
njharman, line 63: I don't see this called anywhere. How do secrets get added?

apps/google_cloud/secrets_manager.py
```diff
@@ -0,0 +1,109 @@
...
+    secret_id: str, payload: str, project_id: str = settings.GOOGLE_CLOUD_PROJECT
+) -> SecretVersion:
+    client = secretmanager.SecretManagerServiceClient()
+    parent = client.secret_path(project_id, secret_id)
+    try:
+        version = client.add_secret_version(request={"parent": parent, "payload": {"data": payload.encode("UTF-8")}})
+        logger.info("`create_secret_version` created a new version for secret with name %s", version.name)
```
njharman, line 70: secret_id, project_id

apps/google_cloud/secrets_manager.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def delete_secret(secret_id: str, project_id: str = settings.GOOGLE_CLOUD_PROJECT) -> None:
+    client = secretmanager.SecretManagerServiceClient()
+    try:
+        client.delete_secret(request={"name": (name := client.secret_path(project_id, secret_id))})
+        logger.info("`delete_secret` deleted secret with name %s", name)
+    except PermissionDenied:
+        logger.warning(
+            "`delete_secret` cannot delete secret with name %s because project %s is not permitted", name, project_id
```
njharman, line 100: secret_id

apps/organizations/signals.py
```diff
@@ -0,0 +1,27 @@
...
+def delete_rp_mailchimp_access_token_secret(sender, instance, *args, **kwargs) -> None:
+    """When an RP is deleted, we delete the mailchimp_access_token_secret, if there is one"""
+    if instance.mailchimp_access_token:
+        logger.info("Deleting mailchimp_access_token_secret for RevenueProgram %s", instance)
+        try:
+            delete_secret(
+                project_id=settings.GOOGLE_CLOUD_PROJECT, secret_id=instance.mailchimp_access_token_secret_name
+            )
+        except GoogleCloudSecretManagerException:
+            logger.exception(
```
njharman, line 24: Should GoogleCloudSecretManagerException as e: and include in log (it will be the first param, the explanation note given when exception raised.

apps/organizations/models.py
```diff
@@ -359,6 +364,35 @@ def stripe_account_id(self):
...
+        return f"MAILCHIMP_ACCESS_TOKEN_FOR_RP_{self.id}_[{settings.ENVIRONMENT}]"
+
+    @cached_property
+    def mailchimp_access_token(self) -> str | None:
+        """The value for the RP's Mailchimp access token stored in Google Cloud Secrets Manager, if any.
+
+        Note that this property will return `None` if there are errors retrieving the secret or if the RP
+        is not configured to use Mailchimp.
+        """
+        if not settings.ENABLE_GOOGLE_CLOUD_SECRET_MANAGER:
```
njharman, line 379: Check for settings should be in google_cloud.secrets_manager

njharman (changes requested):
Don't like it. There's too much interdependency between SecretManagerService Client, google_cloud.secrets_manager.py and where it's used (models.py, signals.py)

If we go to trouble of wrapping third party library we should some of (isolate means from callers of our wrapper)

- isolate implementation details of 3rd party lib
- isolate complexity of 3rd party lib
- genericize access so 3rd party lib can be swapped out for another
- add business logic/logging/features or otherwise alter how 3rd party operates

I would have liked to see more of that in this code. Specifically the isolate complexity and implementation details.

apps/organizations/signals.py
benjaminEwhite: It's common/expected behavior. Now logging when none.

benjaminEwhite:
Closing and tearing down in order to re-open and trigger fresh build because need the Google credentials file to be recreated.

benjaminEwhite:
Closing and reopening to trigger fresh build because of upstream migrations that affect migration dependencies herein.

benjaminEwhite:
Closing and reopening to trigger rebuild.

benjaminEwhite:
closing and reopening to trigger rebuild.

apps/common/secrets.py
```diff
@@ -0,0 +1,203 @@
...
+
+class SecretProvider:
+
+    """A descriptor that retrieves a secret from a secret manager."""
+
+    def __init__(self, version_id: str = "latest", *args, **kwargs) -> None:
+        self.version_id = version_id
+
+    def __get__(self, obj, type=None) -> object:
+        raise NotImplementedError("Subclasses of SecretProviderSecret should implement __get__ but does not")
```
njharman, line 25: comment mark to not test, I think  `# noqa`

apps/common/secrets.py
```diff
@@ -0,0 +1,203 @@
...
+    def __delete__(self, obj) -> None:
+        raise NotImplementedError("Subclasses of SecretProviderSecret should implement delete_secret but does not")
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+def _get_secret_manager_client() -> SecretManagerServiceClient:
+    credentials = service_account.Credentials.from_service_account_info(settings.GS_SERVICE_ACCOUNT)
+    return SecretManagerServiceClient(credentials=credentials)
```
njharman, line 39: not covered

apps/common/secrets.py
```diff
@@ -0,0 +1,203 @@
...
+        except NotFound:
+            logger.info(
+                "GoogleCloudSecretProvider couldn't delete secret %s at path %s because not found",
+                secret_name,
+                secret_path,
+            )
+            return
+        except PermissionDenied:
+            logger.exception("GoogleCloudSecretProvider cannot delete secret %s because permission denied", secret_name)
+            raise SecretProviderException("Permission denied")
```
njharman, line 203: [ignorable]  Man, I'd made a decorator, helper, something. Just to not have to repeat all this code. Whatever happened to Laziness being a programmer virtue?  Otoh, how many levels on indirection before inescapable confusion.

apps/common/secrets.py
```diff
@@ -0,0 +1,203 @@
...
+
+class GoogleCloudSecretProvider(SecretProvider):
+    """A descriptor that retrieves a secret from Google Cloud Secret Manager."""
+
+    # We define client here rather than in init to facilitate easy mocking in tests
+    client = (
+        _get_secret_manager_client()
+        if settings.ENABLE_GOOGLE_CLOUD_SECRET_MANAGER and settings.GS_SERVICE_ACCOUNT
+        else None
+    )
```
njharman, line 50: Why not just mock _get_secret_manager_client?

apps/common/secrets.py
```diff
@@ -0,0 +1,203 @@
...
+        raise NotImplementedError("Subclasses of SecretProviderSecret should implement __set__ but does not")
+
+    def __delete__(self, obj) -> None:
+        raise NotImplementedError("Subclasses of SecretProviderSecret should implement delete_secret but does not")
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+def _get_secret_manager_client() -> SecretManagerServiceClient:
```
njharman, line 37: Since this only works for Google, why is this not a member function of the GoogleCloudSecretProvider 
Where it could also be mocked.

apps/common/tests/test_secrets.py
```diff
@@ -0,0 +1,165 @@
...
+    def __init__(self, **kwargs):
+        for k, v in kwargs.items():
+            setattr(self, k, v)
+
+    val = GoogleCloudSecretProvider(MODEL_ATTR, "some-project-id")
+
+
+@pytest.mark.django_db
+class TestGoogleCloudSecretProvider:
+    @pytest.mark.parametrize("enabled", [True, False])
```
njharman, line 25: Appreciate that there are multiple paths to happiness

njharman (approved):
Approved.

Minor code coverage. Questions/considerations on code structure for testability. Philosophical ramblings.

apps/common/secrets.py
benjaminEwhite: Addressed in 84e4a4d3d16770e6d44ad405453b8ae466505463

apps/common/secrets.py
benjaminEwhite: This had to do with complexity around patching and when the GoogleCloudSecretProvider gets loaded initially vs. when we're able to patch it.  In order to test across context (that is, narrow unit tests of secrets, as well as testing features that use secrets), it proved easiest to have this as module level function.

apps/common/secrets.py
benjaminEwhite: No longer relevant. code outdated.

apps/common/secrets.py
benjaminEwhite: b80dbc2c7cdef4b09dc843e3535b053080035c38

deployment-tasks.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 set -e
 
->&2 echo "Running Migrations"
-python manage.py migrate --noinput
+
+>&2 echo "Running Migrations in deployment-tasks.sh"
+ENABLE_GOOGLE_CLOUD_SECRET_MANAGER=False python manage.py migrate --noinput
```
x110dc, line 6: Why is this here?

docker-entrypoint.sh
```diff
@@ -3,6 +3,8 @@ set -e
 
 >&2 echo "Running docker-entrypoint.sh"
 
+/bin/sh ./google_sa.sh
+
```
x110dc, line 7: Can you explain this change?

google_sa.sh
```diff
@@ -3,14 +3,6 @@ set -e
 
 FILE="google-sa/google-service-account.json"
```
x110dc, line 5: Can you explain this change?

x110dc (changes requested):
Just had a couple of questions.

google_sa.sh
benjaminEwhite: I made this change when I was trying to understand why the account credentials file was not available in Celery tasks before I understood the underlying issue (which was separate). 

IMO the change here is worth keeping as it simplifies the logic in this script --whether the file exists already or not, we're creating it anew. I don't see the value in removing the file only to recreate when can just override if already exists instead.

docker-entrypoint.sh
benjaminEwhite: This was leftover from when I was trying to run down missing Google account credentials file. Change not required. Removed.

deployment-tasks.sh
benjaminEwhite: I've now removed this change.

x110dc (approved):
LGTM.

PR #1037, opened 2023-03-07
DEV-2427: Show currency code in contributor-facing functionality

nrh-cklimas:

This is try #2 on this story.


GuiMend (approved):
LGTM 👍

apps/contributions/models.py
```diff
@@ -339,6 +340,22 @@ def expanded_bad_actor_score(self):
...
+            return {"code": self.currency.upper(), "symbol": settings.CURRENCIES[self.currency.upper()]}
+        except KeyError:
+            logger.error(
+                'Currency settings for stripe account/product "%s"/"%s" misconfigured. Tried to access "%s", but valid options are: %s',
+                self.stripe_account_id,
+                self.stripe_product_id,
+                self.currency,
+                settings.CURRENCIES,
+            )
+            return {"code": "", "symbol": ""}
```
benjaminEwhite, line 357: Should have unit tests on contribution model for `get_currency_dict` happy path, as well as the keyerror path.

benjaminEwhite (changes requested):
I only looked at Python code changes. Overall, they look fine, but need unit tests for the new model method `currency_dict`.

apps/contributions/models.py
```diff
@@ -339,6 +340,22 @@ def expanded_bad_actor_score(self):
             return None
         return self.BAD_ACTOR_SCORES[self.bad_actor_score][1]
 
+    def get_currency_dict(self):
```
benjaminEwhite, line 343: Not a big deal, but maybe could make this an `@property` instead of a method. So like this:

```
@property
def currency_dict(self):
....
```

Then you can use it like `contribution.currency_dict['symbol']`.

apps/contributions/models.py
nrh-cklimas: 👍 Updated in 99038653f150e0a973e71431389ebfd36dd3b43e

apps/contributions/models.py
nrh-cklimas: I left this one alone because I figured it'd be better to do this in conjunction with `get_currency_dict` on the PaymentProvider model, and wanted to keep scope on this PR relatively low.

apps/contributions/models.py
```diff
@@ -266,7 +266,8 @@ def __str__(self):
 
     @property
     def formatted_amount(self):
-        return f"{'{:.2f}'.format(self.amount / 100)} {self.currency.upper()}"
+        currency = self.get_currency_dict()
+        return f"{currency['symbol']}{'{:.2f}'.format(self.amount / 100)} {currency['code']}"
```
benjaminEwhite, line 270: Missed this first time through, but on non-typed functions that get touched in ongoing work, I'm trying to add type hinting.

So `formatted_amount` and `get_currency_dict` could become something like this (note i haven't run locally to confirm this works):

```
# these imports should go at top of file
from typing import TypedDict, Literal

class CurrencyDict(TypedDict):
    code: Literal[tuple([v for v in settings.CURRENCIES.keys())]
    symbol: Literal[tuple([v for v in settings.CURRENCIES.values())]

# ....

    @property
    def formatted_amount(self) -> str

# ....

    def get_currency_dict(self) -> CurrencyDict
```

benjaminEwhite (approved):
One final request to add type hints to functions touched in apps.organizations.models.py, but otherwise, I'm 👍 . I'm going ahead and approving now. No need for me to re-review commits with type hints.

njharman (approved):
Approved. Only comment that unittest should verify that error is not silent, it's actually logged.

apps/contributions/tests/test_models.py
```diff
@@ -576,6 +576,15 @@ def test_formatted_donor_selected_amount(self, trait):
             and contribution.formatted_donor_selected_amount == f"{contribution.amount} {contribution.currency.upper()}"
         )
 
+    @pytest.mark.parametrize("name, symbol", settings.CURRENCIES.items())
+    def test_get_currency_dict(self, name, symbol):
+        contribution = ContributionFactory(currency=name, provider_payment_method_id=None)
+        assert {"code": name, "symbol": symbol} == contribution.get_currency_dict()
+
+    def test_get_currency_dict_bad_value(self):
+        contribution = ContributionFactory(currency="???", provider_payment_method_id=None)
```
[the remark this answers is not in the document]
nrh-cklimas: Updated in 9447a8a424371d31c0c425cfea072ad8a20ead72

PR #1047, opened 2023-03-17
DEV-3337: Add core upgrade prompt to sidebar, migrate to TS

nrh-cklimas:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

n/a

#### What's this PR do?

This PR looks a bit hefty, but mostly it is me converting things to TypeScript/adding unit tests as I touched them.

- Adds a prompt to the sidebar to upgrade to Core (see Jira for the criteria for this appearing). Closing it should cause the prompt to go away immediately and not show up for the rest of the user session. It uses local storage to remember this.
- Adds an `outlined` variant to our base Button component.
- Adds the LinkButton component from https://github.com/newsrevenuehub/rev-engine/pull/1040

#### Why are we doing this? How does it help us?

More functionality around upgrading; continues TypeScript/Jest migration.

#### How should this be manually tested? Please include detailed step-by-step instructions.

To test the basic workflow:

1. Create a new account, but do not connect Stripe.
2. Confirm that the prompt doesn't appear in the sidebar. Log out of your session and back in to be sure.
3. Connect your account to Stripe.
4. Confirm that now the prompt appears.
5. Confirm that clicking on the Learn More button takes you to the pricing page in a new tab, and that nothing happens in the app itself.
6. Log out of your session and back in and confirm the prompt is still there.
7. Close the prompt.
8. Navigate around the app and confirm the prompt doesn't show up again.
9. Hit reload in your browser and confirm the prompt doesn't show up.
10. Log out of your session and log back in.
11. Confirm that the prompt is back.

To test other types of users not seeing it:

1. Log in as a Hub admin. Confirm the prompt doesn't show up.
2. Log in as a user who is on the Plus plan. Confirm the prompt doesn't show up.
3. If possible, log in as a user on the Core plan. Confirm the prompt doesn't show up.

#### Did this PR make changes to the user interface that may require changes to the user-facing docs?

We may want to document the prompt.

#### Have automated unit tests been added? If not, why?

Yes.

#### How should this change be communicated to end users?

n/a

#### Are there any smells or added technical debt to note?

I think I'm going to generalize out the logic of "remember this state until logout" into a hook in DEV-3335.

#### Has this been documented? If so, where?

No.

#### What are the relevant tickets? Add a link to any relevant ones.

https://news-revenue-hub.atlassian.net/browse/DEV-3337

#### Do any changes need to be made before deployment to production (adding environment variables, for example)?

No.

[...]

PR #1040, opened 2023-03-14
DEV-3330: Add max pages published modal

nrh-cklimas:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

N/A

#### What's this PR do?

- Adds a userCanPublish function to the useContributionPageList hook that consults a yet-to-be-created property in the backend
- Changes the Publish button so that it shows a new modal if the user has hit their plan limit
- Refactors out the published popover shown on a published page into a separate component for clarity
- Fixes bugs noted in https://news-revenue-hub.atlassian.net/browse/DEV-3366 and https://news-revenue-hub.atlassian.net/browse/DEV-3338
- Adds a Link base component for consistency in link appearance

#### Why are we doing this? How does it help us?

Adds needed functionality around the Core plan.

#### How should this be manually tested? Please include detailed step-by-step instructions.

To test that publishing works as usual:

1. Create a new page in an organization that is under the published page limit.
2. Publish the page using the Publish button and confirm the page is available, and nothing has changed from previous behavior.

To test that the new modal appears when the publish limit is hit:

1. Create a new page in an organization that has hit the published page limit.
2. Try to publish the page. Confirm that a modal appears explaining that the user has hit the page limit.
3. Confirm that the link in the modal goes to the pricing page, and that the modal can be closed.

To test that the publish modal closes on successful publish (https://news-revenue-hub.atlassian.net/browse/DEV-3366):

1. Create a new page or find an unpublished page.
2. Publish it.
3. Confirm that when you close the "Successfully published" modal, you're back at the page editor (e.g. the original publish modal asking for the slug is gone).

To test that the tooltip re: Stripe only shows when Stripe is not connected:

1. Create a new user.
2. Create a new page.
3. Confirm the publish button is disabled and that a tooltip asking the user to connect to Stripe appears when you point at it.
4. Set up Stripe for this user.
5. Confirm that the publish button is no longer disabled and that no tooltip appears when you point at it.
6. Publish the page.
7. Confirm that the publish button still isn't disabled (it should read "Published" now) and no tooltip appears when you point at it.

#### Did this PR make changes to the user interface that may require changes to the user-facing docs?

We should document what happens when you hit your plan limit.

#### Have automated unit tests been added? If not, why?

Yes.

#### How should this change be communicated to end users?

See doc comment.

#### Are there any smells or added technical debt to note?

[...]

PR #1056, opened 2023-03-23
DEV-3335: Add upgrade prompt to donations page

nrh-cklimas:

n/a

#### What's this PR do?

- Adds an upgrade prompt to the Contributions page.
- Refactors tracking of when users closed an upgrade prompt into a reusable hook.
- Adds a `<Link>` component which is verbatim from https://github.com/newsrevenuehub/rev-engine/pull/1040
- Renames `<CoreUpgradePrompt>` from https://github.com/newsrevenuehub/rev-engine/pull/1047 to `<SidebarCoreUpgradePrompt>` to avoid naming confusion with this other upgrade prompt.

Gets us closer to implementing Core upgrades.

To test the basic workflow:

1. Create a new account, but do not connect Stripe.
2. Confirm that the prompt doesn't appear when you go to the Contributions section. Log out of your session and back in to be sure.
3. Connect your account to Stripe.
4. Confirm that now the prompt appears when you go to the Contributions section.
5. Confirm that clicking on the Learn More button takes you to the pricing page in a new tab, and that nothing happens in the app itself.
6. Log out of your session and back in and confirm the prompt is still there.
7. Close the prompt.
8. Navigate around the app and confirm the prompt doesn't show up again.
9. Hit reload in your browser and confirm the prompt doesn't show up.
10. Log out of your session and log back in.
11. Confirm that the prompt is back.

To test other types of users not seeing it:

1. Log in as a Hub admin. Confirm the prompt doesn't show up.
2. Log in as a user who is on the Plus plan. Confirm the prompt doesn't show up.
3. If possible, log in as a user on the Core plan. Confirm the prompt doesn't show up.

We may want to document the prompt.

Yes.

n/a

I had to make a new hook for storing state in session storage. There are several existing ones in NPM that I documented in the file, but none say they are compatible with the version of React we're using. They probably would work anyway, but this requires changing npm install commands to override the peer dependency specified.

I positioned the upgrade prompt using absolute positioning instead of trying to use the `<Hero>` component, which hard-codes controls in the upper-right corner. We will probably want to revisit the component when we have more flexible needs there.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3335

No.

No.


spa/src/components/base/Button/Button.stories.tsx
```diff
@@ -24,6 +26,9 @@ const Template: ComponentStory<typeof Button> = (args) => {
           <Button {...args} color={color as ButtonProps['color']} disabled />
         </>
       ))}
+      <div style={{ alignItems: 'center', background: 'blue', display: 'grid', padding: 20 }}>
+        <Button {...args} disableRipple variant="outlined" />
+      </div>
```
GuiMend: Will probably get merge conflict in these files as we are adding these changes in 3 different PRs. (In one of them you treated the `disableRipple` inside the Button component

spa/src/components/dashboard/sidebar/DashboardSidebar.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+jest.mock('./navs/ContributionSectionNav');
+
+function tree() {
+  return render(<DashboardSidebar />);
+}
+
+describe('DashboardSidebar', () => {
+  const hasContributionsDashboardAccessToUserMock = hasContributionsDashboardAccessToUser as jest.Mock;
+  const useFeatureFlagsMock = useFeatureFlags as jest.Mock;
+  const useUserMock = useUser as jest.Mock;
```
GuiMend: Not sure if it's necessary, but I'm preferring to use `const useUserMock = jest.mocked(useUser);`. This pattern makes TS available and it identifies the TS return value for useUser (which in my opinion is useful)

spa/src/components/donations/DonationUpgradePrompts/DonationCoreUpgradePrompt/DonationCoreUpgradePrompt.styled.ts
```diff
@@ -0,0 +1,58 @@
...
+    top: 0px;
+
+    svg {
+      height: 16px;
+      width: 16px;
+    }
+  }
+`;
+
+export const Header = styled.h2`
```
GuiMend, line 20: Do we have a `h1` tag before? Semantically it's not great to use h2 if we don't have a h1 before (think we are using the headings more as a styling tool than as semantically correct).

Checking the code for this specific case, we can probably use h1 or a normal p tag as we set size and weight

spa/src/hooks/useSessionState.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    expect(result.current[0]).toBe('pass');
+  });
+
+  it('initially sets state to the default value if session storage is not available', () => {
+    window.sessionStorage.setItem('test', 'not parseable as JSON');
+
+    const { result } = renderHook(() => useSessionState('test', 'pass'));
+
+    expect(result.current[0]).toBe('pass');
+  });
```
GuiMend, line 36: Are these 2 tests correct? They are the same, but have different names

spa/src/components/donations/DonationUpgradePrompts/DonationCoreUpgradePrompt/DonationCoreUpgradePrompt.styled.ts
nrh-cklimas: The h1 in this case is on the page title, Contributions.

spa/src/components/dashboard/sidebar/DashboardSidebar.test.tsx
nrh-cklimas: The issue I've seen with this (and it happens in this test file) is that if you don't mock a return value that fully matches what the function would normally return, TypeScript complains. So you end up having to mock out values that don't have a direct effect on tests, or cast the mock calls to `any` (which defeats the purpose usually, unless you're trying to mock an abnormal response from the mocked function).

Anyway, I can change it in here but I think it will make it more verbose.

spa/src/components/base/Button/Button.stories.tsx
nrh-cklimas: I think rebasing this branch solved this.

spa/src/components/dashboard/sidebar/DashboardSidebar.test.tsx
GuiMend: Agree with that con, but on the upside, there has been times where I forgot a field when mocking a User/Org/RP and that TS flagged the issue. Not sure if you want to create a pattern on how we implement this in our test files, but I prefer `jest.mocked` 😅

spa/src/hooks/useSessionState.test.ts
GuiMend: Maybe if we override the window.sessionStorage function you might be able to test that scenario? 🤔 

Something like https://stackoverflow.com/questions/32911630/how-do-i-deal-with-localstorage-in-jest-tests

nrh-cklimas: Thanks for the link--updated in e52877345dfe3f1852df13aaf495cfba976044c5

spa/src/components/dashboard/sidebar/DashboardSidebar.test.tsx
nrh-cklimas: Updated in e52877345dfe3f1852df13aaf495cfba976044c5

PR #1064, opened 2023-03-28
DEV-3344-login-case-insensitive

harsha09:
#### What's this PR do?
User login should be case-insensitive w.r.t username (Email).
Database still maintains unique emails irrespective of case.

As usernames are not case agnostic the user should be given access if credentials are matched.

UI Page: https://dev-3344.revengine-review.org/sign-in/ 
Admin Page: https://rev-engine-dev-3344-log-cmjfvn.herokuapp.com/nrhadmin/login/?next=/nrhadmin/

Open create-account page -> Fill Email with all lowers -> Provide password -> click on create account -> verify email id
Next iteration -> Fill Email with all upper case -> Provide password -> click on create account 
Page should display `This email is already being used by an account. Try signing in.`

Sign In with the upper-case email id, you should be redirected to the dashboard.

Login to admin page -> Click on Add in users section and follow the same process.

NA

Yes

NA

Django 4 has a nicer way to implement this using functional unique constraints https://docs.djangoproject.com/en/4.0/ref/models/constraints/#expressions

No

[DEV-3344]

NA

No

[DEV-3344]: https://news-revenue-hub.atlassian.net/browse/DEV-3344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/users/models.py
```diff
@@ -55,6 +56,18 @@ def get_role_type(self):
         role_assignment = self.get_role_assignment()
         return (role_assignment.role_type, role_assignment.get_role_type_display()) if role_assignment else None
 
+    def validate_unique(self, exclude) -> None:
```
benjaminEwhite, line 59: Is this actually doing anything?  I commented out and ran tests locally, and all tests passed (and if you look at test coverage, this code never executes).

I removed `lookup="icontains"` from UserSerializer, and that caused tests to fail for view -- so maybe that's all that's needed for API layer?

apps/users/managers.py
```diff
@@ -15,3 +15,7 @@ def create_superuser(self, **kwargs):
         user.is_staff = True
         user.save(using=self._db)
         return user
+
+    def get_by_natural_key(self, username):
+        """Get user by case-insensitive username."""
+        return self.get(**{f"{self.model.USERNAME_FIELD}__iexact": username})
```
benjaminEwhite, line 21: Is this actually required? I tried commenting out and all tests passed locally.

benjaminEwhite (changes requested):
Seems like maybe some of the changes aren't needed??? 

Also, we should have tests of the name case insensitive uniqueness for the Django admin (so in apps.users.test_admin).

apps/users/models.py
harsha09: We will need `validate_unique` because, users can also be added from Django Admin, this function will enforce the unique constraint on Email (no need to convert to lower or upper case) as DB's unique constraint will fail.

If we remove `validate_unique` function email column can contain duplicate values and api will fail if it encounters more than one email.

For example: A user with email `Test@test.com` is already registered, when another user with email `test@test.com` is being added from Django Admin `validate_unique` function will validate the uniqueness irrespective of case.

apps/users/managers.py
harsha09: We need this as this will allow user to be able to login Admin page irrespective of case and using this method in `validate_unique`.

apps/users/models.py
harsha09: Added UT with bb64bab2d1e361f1114dd03b1f7325f2787080f9

apps/users/models.py
```diff
@@ -55,6 +56,18 @@ def get_role_type(self):
...
+        super().validate_unique(exclude)
+        _user = None
+        try:
+            _user = User.objects.get_by_natural_key(getattr(self, self.USERNAME_FIELD))
+        except User.DoesNotExist:
+            return
+
+        if _user and (self.pk is None or _user.pk != self.pk):
+            raise ValidationError("User with this Email already exists.", code="unique_together")
+
```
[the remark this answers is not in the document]
harsha09: Implemented with bb64bab2d1e361f1114dd03b1f7325f2787080f9

x110dc (approved):
This LGTM.

benjaminEwhite (approved):

PR #1065, opened 2023-03-28
DEV-3454 | Fix receipt emails bug

benjaminEwhite:

#### What's this PR do?

This PR solves for a bug that arose in receipt emails. 

Specifically, we discovered that for org's with custom email styles and a non-null value for the underlying donation page's header logo, our Celery task (`send_thank_you_email`) was failing because in attempting to retrieve the header logo url, which is a reference to a Google Cloud asset (via Whitenoise), because that call relies on a `google-service-account.json` file being available in the file system, and our async Celery tasks don't have access to that file system.

The merge of #1020 into main would seem to be what caused this issue. In that PR, we first made reference to the dynamic header logo URL value in email templates.

To solve for the above, this PR makes the following changes:

- Moves several db choice classes to `apps.contributions.choices.py`, to make importable without circular dependencies elsewhere.
- Deeply refactors `send_thank_you_email` so that it no longer needs file system access (and takes a new typed dict class (`SendThankYouEmailData`)).
- Adds a `make_send_thank_you_email_data` function that creates an instance of `SendThankYouEmailData` given a contribution.

Additionally, during code review, Norm raised a good point that the same error could happen (or could easily be coded to happen) in our other two email tasks: `send_templated_email` and `send_templated_email_with_attachment`. With that in mind, I've refactored those two tasks so that they render the message text and html and pass those values to the task, ensuring there is no possibility of the tasking trying to access out of scope file components.

Makes custom styles work in receipt emails

In the review app, [make a contribution on this page](https://billypenn-dev-3454.revengine-review.org/donate/) (where the org has been configured to have a plus plan, send NRE receipts, and a default donation page + style has been set up).

You should get a receipt email, and it should have custom header logo (a random screenshot from my computer).

No

Yes

N/A

No

N/A

- [DEV-3454](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/issues/DEV-3454)
- [DEV-2999](https://news-revenue-hub.atlassian.net/jira/software/c/projects/DEV/issues/DEV-2999)

No

No

[DEV-3454]: https://news-revenue-hub.atlassian.net/browse/DEV-3454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


njharman (changes requested):
Looks good.

1. Do not the send_templated emails also potentially suffer from this issue and need to be similarly refactored? The "Generally AC"
2. Noticed that apps/users/views.py:382 send_templated_email is called without delay (in signal handler for reset_password_token_created). I don't believe Django signals are asynchronous. But shouldn't the send mail be asynchronous .delay()?

apps/contributions/models.py
```diff
@@ -544,8 +488,10 @@ def cancel(self):
     def handle_thank_you_email(self):
         """Send a thank you email to contribution's contributor if org is configured to have NRE send thank you email"""
         logger.info("`Contribution.handle_thank_you_email` called on contribution with ID %s", self.id)
-        if self.revenue_program.organization.send_receipt_email_via_nre:
-            send_thank_you_email.delay(self.id)
+        if (org := self.revenue_program.organization).send_receipt_email_via_nre:
+            logger.info("Contribution.handle_thank_you_email: the parent org (%s) sends emails with NRE", org.id)
```
benjaminEwhite, line 492:
```suggestion
            logger.debug("Contribution.handle_thank_you_email: the parent org (%s) sends emails with NRE", org.id)
```

apps/contributions/models.py
```diff
@@ -544,8 +488,10 @@ def cancel(self):
     def handle_thank_you_email(self):
         """Send a thank you email to contribution's contributor if org is configured to have NRE send thank you email"""
         logger.info("`Contribution.handle_thank_you_email` called on contribution with ID %s", self.id)
```
benjaminEwhite, line 490:
```suggestion
        logger.debug("`Contribution.handle_thank_you_email` called on contribution with ID %s", self.id)
```

benjaminEwhite:
Thanks @njharman 

That's a good point re: the other email task. I ended up pushing up a commit that refactors both `send_templated_email` and `send_templated_email_with_attachment` so that they don't have possibility of accessing out of scope assets.  I've updated the PR description accordingly.

Re: apps/users/views.py:382 send_templated_email -- although Git blame shows that I wrote that code, I don't know off hand why I made it synchronous instead of async. I'm inclined here to leave as is with an "if it ain't broke..." mentality.

njharman (approved):
Ship it

x110dc:
Closing in favor of https://github.com/newsrevenuehub/rev-engine/pull/1090

PR #1075, opened 2023-03-30
DEV-3328 | Back end support for page restrictions

benjaminEwhite:

#### What's this PR do?

This PR adds a new plan — `CorePlan` — and introduces the notion of `publish_limit` as a plan attribute, which is used to control the number of pages an org can publish, given their plan. It also adjusts the `page_limit` property on existing plans, as that limit has changed.

More specifically:

- Adds a `validate_publish_limit` method to `DonationPageAdmin` and calls that method from `clean`.
- Adds a `validate_publish_limit` method to `DonationPageFullDetailSerializer`  and calls it from `validate`.
- Updates `apps.pages.tests.test_admin.py` to be pytest-only
- Updates`apps.pages.tests.test_serializers.py` to be pytest-only

Enable plan upgrades

First, as an admin user, [log in to the Django admin in the review app
](https://dev-3328.revengine-review.org/nrhadmin/). Find an organization that is on the Free plan, and then go to the pages admin.  Ensure there's currently one published page for the org.  Then try to create new page with non-null values for published date fields.  You should get an error message indicating you're at publish limit.

Next, also in the Django admin, set things up such that there are two pages for the free plan org — one that is published, and one that is unpublished. Go to the detail view for the unpublished page. Add non-null values for the published fields, then try saving. You should get the same error message.

Now you can test as an org user. Log in to the dashboard as an org user who's on the free plan. 

Next create one page and publish it. After that, create another page, and try to save a published date for it. You should see a generic error message in the SPA, but if you look in dev tools at the PATCH request to the server, you should see that the server response has error data indicating that the org is at its publish limit.

No

Yes

We may want to consider letting active users know about changes to page limit and published page limit.  

See https://news-revenue-hub.atlassian.net/browse/DEV-3468

No

N/A

[DEV-3328](https://news-revenue-hub.atlassian.net/browse/DEV-3328)

No

- [DEV-3467](https://news-revenue-hub.atlassian.net/browse/DEV-3467)
- [DEV-3468](https://news-revenue-hub.atlassian.net/browse/DEV-3468)

[DEV-3328]: https://news-revenue-hub.atlassian.net/browse/DEV-3328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/pages/serializers.py
```diff
@@ -246,6 +245,35 @@ def validate_page_limit(self, data):
                 {"non_field_errors": [f"Your organization has reached its limit of {pl} page{'s' if pl > 1 else ''}"]}
             )
 
+    def validate_publish_limit(self, data):
+        """Ensure that publishing a page would not push parent org over its publish limit
+
+        NB: publish_limit is not a serializer field, so we have to explicitly call this method from
+        .validate() below.
+        """
+        logger.info("DonationPageFullDetailSerializer.validate_publish_limit called with data: %s", data)
```
njharman, line 254: debug level log?

apps/pages/serializers.py
```diff
@@ -246,6 +245,35 @@ def validate_page_limit(self, data):
...
+        if self.context["request"].method not in ["POST", "PATCH"]:
+            logger.info(
+                "DonationPageFullDetailSerializer.validate_publish_limit returning early because request doesn't attempt to createa or update a page"
```
njharman, line 257: typo "createa or" 

Also debug level log?

apps/pages/serializers.py
```diff
@@ -246,6 +245,35 @@ def validate_page_limit(self, data):
...
+            )
+            return
+        org = self.instance.revenue_program.organization if self.instance else data["revenue_program"].organization
+        offset = 0 if self.instance else 1
+        if DonationPage.objects.filter(revenue_program__organization=org).count() + offset > (
+            pl := org.plan.publish_limit
+        ):
```
njharman, line 264: Can we explain this offset. Or is it obvious to any def who understands these models better than I?

apps/pages/serializers.py
njharman, line 257: Uncovered by test

njharman (changes requested):
1. Communicate to endusers: mentioned page limit changed, do not our user docs need to be updated?
2. Pre deployment step: Are we sure, do we need to check that no existing production plans exceed their limits?

Looks good, minor typo, coverage, doc, log requests.

apps/pages/serializers.py
benjaminEwhite: b0b1009c8307

apps/pages/serializers.py
benjaminEwhite: b0b1009c8307

apps/pages/serializers.py
benjaminEwhite: b0b1009c8307

apps/pages/serializers.py
benjaminEwhite: Removed this path in b0b1009c8307

Validation only happens on post and patch anyways,  so wasn't necessary.

njharman (approved):
Ship it!

x110dc (approved):

njharman (approved):
looks good

apps/pages/serializers.py
```diff
@@ -246,6 +245,32 @@ def validate_page_limit(self, data):
...
+        """Ensure that publishing a page would not push parent org over its publish limit
+
+        NB: publish_limit is not a serializer field, so we have to explicitly call this method from
+        .validate() below.
+        """
+        logger.debug("DonationPageFullDetailSerializer.validate_publish_limit called with data: %s", data)
+        org = self.instance.revenue_program.organization if self.instance else data["revenue_program"].organization
+        # this method gets run both in create and update contexts, so we need to account for the fact that with an offset.
+        # What we're trying to compute is the total number of published pages for the org if the current request was processed
+        offset = 0 if self.instance else 1
```
nrh-cklimas, line 258: @benjaminEwhite @njharman I'm wondering if this line is our problem, and we need to be a little more nuanced about this logic. I think this assumes that the created page will be published (adding 1 to the count of existing published pages in line 261), but that is not actually true in most cases. Does that seem right to you?

Here's the unit test I wrote to try to exercise this:

```python
# Skipping PLUS here because there is no publish limit on that plan.

    @pytest_cases.parametrize("plan", (Plans.FREE.value, Plans.CORE.value))
    def test_can_create_when_at_publish_limit(self, plan, hub_admin_user):
        org = OrganizationFactory(plan_name=plan)
        rp = RevenueProgramFactory(organization=org)
        print(org.plan.name, org.plan.publish_limit, org.plan.page_limit)
        DonationPageFactory.create_batch(org.plan.publish_limit, published_date=timezone.now(), revenue_program=rp)
        serializer = DonationPageFullDetailSerializer(
            data={
                "slug": "my-new-page",
                "revenue_program": rp.pk,
            }
        )
        request = APIRequestFactory().post("/")
        request.user = hub_admin_user
        serializer.context["request"] = request
        assert serializer.is_valid() is True
```

apps/pages/serializers.py
```diff
@@ -246,6 +245,38 @@ def validate_page_limit(self, data):
...
+
+        NB: publish_limit is not a serializer field, so we have to explicitly call this method from
+        .validate() below.
+        """
+        logger.debug("DonationPageFullDetailSerializer.validate_publish_limit called with data: %s", data)
+        org = self.instance.revenue_program.organization if self.instance else data["revenue_program"].organization
+        # this method gets run both in create and update contexts, so we need to account for the fact that with an offset.
+        # What we're trying to compute is the total number of published pages for the org if the current request was processed
+        offset = 0
+        # If the page already exists but is gaining a publish date:
```
nrh-cklimas, line 259: @benjaminEwhite I'm not sure if this logic can be simpler.

apps/pages/serializers.py
```diff
@@ -270,6 +301,8 @@ def validate_sidebar_element_permissions(self, data):
 
     def validate(self, data):
         self.validate_page_limit(data)
+        if "published_date" in data:
```
nrh-cklimas, line 304: @benjaminEwhite not sure if we need to do something here if the object is already published.

apps/pages/serializers.py
```diff
@@ -246,6 +245,38 @@ def validate_page_limit(self, data):
...
+        org = self.instance.revenue_program.organization if self.instance else data["revenue_program"].organization
+        # this method gets run both in create and update contexts, so we need to account for the fact that with an offset.
+        # What we're trying to compute is the total number of published pages for the org if the current request was processed
+        offset = 0
+        # If the page already exists but is gaining a publish date:
+        if self.instance and self.instance.published_date is None and data["published_date"] is not None:
+            offset = 1
+        # If the page is being created with a publish date:
+        if not self.instance and data["published_date"] is not None:
+            offset = 1
```
benjaminEwhite, line 264:
```suggestion
        if any([
            # If the page already exists but is gaining a publish date
            self.instance and self.instance.published_date is None and data["published_date"] is not None,
            # If the page is being created with a publish date:
            not self.instance and data["published_date"] is not None,
        ]):
            offset = 1
        # If the page is being created with a publish date:
        if not self.instance and data["published_date"] is not None:
            offset = 1
```

apps/pages/serializers.py
nrh-cklimas: Two things I wasn't clear on here were...
- In methods like `validate_publish_limit` we're checking both `data` and `self.instance`. Should I do that here too?
- Originally I wrote this logic to prevent the validator from blocking people from creating new pages if the publish limit had been met. I think that may not actually be needed anymore because the publish limit validator should now be smart enough to not block that.

apps/pages/serializers.py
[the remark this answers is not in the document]
nrh-cklimas: Works for me--updated in 16a47fbaa6e7042473a6bc870db552e757b71c24

apps/pages/serializers.py
benjaminEwhite: In `validate`? I don't think we need to do anything with self.instance here.  It's enough to know that published_date appears in the serializer data.

Yeah, looks to me like you can still create new page (if unpublished) when at publish limit. And that's tested in `test_can_create_when_at_publish_limit`, right?

nrh-cklimas: Yup, that's correct re: the unit test. It sounds like we're good to go here.

PR #1080, opened 2023-04-03
DEV-2301: Add page name editing

nrh-cklimas:

n/a

#### What's this PR do?

- Makes the page name editable in the top bar.
- Changes `useEditablePage()` so that it changes the page name to prevent a conflict when saving.
- Adds a constraint to the page model to block duplicate page names in the same revenue program.
  - There is data on staging that fails this constraint. But we don't have any on prod.

Improved UX.

1. Edit a page.
2. Click the pencil icon next to the page name.
3. Confirm the page name becomes editable. Type in a new page name.
4. Click away from the field and confirm the page name is no longer editable and shows the changed value.
5. Save changes and confirm that the page name changes in the pages list.
6. Confirm the page slug was not changed by your edits above.
7. Edit the page again and click the button to edit the page name.
8. Delete everything from the field and click away to save.
9. Confirm the page name reverts to what it used to be.
10. Click the button to edit the page name.
11. Replace the page name with one or more spaces.
12. Confirm that when you click away, the page name reverts to what it used to be.
13. Click the button to edit the page name.
14. Enter text into the field and hit the Enter key.
15. Confirm that that takes the name out of editable mode and that the change is saved.
16. Click the button to edit the page name.
17. Enter text into the field and hit the Escape key.
18. Confirm that the page name reverts back to what it used to be.
19. Click the button to edit the page name.
20. Change the page name and click away to save changes.
21. Click the back button in the editor.
22. Confirm you see a "Are you sure? You have pending changes" prompt.
23. Click to continue.
24. Confirm that the page name didn't change.

Yes, we need to document the process for renaming pages.

Yes.

Updated docs.

I didn't put tests on the Django model. I talked with @benjaminEwhite about it and his feeling was at best we could test that the constraint exists, but it felt to me that it was testing config instead of functionality. Not sure how others feel about that.

I also moved the constraint to use [UniqueConstraint](https://docs.djangoproject.com/en/4.1/ref/models/constraints/#uniqueconstraint) because it seems like it is current Django best practice, but that means that we lose the custom error message we had before. I've read Django 4.1 would give us that functionality, but we're not there yet.

Not yet.

https://news-revenue-hub.atlassian.net/browse/DEV-2301

No.

No.


spa/src/components/pageEditor/PageName/PageName.tsx
```diff
@@ -0,0 +1,85 @@
...
+  function handleFinishEditing(event?: FormEvent) {
+    if (editValue.trim() !== '' && editValue !== updatedPagePreview?.name) {
+      setPageChanges((changes) => ({ ...changes, name: editValue }));
+    }
+
+    setEditing(false);
+
+    if (event) {
+      event.preventDefault();
+    }
```
GuiMend, line 22: Shouldn't this always be called the first thing in the function? Never seen it like this.

GuiMend (approved):

spa/src/components/pageEditor/PageName/PageName.tsx
nrh-cklimas: It might be a code style thing--I haven't seen this specifically, but I could see the usefulness of doing so to make sure you always prevent the default regardless of what happens in the rest of the function.

GuiMend: It's up to you if you'd like to leave it as is or change it to the suggestion. 
I'm good either way 👍

spa/src/components/pageEditor/PageName/PageName.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+import PageName from './PageName';
+import { useEditablePageContext } from 'hooks/useEditablePage';
+
+jest.mock('hooks/useEditablePage');
+
+function tree() {
+  return render(<PageName />);
+}
+
+describe('PageName', () => {
```
GuiMend, line 12: Didn't find a test asserting the tooltip is being used.

GuiMend (approved):
Added one comment about adding a test, but is not required

spa/src/components/pageEditor/PageName/PageName.test.tsx
nrh-cklimas: Good callout--aeb83a090f758fe356dbb8101f0494ad50d91a57

PR #1084, opened 2023-04-05
DEV-3441-update-stripe-methods-that-send-customer-address-to-include-line-2-field

GuiMend:

This is the first of 2 stacked PRs:
1. This one
2. https://github.com/newsrevenuehub/rev-engine/pull/1085

#### What's this PR do?
- Adds line2 support for addresses in stripe connection

So that users can add complementary information on address (apartment, suite, etc).

I don't know how to test this PR by itself. When the FE ticket is ready, we will be able to add line2 information to where we store addressed (inside stripe)

no

yes

n/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-3441

no

FE ticket: https://news-revenue-hub.atlassian.net/browse/DEV-1657


apps/contributions/models.py
```diff
@@ -71,6 +71,7 @@ def create_stripe_customer(
         customer_name=None,
         phone=None,
         street=None,
+        complement=None,
```
njharman, line 74: Might be worth a comment along the lines of # apt, room, aka "line2".

Huh, I've not encountered this name for this information before. Googling it seems to be more popular in non-USA parts of world. I'm not against it. Better than something like "address_line2" which I typically see.

njharman (approved):
Small, looks good. I would prefer that at least one test that things still work with blank/no address complement. I didn't see one.

I'm pre-approving, rather than force another rereview as the comment suggestion and above are such small requests (and not "errors").

GuiMend:
@njharman , I tried writing tests using:
```
@patch("stripe.Customer.create", return_value={"id": "cus_fakefakefake"})
@pytest.mark.parametrize("mailing_complement", ("AP 1", None))
def test_create_stripe_customer_empty_complement(self, mock_create_customer, *args):
```
But I wasn't able to get the "mailing_complement" argument from `args`. Do you know what I could have done wrong? Seems identical to the other pytests that use `parametrize`

njharman:
> @njharman , I tried writing tests using:
> 
> ```
> @patch("stripe.Customer.create", return_value={"id": "cus_fakefakefake"})
> @pytest.mark.parametrize("mailing_complement", ("AP 1", None))
> def test_create_stripe_customer_empty_complement(self, mock_create_customer, *args):
> ```
> 
> But I wasn't able to get the "mailing_complement" argument from `args`. Do you know what I could have done wrong? Seems identical to the other pytests that use `parametrize`

I would have to look up to remember correctly, two things come to mind.


1. the order of patch/pytest and arguments, I thought it was "inside out" or lowest first. that is parametize would be put in mock_create_customer, but now that I write it out that doesn't seem right. pytest would complain names don't match (I think)
2. I've never seen and would not reccomend using *args for paramtized values. I'm surprized pytest didn't complain. Use named arg
`def ...(self, mock_creat_customer, mailing_complement, ...)`
3. 
Bonus third, I've had issues with some decorators not returning the actual function (at least not the function that pytest wants to muck around with).  Either just error or pytest decorates the wrong callable.

njharman (approved):

PR #1093, opened 2023-04-10
DEV-3442: Configure address editor block

GuiMend:

n/a

#### What's this PR do?
- Make address fields optional on BE so that it doesn't block a donation that doesn't have address information
- Continue transcribe of JS to TS
- Add additional tests to cover updated TS files
- Remove duplicate mock file
- Adds the following to address editor block:
  - Address optionality
  - Show only Zip and Country fields

Increase flexibility on address information when it's not required

Edit address, make a contribution and see that the flow works unchanged:

1) Check address optional fields
- Make sure contribution no longer requires any address fields to move forward 

2) Check Optional and go back to Required:
- Make sure it keeps working as expected

3) Make no changes
- Default/legacy address editor configurations should keep working

4) Select only Zip and Country code + Optional
- Make sure only these 2 fields show up and they are optional

5) Select only Zip and Country code + Required
- Make sure only these 2 fields show up and they are required

6) Select only zip + country
- Make sure State label configuration hides
 
Yes

yes

Addresses for contribution now are configurable and more flexible to fit your needs.

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-3442

no

https://news-revenue-hub.atlassian.net/browse/DEV-3443

Follow-up ticket to solve issue with multiple batch updates on 1 render:
https://news-revenue-hub.atlassian.net/browse/DEV-3506


spa/src/components/donationPage/pageContent/DDonorAddress.test.tsx
```diff
@@ -107,11 +107,28 @@ describe('DDonorAddress', () => {
...
       expect(screen.getByRole('textbox', { name: visibleName })).toBeRequired();
     });
 
+    it('is not required if addressOptional is true', () => {
+      tree({}, { element: { ...element, content: { addressOptional: true } } });
+      expect(screen.getByRole('textbox', { name: visibleName })).not.toBeRequired();
+    });
+
+    it(`if zipAndCountryOnly = true -> ${showZipAndCountryOnly ? 'show' : 'hide'}`, () => {
+      tree({}, { element: { ...element, content: { zipAndCountryOnly: true } } });
```
nrh-cklimas, line 121: Let's add an `expect.assertions(1);` so that if something gets garbled in the conditional logic below, this test fails.

nrh-cklimas (changes requested):

spa/src/components/pageEditor/elementEditors/donorAddress/DonorAddressEditor.tsx
```diff
@@ -34,38 +36,81 @@ export function DonorAddressEditor({ elementContent, onChangeElementContent }: D
...
-        ></FormControlLabel>
-        <FormControlLabel
-          control={
-            <Checkbox
-              checked={elementContent.additionalStateFieldLabels?.includes('region')}
-              onChange={(event) => handleCheckboxChange('region', event)}
+            <FormControlLabel
+              control={
+                <Checkbox
+                  checked={!!elementContent.additionalStateFieldLabels?.includes('region')}
```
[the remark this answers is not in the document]
GuiMend: Same here

spa/src/components/pageEditor/elementEditors/donorAddress/DonorAddressEditor.tsx
```diff
@@ -34,38 +36,81 @@ export function DonorAddressEditor({ elementContent, onChangeElementContent }: D
...
+              label={
+                <>
+                  State <span style={{ fontStyle: 'italic' }}>(required)</span>
+                </>
+              }
+            />
+            <FormControlLabel
+              control={
+                <Checkbox
+                  checked={!!elementContent.additionalStateFieldLabels?.includes('province')}
```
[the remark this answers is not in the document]
nrh-cklimas: Ah, right.

nrh-cklimas (approved):

apps/contributions/serializers.py
```diff
@@ -145,11 +145,11 @@ class BadActorSerializer(serializers.Serializer):
     first_name = serializers.CharField(max_length=40)
     last_name = serializers.CharField(max_length=80)
     email = serializers.EmailField(max_length=80)
-    street = serializers.CharField(max_length=255)
-    city = serializers.CharField(max_length=40)
-    state = serializers.CharField(max_length=80)
-    country = serializers.CharField(max_length=80)
-    zipcode = serializers.CharField(max_length=20)
+    street = serializers.CharField(max_length=255, required=False, default="", allow_blank=True)
```
nrh-cklimas, line 148: I _think_, but ask a real BE dev, that `allow_blank=True` implies `required=False`.

spa/src/components/donationPage/DonationPage.js
```diff
@@ -199,12 +199,12 @@ function DonationPage({ page, live = false }) {
       email: data.email,
       phone: data.phone || '', // stripe will complain if its null or undefined, and it's an optional field
       address: {
-        city: data.mailing_city,
-        country: data.mailing_country,
-        line1: data.mailing_street,
+        city: data.mailing_city ?? '',
```
nrh-cklimas, line 202: This feels like it's worth a comment about Stripe allowing empty strings but not `undefined`, since that was confusing initially.

spa/src/components/pageEditor/elementEditors/donorAddress/DonorAddressEditor.tsx
```diff
@@ -1,21 +1,58 @@
...
 }
 
-export function DonorAddressEditor({ elementContent, onChangeElementContent }: DonorAddressEditorProps) {
+export function DonorAddressEditor({
+  elementContent,
+  onChangeElementContent,
+  onChangeElementRequiredFields
+}: DonorAddressEditorProps) {
   const zipAndCountryOnly = !!elementContent.zipAndCountryOnly;
+  const prevZipAndCountryOnly = usePreviousState(zipAndCountryOnly);
```
nrh-cklimas, line 30: Instead of using `usePreviousState` and the effect below, let's call `handleChangeRequiredFields` from the event handlers below (if I'm understanding correctly, the effect you have should only fire in response to changes in the controls below). The new React docs have a [good explanation of best practice here](https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers).

nrh-cklimas (changes requested):
A couple small suggested changes.

spa/src/components/pageEditor/elementEditors/donorAddress/DonorAddressEditor.tsx
GuiMend: I tried doing that initially @nrh-cklimas , but if we call `onChangeElementContent` and `onChangeElementRequiredFields` it only runs the first function called. The second one is "skipped", that's why I went with this solution.

In the function `addBatchChange` which is responsible for doing the update, we have `setBatchChanges((existing) => ({ ...existing, ...value }))`, so it should keep the "previousState" preserved, which would mean that running both function together is possible. But the issue is that we want to save the `elements` prop, that is an array inside of `value`, and not `existing`, so the previous state isn't preserved as we want it to.

GuiMend: For that to work, we would have to refactor `addBatchChange` function, which could cause some side effects as it's a really important function. (Not sure we would want to go there in this ticket).

One additional "benefit" from the side effect that you pointed out on the reference link, is that in our case it's good to have the function be called on the refresh because the default "requiredFields" is an empty array, and when the function runs because of the side effect it updates the "requiredFields" to the correct value.

@nrh-cklimas , do we have a way of setting a default "requiredFields"? I didn't find it looking around.

nrh-cklimas: I created DEV-3506 as a followup to fix the main problem you've described here. Could you put a comment here with that story name so we can follow up on it there?

We need to address the second part of your comment though in this PR if I'm not mistaken--a user could add an Address block, never edit it, and the resulting contribution page would be in an incorrect state with required fields. (Let me know if I have that wrong.)

[AddElementModal](https://github.com/newsrevenuehub/rev-engine/blob/main/spa/src/components/pageEditor/editInterface/pageElements/addElementModal/AddElementModal.js#L22) is where new elements are added to the page. We have functionality there for setting default content and we'd want to expand that aspect of it to include default required fields. I think now would be a great time to refactor that into a hook that we can get good unit test coverage on, because complexity is starting to get high on the component and we have no unit tests on it now. Don't hesitate to follow up with me on Slack on how we can approach it.

GuiMend:
@nrh-cklimas, I talked with @benjaminEwhite about the address validation issues and not having the "required" validation for legacy donation pages for a while until we fix the entire address data structure on https://news-revenue-hub.atlassian.net/browse/DEV-3440 is acceptable as we already have the FE validation.

spa/src/components/pageEditor/editInterface/pageElements/addElementModal/AddElementModal.js
```diff
@@ -45,6 +49,7 @@ function AddElementModal({ addElementModalOpen, setAddElementModalOpen, destinat
     return {
       uuid: uuidv4(),
       content: defaultContent[type] ? { ...defaultContent[type] } : undefined,
+      ...(defaultRequiredFields[type] && { requiredFields: defaultRequiredFields[type] }),
```
nrh-cklimas, line 52: Could we make it parallel how `content` is set above it? I think it achieves the same purpose as the line above it, so let's make it look the same if we can.

Is it possible to add tests, either in Cypress or Jest, to cover this new behavior?

benjaminEwhite (approved):

nrh-cklimas (changes requested):

spa/src/components/pageEditor/editInterface/pageElements/addElementModal/AddElementModal.tsx
```diff
@@ -66,12 +79,12 @@ function AddElementModal({ addElementModalOpen, setAddElementModalOpen, destinat
...
     const dynamicElements = destination === 'sidebar' ? permittedSidebarElements : permittedPageElements;
     return Object.keys(dynamicElements).map((elName, i) => {
-      const element = dynamicElements[elName];
+      const element = dynamicElements[Number(elName)];
       const els = destination === 'sidebar' ? sidebarElements : elements;
       // An element is disabled if it's unique and already present.
       const disabled = element.unique && els?.some((el) => el.type === element.type);
       return (
-        <S.PageItemWrapper key={element.displayName + i}>
+        <S.PageItemWrapper key={`${element?.displayName}${i}`}>
```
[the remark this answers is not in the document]
GuiMend: It's because `displayName` is `?` inside ContributionPageElement TS
```js
export interface ContributionPageElement {
 ...
  /**
   * User-visible name of the element.
   */
  displayName?: string;
```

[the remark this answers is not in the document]
nrh-cklimas: If that's the case, maybe we want something like

```tsx
<S.PageItemWrapper key={`${element.displayName ?? ''}${i}`}></code>
```

Or just leave it without the `?`, since that's operating on `element`, not `element.displayName`.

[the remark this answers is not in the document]
GuiMend: What about just affirming that it won't be undefined/null? It should always have a displayName in this part of the code.
`element.displayName!`

nrh-cklimas (approved):

spa/src/components/base/Radio/Radio.stories.tsx
```diff
@@ -30,7 +30,10 @@ interface TemplateProps {
 }
 
 const Template: Story<TemplateProps> = (props: TemplateProps) => (
-  <FormControlLabel control={<Radio checked={props.checked} />} disabled={props.disabled} label="Radio Label" />
+  <>
+    <FormControlLabel control={<Radio checked={props.checked} />} disabled={props.disabled} label="Radio Label" />
+    <FormControlLabel control={<Radio checked={props.checked} />} disabled={true} label="Disabled Radio Label" />
```
nrh-cklimas, line 35:
```suggestion
    <FormControlLabel control={<Radio checked={props.checked} />} disabled label="Disabled Radio Label" />
```

spa/src/components/pageEditor/elementEditors/donorAddress/DonorAddressEditor.styled.ts
```diff
@@ -25,9 +26,16 @@ export const RadioGroup = styled(BaseRadioGroup)`
 `;
 
 export const Text = styled.p`
+  color: ${({ theme }) => theme.colors.muiGrey[900]};
```
nrh-cklimas, line 29:
```suggestion
  color: ${({ theme }) => theme.basePalette.greyscale.black};
```

spa/src/components/pageEditor/elementEditors/donorAddress/DonorAddressEditor.styled.ts
```diff
@@ -25,9 +26,16 @@ export const RadioGroup = styled(BaseRadioGroup)`
...
 
 export const StyledFormControlLabel = styled(FormControlLabel)`
-  margin: 24px 0 15px;
+  margin-top: 40px;
+`;
+
+export const Disclaimer = styled.p`
+  margin-top: 20px;
+  font-size: ${({ theme }) => theme.fontSizesUpdated.sm};
+  color: ${({ theme }) => theme.colors.muiGrey[600]};
```
nrh-cklimas, line 40:
```suggestion
  color: ${({ theme }) => theme.basePalette.greyscale.grey1};
```

spa/src/components/pageEditor/elementEditors/donorAddress/DonorAddressEditor.styled.ts
nrh-cklimas: Just noting the updates on the changed lines--looks like there are other colors in here that could be changed, but up to you if you want to tackle that here.

nrh-cklimas (changes requested):

PR #1096, opened 2023-04-13
DEV-2884: make Apple Pay domain verification a background job after s…

leo-dcfa:
…tripe oauth succcess

#### What's this PR do?

- Adds automation to verify apple domain as a celery task after successful stripe oauth callback

We expect the Apple Pay domain to be created as part of the onboarding process

Cannot be tested manually outside of production, see unit tests for details

N/A

Yes

N/A

N/A

N/A

[DEV-2884]

N/A

N/A

[DEV-2884]: https://news-revenue-hub.atlassian.net/browse/DEV-2884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/tasks.py
```diff
@@ -118,3 +120,21 @@ def task_pull_payment_intents(self, email_id, customers_query, stripe_account_id
...
+
+
+@shared_task(bind=True, autoretry_for=(RateLimitError,), retry_backoff=True, retry_kwargs={"max_retries": 3})
+def task_verify_apple_domain(self, revenue_program_slug: str):
+    logger.info("[task_verify_apple_domain] called with revenue_program_slug %s.", revenue_program_slug)
+    revenue_program = RevenueProgram.objects.get(slug=revenue_program_slug)
+    try:
+        revenue_program.stripe_create_apple_pay_domain()
+        logger.info(
+            "[task_verify_apple_domain] Apple Pay domain verified for revenue program %s with slug %s",
```
benjaminEwhite, line 132: We should standardize how we offset function/method names. I like this style you're using here. For a while I was doing ``foo`` but recently have just been doing `foo`.  Square brackets seem like a good way to go.

apps/contributions/tasks.py
```diff
@@ -118,3 +120,21 @@ def task_pull_payment_intents(self, email_id, customers_query, stripe_account_id
...
+            revenue_program,
+            revenue_program_slug,
+        )
+    except stripe.error.StripeError as ex:
+        logger.exception(
+            "[task_verify_apple_domain] task failed for slug %s due to exception: %s", revenue_program_slug, ex.error
```
benjaminEwhite, line 138: Would we get `ex.error` anyways just by virtue of exception being logged via .exception?

apps/contributions/tests/test_tasks.py
```diff
@@ -89,3 +92,19 @@ def test_task_pull_payment_intents(self, sub_cache_mock, contrib_cache_mock, fet
...
+    with patch("apps.contributions.tasks.RevenueProgram.objects.get") as mock_rp_get:
+        mock_rp_get.return_value = mock_revenue_program
+        if should_raise_stripe_error:
+            mock_revenue_program.stripe_create_apple_pay_domain.side_effect = stripe.error.StripeError()
+            with pytest.raises(stripe.error.StripeError):
+                task_verify_apple_domain(slug)
+        else:
+            task_verify_apple_domain(slug)
+        assert mock_revenue_program.stripe_create_apple_pay_domain.called
+        mock_rp_get.assert_called_with(slug=slug)
```
benjaminEwhite, line 110: It's a little hard to follow what this test is doing and why. Would be a lot cleaner, IMO, to just have a `test_task_verify_apple_domain_happy_path` and a `test_task_verify_apple_domain_when_stripe_error`.   Also, not sure why we're mocking out `RevenueProgram.get`. We should ensure there's an rp with the expected slug in the db, then just test that calling the task with that slug causes it to be retrieved etc.

apps/contributions/views.py
```diff
@@ -79,23 +82,33 @@ def stripe_oauth(request):
     revenue_program = RevenueProgram.objects.get(id=revenue_program_id)
 
     try:
+        logger.info("Attempting to perform oauth with Stripe for revenue program %s", revenue_program_id)
```
benjaminEwhite, line 85: Prefix the log message with `[stripe_oauth]`

apps/contributions/views.py
```diff
@@ -79,23 +82,33 @@ def stripe_oauth(request):
...
         oauth_response = stripe.OAuth.token(
             grant_type="authorization_code",
             code=code,
         )
         payment_provider = revenue_program.payment_provider
         if not payment_provider:
+            logger.info(
+                "Payment provider not yet created for revenue program %s, proceeding with creation...",
```
benjaminEwhite, line 93: Prefix

apps/contributions/views.py
```diff
@@ -79,23 +82,33 @@ def stripe_oauth(request):
...
+                revenue_program_id,
+            )
             payment_provider = PaymentProvider.objects.create(
                 stripe_account_id=oauth_response["stripe_user_id"],
                 stripe_oauth_refresh_token=oauth_response["refresh_token"],
             )
+            logger.info("Payment provider for revenue program %s created: %s", revenue_program_id, payment_provider.id)
```
benjaminEwhite, line 100: Prefix

apps/contributions/views.py
```diff
@@ -79,23 +82,33 @@ def stripe_oauth(request):
...
             revenue_program.payment_provider = payment_provider
             revenue_program.save()
         else:
+            logger.info(
+                "Updating existing payment provider %s for revenue program %s", payment_provider.id, revenue_program_id
```
benjaminEwhite, line 105: And here -- not going to make this comment again, but apply it anywhere else relevant.

apps/contributions/views.py
```diff
@@ -79,23 +82,33 @@ def stripe_oauth(request):
...
+            )
             payment_provider.stripe_account_id = oauth_response["stripe_user_id"]
             payment_provider.stripe_oauth_refresh_token = oauth_response["refresh_token"]
             payment_provider.save()
-
+        logger.info("Delaying Apple Pay domain verification for revenue program slug: %s", revenue_program.slug)
```
benjaminEwhite, line 110: We're not really delaying anything, just calling an async task.

leo-dcfa: good point. Badly named methods lead to bad assumptions 😂

apps/contributions/tasks.py
leo-dcfa: I tried it, but the linter complained so I changed it to `ex.error`

apps/contributions/tasks.py
leo-dcfa: nice - I wonder if we could achieve the same using some logging configuration, instead of doing it manually on each log statement

apps/contributions/tests/test_tasks.py
leo-dcfa: It's a good point, I'll separate into two methods. It's also the standard you establish in rev engine, so it makes sense

Concerning mock, depends how you see this - having a record on a db is closer to an integration than a unit test. We technically don't care about the result of mock_rep_get; we just care that the method is called with the correct parameters.

apps/contributions/views.py
leo-dcfa: fixed

apps/contributions/views.py
leo-dcfa: Fixed

apps/contributions/views.py
leo-dcfa: Fixed

apps/contributions/views.py
leo-dcfa: Fixed

apps/contributions/tasks.py
benjaminEwhite: Yeah, probably configure the logging to automatically prepend the function name with desired formatting. Think we could do something with a decorator to achieve this:

https://ankitbko.github.io/blog/2021/04/logging-in-python/

And looks like we could use `inspect` to derive function names to pass to the logger: https://stackoverflow.com/questions/5067604/determine-function-name-from-within-that-function-without-using-traceback

Not suggesting that we do that in this ticket, but it would be a worthwhile thing to document and standardize in our code base.

apps/contributions/tasks.py
benjaminEwhite: Oh, more importantly, you're re-raising on next line, which makes my original comment irrelevant.

apps/contributions/views.py
```diff
@@ -79,25 +82,45 @@ def stripe_oauth(request):
...
                 stripe_account_id=oauth_response["stripe_user_id"],
                 stripe_oauth_refresh_token=oauth_response["refresh_token"],
             )
+            logger.info(
+                "[stripe_oauth] Payment provider for revenue program %s created: %s",
+                revenue_program_id,
+                payment_provider.id,
+            )
             revenue_program.payment_provider = payment_provider
             revenue_program.save()
```
benjaminEwhite, line 106: As long as you're touching contiguous code, I'd like us to start being disciplined about using `update_fields` when we save model instances because it tends to protect against a certain class of race conditions (that stem from saving an instance's local state + some single changed attribute without reference to global state change representing the instance).  DEV-3073 (see [here](https://github.com/newsrevenuehub/rev-engine/pull/927)) standardizes this pattern for views touching the contribution model.

Additionally, in that ticket/PR, I added testing that proves that the underlying model instance gets saved at most _once_ per request session, and with the expected update fields, and that a django reversion revision gets created. Together, these comprise a guarantee that changes to a given model are deliberate (via update_fields), predictable ( via guarantee on number of saves), and auditable (via revisions being created that will provide a clear trail of when model changed and how/why).

benjaminEwhite (changes requested):
@leo-dcfa this looks good overall. Have one suggestion for a follow up ticket. And then have a suggestion around using `update_fields` in the calls to `.save()` in `apps.contributions.views.py`.  Ideally, we'd adopt a similar pattern as in #927 whereby we have a test that proves that the relevant model instance(s) get(s) saved an expected number of times (ideally only once) per request session, and that `update_fields` is called with right values when saved, and that a revision is created.  

In case it's helpful, you can see the pattern I used for creating an `update_data` set at the top of view functions that gets fields dynamically added. i mutate the instance state and add the mutated property name to `update_data`, then at the end of the request cycle I save the instance if there are any fields in `update_data`. When I save, i set `update_fields` to the value of `update_data` plus the keys for `created` and `modified`, as they are autoadded.

I know you're about to be off so I'd be fine with the less ambitious version here of simply using `update_fields` without refactoring tests right now to guarantee this.

apps/contributions/views.py
leo-dcfa: Is this part of scope of current ticket though? I just added a few logging statements here

apps/contributions/tasks.py
leo-dcfa: agree it belongs in a separate ticket

leo-dcfa: New ticket created https://news-revenue-hub.atlassian.net/browse/DEV-3520

benjaminEwhite (approved):

coverage.xml
```diff
@@ -0,0 +1,4906 @@
+<?xml version="1.0" ?>
+<coverage version="7.2.1" timestamp="1681767534851" lines-valid="4412" lines-covered="1474" line-rate="0.3341" branches-valid="908" branches-covered="79" branch-rate="0.087" complexity="0">
```
x110dc, line 2: This file shouldn't be committed.

x110dc (changes requested):

coverage.xml
leo-dcfa: good catch!

x110dc (approved):

PR #1111, opened 2023-05-01
DEV-3521: Fix scheduled payment email header

nrh-cklimas:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

n/a

#### What's this PR do?

- Fixes alignment issue with scheduled payment reminder emails
- Adds a debug-only route to the Django app, `/__debug_email__/recurring-contribution/` that allow previewing emails.

#### Why are we doing this? How does it help us?

- Fixes a display bug
- Makes it easier to work on styled emails

#### How should this be manually tested? Please include detailed step-by-step instructions.

To test styled emails for regressions:

1. Set up a recurring contribution to an org whose default page has a style set.
2. Send a reminder email for this contribution.
3. Confirm the email appearance is correct (header color and logo are correct, appear above text header) and contains accurate info.
4. Set up a recurring contribution to an org whose default page has a style set, but no header logo.
2. Send a reminder email for this contribution.
3. Confirm the email appearance is correct (has left-aligned NRE logo at top, no background color in header) and contains accurate info.

To test unstyled emails for regressions:

1. Set up a recurring contribution to an org whose default page doesn't have a style set.
2. Send a reminder email for this contribution.
3. Confirm the email appearance is correct (has left-aligned NRE logo at top) and contains accurate info.

To test free plan emails for regressions:

1. Set up a recurring contribution to an org whose default page has a style set, but whose org is on the free plan.
2. Send a reminder email for this contribution.
3. Confirm the email appearance is correct (does not use custom style, has left-aligned NRE logo at top) and contains accurate info.

#### Did this PR make changes to the user interface that may require changes to the user-facing docs?

No.

#### Have automated unit tests been added? If not, why?

I updated existing Python tests.

#### How should this change be communicated to end users?

n/a

#### Are there any smells or added technical debt to note?

No.

#### Has this been documented? If so, where?

 I also added notes to the readme explaining how the debug email routes work.

#### What are the relevant tickets? Add a link to any relevant ones.

https://news-revenue-hub.atlassian.net/browse/DEV-3521

#### Do any changes need to be made before deployment to production (adding environment variables, for example)?

No.

#### Are there next steps? If so, what? Have tickets been opened for them? List next-step tickets here:

No.

PR #1118, opened 2023-05-05
DEV-3519: add header bar to email receipts

nrh-cklimas:

n/a

#### What's this PR do?

- Syncs up header bar in contribution receipts with the appearance on recurring contribution reminders (https://github.com/newsrevenuehub/rev-engine/pull/1111).
- Adds debug URL for contribution receipts.

- Fixes appearance for logos that are meant for a dark background.
- Makes email dev work easier.

To test styled emails for regressions:

1. Make a contribution to an org whose default page has a style set.
2. Confirm the receipt email appearance is correct (header color and logo are correct, appear above text header) and contains accurate info.
3. Make a contribution to an org whose default page has a style set, but no header logo.
4. Confirm the receipt email appearance is correct (has left-aligned NRE logo at top, no background color in header) and contains accurate info.

To test unstyled emails for regressions:

1. Make a contribution to an org whose default page doesn't have a style set.
2. Confirm the receipt email appearance is correct (has left-aligned NRE logo at top) and contains accurate info.

To test free plan emails for regressions:

1. Make a contribution to an org whose default page has a style set, but whose org is on the free plan.
2. Confirm the receipt email appearance is correct (does not use custom style, has left-aligned NRE logo at top) and contains accurate info.

No.

Updated existing tests.

n/a

No.

Updated the readme.

https://news-revenue-hub.atlassian.net/browse/DEV-3519

No.

No.

### DIFF
--- a/apps/emails/tests/test_tasks.py
+++ b/apps/emails/tests/test_tasks.py
@@ -133,6 +133,7 @@ class TestSendThankYouEmail:
                     },
                     "font": {"heading": "mock-header-font", "body": "mock-body-font"},
                 }
+                style.save()
                 page = DonationPageFactory(revenue_program=revenue_program, styles=style, header_logo="mock-logo")
                 revenue_program.default_donation_page = page
                 revenue_program.save()
@@ -143,17 +144,17 @@ class TestSendThankYouEmail:
 
         default_logo = os.path.join(settings.SITE_URL, "static", "nre-logo-yellow.png")
         custom_logo = 'src="/media/mock-logo"'
-        custom_header_background = "header-background: #mock-header-background !important"
-        custom_button_background = "header-background: #mock-button-color !important"
+        custom_header_background = "background: #mock-header-background !important"
+        custom_button_background = "background: #mock-button-color !important"
 
         if revenue_program.organization.plan.name == FreePlan.name or not default_style:
             expect_present = default_logo
             expect_missing = (custom_logo, custom_button_background, custom_header_background)
 
         else:
-            expect_present = (custom_logo,)
-            # Email template doesn't have a button to apply the custom button color to and also doesn't have a header background to customize
-            expect_missing = (custom_button_background, custom_header_background, default_logo)
+            expect_present = (custom_logo, custom_header_background)
+            # Email template doesn't have a button to apply the custom button color to
+            expect_missing = (custom_button_background, default_logo)
 
         for x in expect_present:
             assert x in mail.outbox[0].alternatives[0][0]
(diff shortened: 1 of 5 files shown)


apps/emails/tests/test_tasks.py
```diff
@@ -143,8 +143,8 @@ def test_contribution_confirmation_email_style(
 
         default_logo = os.path.join(settings.SITE_URL, "static", "nre-logo-yellow.png")
         custom_logo = 'src="/media/mock-logo"'
-        custom_header_background = "background: #mock-header-background !important"
-        custom_button_background = "background: #mock-button-color !important"
+        custom_header_background = "header-background: #mock-header-background !important"
+        custom_button_background = "header-background: #mock-button-color !important"
```
GuiMend, line 147: @nrh-cklimas this change doesn't make sense. `background` is the css atribute, so `header-background` isn't a valid test option. 

I believe what you want to do is this:
image

The tests aren't failing with this update because we are actually testing that those custom colors aren't being applied to the email (which is True as they don't exist)

GuiMend (changes requested):
Need a fix in the python tests

apps/emails/tests/test_tasks.py
nrh-cklimas: @GuiMend yikes--that makes sense. My brain read `background` as a class name which of course it's not. Thanks for the suggested fix. Only thing I noticed is, it doesn't seem like `style.save()` is needed on line 136? The tests pass without it but I'm wondering if there's more to it that I'm not understanding.

GuiMend: It's not needed for this test, but I've added it for "context consistency" (not sure it's named this way).

If we don't save the `style` there, the custom style set in the lines previous are only accessible in this current context. So if we in the future modify the email send function and it happens to get the data from another context instead of getting it passed in as props in line `143` we don't get any weird error because we didn't save the style.

I faced this issue in a another PR I worked on and it was really difficult to find this issue (at least for someone without a lot of BE experience).

nrh-cklimas: Thanks for the explanation! Updated in 51c947cfce0450f042c609e1b1edcedacec260df

apps/emails/tests/test_tasks.py
```diff
@@ -151,9 +152,9 @@ def test_contribution_confirmation_email_style(
             expect_missing = (custom_logo, custom_button_background, custom_header_background)
 
         else:
-            expect_present = (custom_logo,)
+            expect_present = (custom_logo, custom_header_background)
             # Email template doesn't have a button to apply the custom button color to and also doesn't have a header background to customize
```
GuiMend, line 156: @nrh-cklimas could you also update the comment?
```suggestion
            # Email template doesn't have a button to apply the custom button color to
```

GuiMend (approved):
👍

apps/emails/tests/test_tasks.py
nrh-cklimas: 316ea5c1f1de1e273fa01f9b4a5a06cb977c91e1

PR #1128, opened 2023-05-18
[draft] DEV-3308: send test email 

GuiMend:

This is a draft PR to check with other BE engineers if the approach is good.
Also implements a "send test email" button in the styles modal to be able to test the new feature


apps/organizations/views.py
```diff
@@ -223,6 +224,52 @@ def handle_stripe_account_link(request, rp_pk):
...
+def handle_send_test_email(request):
+    """This endpoint sends test emails to the user so that they can see what it looks like.
+    Available email types are:
+
+    1. receipt: this is the email that is sent to the user when they make a payment
+    2. reminder: this is the email that is sent to the user reminding of a future payment
+    3. magic_link: this is the email that is sent to the user when they click on the magic link to log in the contributor portal
+    """
+    rp_pk = request.data["revenue_program"]
+    email_type = request.data["email_type"]
```
nrh-cklimas, line 240: This should be `email_name` according to AC 1b.

apps/organizations/views.py
```diff
@@ -223,6 +224,52 @@ def handle_stripe_account_link(request, rp_pk):
     return Response(data)
 
 
+@create_revision()
```
benjaminEwhite, line 227: Shouldn't need this. Only need if we're saving changes to a model instance, and in this case we're not doing that.

apps/organizations/views.py
```diff
@@ -223,6 +224,52 @@ def handle_stripe_account_link(request, rp_pk):
...
+
+    1. receipt: this is the email that is sent to the user when they make a payment
+    2. reminder: this is the email that is sent to the user reminding of a future payment
+    3. magic_link: this is the email that is sent to the user when they click on the magic link to log in the contributor portal
+    """
+    rp_pk = request.data["revenue_program"]
+    email_type = request.data["email_type"]
+
+    revenue_program = get_object_or_404(RevenueProgram, pk=rp_pk)
+    if not request.user.is_superuser:
```
benjaminEwhite, line 243: This may already be accounted for by the code comment about setting permissions correctly, but believe this should be able to use the permission_classes decorator plus permissions in apps/api/permissions.py to accomplish this.

apps/organizations/views.py
```diff
@@ -223,6 +224,52 @@ def handle_stripe_account_link(request, rp_pk):
...
+
+    if email_type == "receipt":
+        data = make_send_test_thank_you_email_data(request.user, revenue_program)
+        send_thank_you_email.delay(data)
+    elif email_type == "reminder":
+        # TODO: send reminder email
+        data = ""
+    elif email_type == "magic_link":
+        # TODO: send magic link email
+        data = ""
```
benjaminEwhite, line 268: Could use a `match` statement here.
```suggestion
    match email_type:
        case "receipt":
            data = make_send_test_thank_you_email_data(request.user, revenue_program)
            send_thank_you_email.delay(data)
            .... etc.
```

benjaminEwhite (commented):
@GuiMend this is in right direction. Left some comments and suggestions.

apps/organizations/views.py
GuiMend: Not sure we can, didn't find a permission class to handle this scenario. 
What I wanted to do here is only check is user has a role_type that can access RP's if the user isn't a superuser. If you have any suggestions on how to do this I'd like to know 😄

benjaminEwhite (changes requested):
Left some comments/suggestions.

benjaminEwhite (commented):
@GuiMend there are still some unresolved conversations from last review:

- https://github.com/newsrevenuehub/rev-engine/pull/1128/files#r1206886235
- https://github.com/newsrevenuehub/rev-engine/pull/1128/files#r1207048203

apps/emails/tasks.py
```diff
@@ -124,14 +133,77 @@ def make_send_thank_you_email_data(contribution) -> SendThankYouEmailData:
...
+
+    return SendMagicLinkEmailData(
+        magic_link=get_test_magic_link(user, revenue_program),
+        email=user.email,
+        rp_name=revenue_program.name,
+        style=asdict(revenue_program.transactional_email_style),
+    )
+
+
+def get_test_magic_link(user, revenue_program) -> str:
```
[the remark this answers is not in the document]
GuiMend: Done in [b2c0628](https://github.com/newsrevenuehub/rev-engine/pull/1128/commits/b2c0628a956ab10559a80370b3eec42b3787063b)

apps/emails/tasks.py
```diff
@@ -124,14 +133,77 @@ def make_send_thank_you_email_data(contribution) -> SendThankYouEmailData:
...
+    from apps.api.serializers import ContributorObtainTokenSerializer
+    from apps.api.views import construct_rp_domain
+    from apps.contributions.models import Contributor
+
+    # --- Generate magic link ---
+    serializer = ContributorObtainTokenSerializer(data={"email": user.email, "subdomain": revenue_program.slug})
+    serializer.is_valid(raise_exception=True)
+    contributor, created = Contributor.objects.get_or_create(email=user.email)
+    if created:
+        logger.info("[make_send_test_contribution_email_data] Created new contributor with email %s", user.email)
```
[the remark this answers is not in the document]
GuiMend: Done in [c92d8f5](https://github.com/newsrevenuehub/rev-engine/pull/1128/commits/c92d8f50741cc88902b9f170f035a4340748f1df)

benjaminEwhite (approved):

PR #1134, opened 2023-05-26
DEV-3477-animate-on-the-contributions-dashboard-upgrade-prompt

GuiMend:

#### What's this PR do?
Animate upgrade prompt show that it's more visible to the users.

We want to make the upgrade prompt more visible for users.

- Log in as a Free user, connect to stripe and publish a page
- Go in and out of the Contributions dashboard page
- You should see the prompt animating 

No

Yes

N/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-3477

no

no

### DIFF
--- a/spa/src/components/donations/DonationUpgradePrompts/DonationUpgradePrompts.test.tsx
+++ b/spa/src/components/donations/DonationUpgradePrompts/DonationUpgradePrompts.test.tsx
@@ -6,7 +6,7 @@ import {
 } from 'constants/authConstants';
 import { DONATIONS_CORE_UPGRADE_CLOSED } from 'hooks/useSessionState';
 import useUser from 'hooks/useUser';
-import { fireEvent, render, screen } from 'test-utils';
+import { act, fireEvent, render, screen, waitFor } from 'test-utils';
 import DonationUpgradePrompts from './DonationUpgradePrompts';
 import useContributionPageList from 'hooks/useContributionPageList';
 
@@ -53,6 +53,27 @@ describe('DonationUpgradePrompts', () => {
     it('shows a Core upgrade prompt', () => {
       tree();
       expect(screen.getByTestId('mock-donation-core-upgrade-prompt')).toBeInTheDocument();
+    });
+
+    describe('Prompt Highlight', () => {
+      it('shows a highlight', () => {
+        tree();
+        expect(screen.getByTestId('prompt-highlight-true')).toBeInTheDocument();
+      });
+
+      it('hides the highlight after 1 second', async () => {
+        jest.useFakeTimers();
+        tree();
+        expect(screen.getByTestId('prompt-highlight-true')).toBeInTheDocument();
+        act(() => {
+          jest.runAllTimers();
+        });
+        await waitFor(() => {
+          expect(screen.getByTestId('prompt-highlight-false')).toBeInTheDocument();
+        });
+
+        jest.useRealTimers();
+      });
     });
 
     it('hides the prompt when closed', () => {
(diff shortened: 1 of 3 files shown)


spa/src/components/donations/DonationUpgradePrompts/DonationUpgradePrompts.test.tsx
```diff
@@ -170,4 +170,70 @@ describe('DonationUpgradePrompts', () => {
...
+      expect(screen.getByTestId('prompt-highlight-true')).toBeInTheDocument();
+      act(() => {
+        jest.runAllTimers();
+      });
+      await waitFor(() => {
+        expect(screen.getByTestId('prompt-highlight-false')).toBeInTheDocument();
+      });
+    });
+
+    it.skip('animation flow is done correctly', async () => {
```
GuiMend, line 220: @nrh-cklimas I couldn't manage to make this tests pass. I'm having an issue making the timer of `setTimeout` run after `showAnimation` is set to `true`. Any ideas on how to fix this?
Btw, not entirely sure this is something we need to test (making sure the `showAnimation` gets set to `true` is important to show the popup, but hiding the highlight is more of a UI thing, so I could probably just remove these tests)

```js
useEffect(() => {
    if (showAnimation) {
      setTimeout(() => {
        setShowHighlight(false);
      }, 500);
    }
  }, [showAnimation]);
```

nrh-cklimas: I don't fully understand it, but I think using `waitFor` in conjunction with fake timers doesn't work well, probably because RTL is using `setTimeout` too. ([GH issue I found](https://github.com/testing-library/react-hooks-testing-library/issues/631)) It kind of makes sense that if we're faking time, we shouldn't need to use `waitFor` since we're manipulating time ourselves, so we shouldn't need to wait for anything to happen.

In general, I think it's nice to have unit tests around animations but it's a very close cousin to testing appearance, which we don't do. It's best left for Storybook imho.

Here are versions of two skipped tests that work locally. The two acts()s in the first test are to allow the effects to trigger correctly.

```tsx
    it('hides the highlight 1 second after animation', () => {
      tree();
      expect(screen.getByTestId('prompt-highlight-true')).toBeInTheDocument();
      act(() => jest.advanceTimersByTime(animationTimeout));
      act(() => jest.advanceTimersByTime(1000));
      expect(screen.getByTestId('prompt-highlight-false')).toBeInTheDocument();
    });

    it('triggers fades in the correct order', () => {
      tree();
      expect(screen.getByTestId('show-animation-false')).toBeInTheDocument();
      act(() => jest.advanceTimersByTime(animationTimeout));
      expect(screen.getByTestId('show-animation-true')).toBeInTheDocument();
      expect(screen.getByTestId('prompt-highlight-true')).toBeInTheDocument();
      act(() => jest.runAllTimers());
      expect(screen.getByTestId('prompt-highlight-false')).toBeInTheDocument();
    });

```

spa/src/components/donations/DonationUpgradePrompts/DonationUpgradePrompts.tsx
```diff
@@ -15,10 +17,25 @@ export function DonationUpgradePrompts() {
   const { isOrgAdmin } = getUserRole(user);
   const { pages } = useContributionPageList();
   const [coreUpgradePromptClosed, setCoreUpgradePromptClosed] = useSessionState(DONATIONS_CORE_UPGRADE_CLOSED, false);
+  const [showAnimation, setShowAnimation] = useState(false);
+  const [showHighlight, setShowHighlight] = useState(true);
+
+  useEffect(() => {
```
nrh-cklimas, line 23: This and the effect on line 31 should return cleanup functions that clear the timeout.

nrh-cklimas (approved):

PR #1141, opened 2023-06-04
DEV-3593 | Hotfix to suppress revenue_programs list when superuser from user serializer

benjaminEwhite:

As a short term fix to enable full e2e testing of Mailchimp integration (inclusive of switchboard), we can suppress revenue_programs from the user serializer for superusers.

Once DEV-3589 has shipped, the changes in this PR can be reverted (though I'm not sure there's a strong reason to)

#### What's this PR do?

When a superuser logs in via /api/v1/token/, the returned value for `revenue_programs` will be an empty list, to prevent an acute bug that causes timeouts. 

Scrappy way to unblock testing e2e

In staging or test should be able to authenticate with a superuser with no app error.  Can verify in this review app, but that's not a proof that it works since it's something about test / staging that  causes the timeout to be worse than review (perhaps stale access tokens or similar due to intra-environment sharing).  Logging in as superuser should also be noticeably faster than compared to a hubadmin.

No

Yes, updated existing

N/A

No

N/A

[DEV-3589](https://news-revenue-hub.atlassian.net/browse/DEV-3589)
[DEV-3593](https://news-revenue-hub.atlassian.net/browse/DEV-3593)

No

N/A

[DEV-3589]: https://news-revenue-hub.atlassian.net/browse/DEV-3589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


poetry.lock
benjaminEwhite, line 1: This diff was what resulted from running `poetry update django` in order to update that package to solve for safety warning. See broken pre-commit in actions for this PR.

x110dc (approved):

PR #1143, opened 2023-06-09
DEV-3394  Automate default donation page designation when plan is on core

benjaminEwhite:

#### What's this PR do?

Registers a post-save signal receiver function called `handle_set_default_donation_page_on_select_core_plan` with `Organization` as sender.

Under well-defined conditions, this signal handler will set a default donation page on one of the org's revenue programs. Specifically, if...

- the org is on core plan
- there is no default donation page for the relevant RP
- the org is being newly created with plan set as core OR org is being updated and plan name is changing to core

... then the handler attempts to set an existing page controlled by the RP as the default donation page.

Will provide default styles for orgs on core after they upgrade.

From the Django admin, find an org that is not on the Core plan.  Choose an org with one RP that has at least one page but no default donation page.

Update the org's plan name to Core and click save. Look at the detail view for the RP and you should see that it now has a value set for default donation page.

No

Yes

N/A

No

N/A

[DEV-3394](https://news-revenue-hub.atlassian.net/browse/DEV-3394)

No

No

[DEV-3394]: https://news-revenue-hub.atlassian.net/browse/DEV-3394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/organizations/signals.py
```diff
@@ -34,20 +37,91 @@ def revenue_program_post_save(sender, instance, created, **kwargs) -> None:
...
+        logger.debug("Org %s is not on CorePlan, skipping", instance.id)
+        return
+    if not (rp := instance.revenueprogram_set.first()):
+        logger.debug("No RP found for organization %s, skipping", instance.id)
+        return
+    is_update_to_core_plan = all(
+        [
+            not created,
+            (update_fields := kwargs.get("update_fields", None) or {}),
+            "plan_name" in update_fields,
```
njharman, line 104: Don't believe (this) update_fields used below. Simplier, less local name pollution as 

`"plan_name" in (kwargs.get("update_fields", None) or {})`

apps/organizations/signals.py
```diff
@@ -34,20 +37,91 @@ def revenue_program_post_save(sender, instance, created, **kwargs) -> None:
...
+    is_create_with_core_plan = all([created, instance.plan_name == CorePlan.name])
+    if not any([is_update_to_core_plan, is_create_with_core_plan]):
+        logger.debug(
+            "Skipping for org %s because is_update_to_core_plan is %s and is_create_with_core_plan is %s",
+            instance.id,
+            is_update_to_core_plan,
+            is_create_with_core_plan,
+        )
+        return
+    if rp.default_donation_page:
```
njharman, line 117: Minor/meh.  I'd put rp default page check above update/create after the related does rp exist check

njharman (approved):
Lgtm. Only a couple "opinions" ;) re structure.

+1 the handler function renames, info -> debug logging changes.

apps/organizations/signals.py
benjaminEwhite: a3bab546f7

apps/organizations/signals.py
```diff
@@ -34,20 +37,92 @@ def revenue_program_post_save(sender, instance, created, **kwargs) -> None:
...
         )
         del instance.mailchimp_access_token
     else:
-        logger.info("No mailchimp_access_token_secret to delete for rp %s", instance.id)
+        logger.debug("No mailchimp_access_token_secret to delete for rp %s", instance.id)
+
+
+def get_page_to_be_set_as_default(rp) -> DonationPage:
+    """Return the page to be set as default donation page for an RP"""
+    logger.debug("get_page_to_be_set_as_default called on rp %s", rp.id)
```
x110dc, line 61: May want to document a text description of how we pick the default page. And in the PR description under documentation we'll want to note this in case there are questions the support team gets.

x110dc (changes requested):
LGTM. I just had one request that we make sure we document how we select the default page/RP for the support team's use.

apps/organizations/signals.py
benjaminEwhite: Attaching a first pass documenting default donation pages, but in Jira.

x110dc (approved):

PR #1144, opened 2023-06-09
DEV-3616 make MailchimpRevenueProgramForSwitchboard.mailchimp_server_prefix and .stripe_account_id default to empty string

benjaminEwhite:

#### What's this PR do?

This updates the definition of `mailchimp_server_prefix` and `stripe_account_id` properties on `MailchimpRevenueProgramForSwitchboard` such that they return an empty string if their related db values are None. We're doing this so that Switchboard can type those fields in response as string and not have to worry about None/null (it was expecting only string, and broke when got null).

Solve a bug in switchboard and simplify the interface.

- In the Django admin, find an RP that has null value for payment provider (or with a payment provider that doesn't have stripe account id set) and a null value for mailchimp_server_prefix
- In the review app, log in as a superuser at https://dev-3616.revengine-review.org/api/v1/token/
- Now go to https://dev-3616.revengine-review.org/api/v1/revenue-programs/ID-OF-YOUR-RP/mailchimp/ . You should see empty string values for "stripe_account_id" and `"mailchimp_server_prefix" in the response data.

No

Yes

N/A

No

N/A

[DEV-3616](https://news-revenue-hub.atlassian.net/browse/DEV-3616)

No

No

[DEV-3616]: https://news-revenue-hub.atlassian.net/browse/DEV-3616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/organizations/serializers.py
+++ b/apps/organizations/serializers.py
@@ -168,11 +168,11 @@ class MailchimpRevenueProgramForSwitchboard(serializers.ModelSerializer):
     """Primary consumer is switchboard API. This is a read-only serializer."""
 
     mailchimp_integration_connected = serializers.ReadOnlyField()
-    mailchimp_server_prefix = serializers.ReadOnlyField(allow_null=True)
+    mailchimp_server_prefix = serializers.SerializerMethodField()
     mailchimp_store = serializers.SerializerMethodField()
     mailchimp_one_time_contribution_product = serializers.SerializerMethodField()
     mailchimp_recurring_contribution_product = serializers.SerializerMethodField()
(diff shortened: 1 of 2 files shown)


leo-dcfa (commented):
@benjaminEwhite I think this should be fixed in switchboard api instead, no? Make the `stripe_account_id` and `mailchimp_server_prefix` `str | None` instead, where the latter has a value if `mailchimp_integration_connected` is `True

PR #1177, opened 2023-07-18
DEV-3748 | Implement organization stripe webhook receiver endpoint

benjaminEwhite:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

This is PR 3 of 5 stacked PRs to fully implement server-side support for self-upgrade webhook events.

1. #1174
2. #1178
3. #1177 (this PR)
4. #1185 
5. #1176 

**NOTE** In short term, this PR  has `DEV-3746` as its base, but that branch is itself going to be merged into DEV-1750, which is the intended  base for this PR.  Once #1178  has been merged, DEV-1750 should be merged in here, and base should be DEV-1750.

#### What's this PR do?

This PR implements the business logic for self-upgrades in an API endpoint that is designed to receive POST requests made by Stripe webhook events.

The specific endpoint is `/api/v1/organizations/handle_stripe_webhook`. The happy path flow for this endpoint is:

1. Construct a Stripe event object from the request data (which by default includes header validation)
2. The event will include a "client_reference_id". We try to find an org whose `uuid` value (a net new field on the model) is the same as "client_reference_id".
3. We then set that org's stripe subscription ID (a net new field on the model) to the Stripe subscription ID in request data.
4. We set org's plan name to CorePlan.name.
6. We save the org creating a reversion revision and using update fields.

Some important implementation details:

- We rename some env vars:
  + STRIPE_WEBHOOK_SECRET_FOR_CONTRIBUTIONS => STRIPE_WEBHOOK_SECRET_CONTRIBUTIONS
  + STRIPE_WEBHOOK_SECRET_FOR_UPGRADES => STRIPE_WEBHOOK_SECRET_UPGRADES
  + HOOKDECK_WEBHOOK_SOURCE => HOOKDECK_STRIPE_WEBHOOK_SOURCE_CONTRIBUTIONS
- We add some env vars:
  + STRIPE_CORE_PRODUCT_ID
  + HOOKDECK_STRIPE_WEBHOOK_SOURCE_UPGRADES
- In an earlier ticket for Hookdeck refactoring (DEV-3741), we didn't know we needed a requirement around having distinct Hookdeck sources, but that need became apparent in trying to do e2e testing on full user flow. In this ticket, we refactor `bootstrap_endpoint` such that it gets passed a `source_id` parameter.
- The call signature for the `create_stripe_webhooks` command changes to use optional args instead of positional
[...]

PR #1174, opened 2023-07-14
Dev 3741 refactor hookdeck integration to accommodate creation of two separate webhook endpoints and accounts

benjaminEwhite:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

This is PR 1 of 5 stacked PRs to fully implement server-side support for self-upgrade webhook events.

1. #1174 (this PR)
2. #1178
3. #1177
4. #1185 
5. #1176 

#### What's this PR do?

Updates our Hookdeck integration to support creation of two distinct webhook receivers (managed by two distinct Stripe accounts).  Note that much of the diff in this PR (in terms of file count) stems from updating references to configuration variables who had their name change.

- Updates the signature for `apps.common.hookdeck.bootstrap`, retiring `destination_url` in favor of `webhooks_url_contributions` (for pre-existing contribution management-related webhooks) and `webhooks_url_upgrades` (for the new webhook receiver that will only listen for events tied to the upgrades Stripe-billing related account). 
- The signature change for  `apps.common.hookdeck.bootstrap` corresponds to a change in side effects -- this command now causes both webhook endpoints to be configured for Hookdeck.
- Adds a "dummy" view for the upgrade webhook endpoint (as detail view `OrganizationViewset.handle_stripe_webhook`) that returns a 200 without doing anything else.
- Churns some var names in settings.base..py
  + `STRIPE_WEBHOOK_SECRET` becomes `STRIPE_WEBHOOK_SECRET_FOR_CONTRIBUTIONS`
  + `STRIPE_LIVE_SECRET_KEY ` becomes `STRIPE_LIVE_SECRET_KEY_CONTRIBUTIONS `
  + `STRIPE_TEST_SECRET_KEY` becomes `STRIPE_TEST_SECRET_KEY_CONTRIBUTIONS`

#### Why are we doing this? How does it help us?

This is an incremental step that moves us toward self-upgrade.

#### How should this be manually tested? Please include detailed step-by-step instructions.

Confirm that when a review app with changes in this PR is first created on Heroku, the post deploy script calls `bootstrap-review-app.py` which causes two destinations and two connections to be created. 

[...]

PR #1185, opened 2023-07-20
DEV-3760 | add uuid to inline org serializer

benjaminEwhite:

This is PR 4 of 5 stacked PRs to fully implement server-side support for self-upgrade webhook events.

1. #1174 
2. #1178
3. #1177
4. #1185 (this PR) 
5. #1176  (this goes into main/prod)

Note that the base for this PR is `DEV-1750-add-stripe-webhook-event-listeners-for-self-upgrade`. That branch gets merged to main in #1176 

#### What's this PR do?

- Reconfigures pricing table to use org.uuid instead of org.id as value for `client_reference_id`
- Exposes `uuid` on `OrganizationInlineSerializer`, which is the serializer used in the user object as returned by authentication endpoint, and in turn is the data source for the SPA side functionality around self upgrade flow.
- Adds a new serializer test class for `OrganizationInlineSerializer` and minimally tests that serializer has expected fields.
- Makes hookdeck module handle empty values more gracefully (this came into scope in debugging some issues).
- Renames and refactors the migration that adds UUID to organization. It turned out that needed a `RunPython` function to handle generating the UUIDs (the original migration was causing build failures).  So,  `0021_3748_add_uuid_and_stripe_subscription_id_to_org.py` becomes `0021_DEV-1750_add_uuid_and_stripe_subscription_id_to_org`. Note that I've named the migration DEV-1750... because it will be when the parent ticket/branch DEV-1750 is merged that migrations will run in main.

So we don't use predictably incrementing integer values for matching webhooks with organizations.

The backend has been developed with uuid in mind so will also enable full e2e testing on self-upgrade + webhook flow.

1. In review app, get an org user in state where they're on free plan, and have the `spa-self-upgrade-access` feature flag turned on.
2. Go through upgrade flow.
3. Find/refresh the organization in the Django admin. It should now have core plan, and it should have a value for `stripe_subscription_id`.

No.

Yes.

N/A. We'll QA the entire self-upgrade flow via DEV-1750 parent ticket / #1176 

No

No

- [DEV-3760](https://news-revenue-hub.atlassian.net/browse/DEV-3760)
- [DEV-1750](https://news-revenue-hub.atlassian.net/browse/DEV-1750)

No.

[DEV-3760]: https://news-revenue-hub.atlassian.net/browse/DEV-3760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nrh-cklimas (approved):
TypeScript changes look good to me.

njharman (approved):

x110dc (approved):

apps/organizations/migrations/0021_DEV-1750_add_uuid_and_stripe_subscription_id_to_org.py
```diff
@@ -1,9 +1,15 @@
...
 from django.db import migrations, models
 import uuid
 
 
+def create_uuid(apps, schema_editor):
+    Organization = apps.get_model("organizations", "Organization")
+    for organization in Organization.objects.all():
+        organization.uuid = uuid.uuid4()
+        organization.save(update_fields={"uuid", "modified"})
+
```
x110dc, line 11: I haven't thought about migrations in a while so this may be a non-issue, but could this be potentially destructive if this migration is run more than once? Under what circumstances could that happen? If that is unlikely to ever happen in production that seems fine.

x110dc (changes requested):
I just had one question about the migration. Probably nothing.

apps/organizations/migrations/0021_DEV-1750_add_uuid_and_stripe_subscription_id_to_org.py
benjaminEwhite: That's a good point. I pushed up this commit that bf6e2dc7f4d8e filters down to only org objects whose uuid is null. Does that address your concern?

bf6e2dc7f4d8e

PR #1200, opened 2023-08-09
DEV-3779: Style this email in accordance with design spec

GuiMend:

n/a

#### What's this PR do?
Update style for email template: Upgrade to core confirmation

Make email template styling consistent

yes

yes

n/a

n/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-3779

no

Parent ticket: https://news-revenue-hub.atlassian.net/browse/DEV-3318


apps/organizations/views.py
```diff
@@ -142,6 +143,11 @@ def send_upgrade_success_confirmation_email(cls, org: Organization):
             .distinct("user__email")
         ):
             context = {
+                "logo_url": os.path.join(settings.SITE_URL, "static", "nre_logo_black_yellow.png"),
+                "plus_icon": os.path.join(settings.SITE_URL, "static", "plus-icon.png"),
+                "mail_icon": os.path.join(settings.SITE_URL, "static", "mail-icon.png"),
+                "paint_icon": os.path.join(settings.SITE_URL, "static", "paint-icon.png"),
+                "check_icon": os.path.join(settings.SITE_URL, "static", "check-icon.png"),
```
benjaminEwhite, line 150: I think you should be able to use the `static` tag in the mail template in order to avoid sending additional context.  See Django docs here: https://docs.djangoproject.com/en/3.2/howto/static-files/

benjaminEwhite: In which case, not something we need to test around, since you'd be testing Django implementation.

apps/emails/tests/test_templates.py
```diff
@@ -10,6 +12,11 @@
 def upgrade_confirmation_email_context():
     return {
         "mailchimp_integration_url": "some-url",
+        "logo_url": os.path.join(settings.SITE_URL, "static", "nre_logo_black_yellow.png"),
+        "plus_icon": os.path.join(settings.SITE_URL, "static", "plus-icon.png"),
+        "mail_icon": os.path.join(settings.SITE_URL, "static", "mail-icon.png"),
+        "paint_icon": os.path.join(settings.SITE_URL, "static", "paint-icon.png"),
+        "check_icon": os.path.join(settings.SITE_URL, "static", "check-icon.png"),
```
benjaminEwhite, line 19: See comment below. Think you probably don't need this if use the `static` template tag.

benjaminEwhite (changes requested):
From AC perspective, looks all good.

One small request to use the `static` template tag instead of sending in context for the image assets.

apps/organizations/views.py
GuiMend: @benjaminEwhite , I tried doing it like you suggested but I'm not able to make it work yet.
Could you review my code to see if you spot what I'm doing wrong?

image

Btw, I tried adding the icons into `revengine/static/...` and I also tried creating a static folder into the parent folder for `templates` (in the same format as `admin_limited_select.html` file)

benjaminEwhite: @GuiMend I looked into this and I believe it would require some configuration of whitenoise that we don't already do. On other hand, I'm finding pre-existing examples of the approach you originally took. I thought this would be a quick fix, but don't think it's worth the effort just to avoid sending context. No need to change.

benjaminEwhite (approved):

PR #1215, opened 2023-08-23
DEV-3865 | Update caches on subscription update and cancelation

benjaminEwhite:

#### What's this PR do?

This PR updates the `SubscriptionViewSet.partial_update` and  `.destroy` methods to update subscriptions and PIs in cache after update/cancelation.

I also did a bit of cleanup of tests of this viewset, consolidating all tests for it to pytest and using a single test class for all tests related to this view (they were formerly broken up across > 1).

Provide contributors with feedback so they know subscription cancellations have had intended effect.

Yes.

No

Yes

N/A

No

N/A

[DEV-3865](https://news-revenue-hub.atlassian.net/browse/DEV-3865)

No

N/A

[DEV-3865]: https://news-revenue-hub.atlassian.net/browse/DEV-3865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/views.py
```diff
@@ -356,18 +359,18 @@ class SubscriptionsViewSet(viewsets.ViewSet):
...
     serializer_class = serializers.SubscriptionsSerializer
 
-    def _fetch_subscriptions(self, request):
+    @staticmethod
+    def _fetch_subscriptions(request: Request) -> List[StripePiAsPortalContribution]:
         revenue_program_slug = request.query_params.get("revenue_program_slug")
         try:
             revenue_program = RevenueProgram.objects.get(slug=revenue_program_slug)
         except RevenueProgram.DoesNotExist:
-            return Response({"detail": "Revenue Program not found"}, status=status.HTTP_404_NOT_FOUND)
```
benjaminEwhite, line 364: I came across this and realized that it will not lead to a response. Instead it will lead to an unexpected error in calling context, because caller expects this to return a list of `StripePiAsPortalContribution`s, not a Response.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -176,7 +187,7 @@ class StripeContributionsProvider:
     FETCH_SUB_EXPAND_FIELDS = ["data.default_payment_method"]
 
     def __init__(self, email_id, stripe_account_id) -> None:
-        self.email_id = email_id
+        self.email_id = email_id.lower()
```
njharman, line 190: I don't see a test that verifies email_id will be lowercased. I could have missed.  Probably more important for the cache key creation in `__init__`s below.

Reasoning being, if it mattered to add to code (supposedly to catch some issue), then it matters enough to protect against regression with test case.

apps/contributions/views.py
njharman: +1 (for adding test :) oh, and for making method that does not use self, @static.

apps/contributions/views.py
```diff
@@ -356,18 +359,18 @@ class SubscriptionsViewSet(viewsets.ViewSet):
...
+            logger.warning("Revenue program not found for slug %s", revenue_program_slug)
+            raise Http404("Revenue program not found")
+        cache_provider = SubscriptionsCacheProvider(request.user.email, revenue_program.stripe_account_id)
         subscriptions = cache_provider.load()
         if not subscriptions:
-            task_pull_serialized_stripe_contributions_to_cache(
-                self.request.user.email, revenue_program.stripe_account_id
-            )
+            task_pull_serialized_stripe_contributions_to_cache(request.user.email, revenue_program.stripe_account_id)
         subscriptions = cache_provider.load()
```
njharman, line 374: If I understand logic; should this (second call to cache_provider.load()) not be indented into the `if not subscriptions`. As it is now subscriptions will be pulled from cache twice (when it was there first time).

apps/contributions/stripe_contributions_provider.py
```diff
@@ -318,7 +329,7 @@ class ContributionsCacheProvider:
 
     def __init__(self, email_id, stripe_account_id) -> None:
         self.stripe_account_id = stripe_account_id
-        self.key = f"{email_id}-payment-intents-{self.stripe_account_id}"
+        self.key = f"{email_id.lower()}-payment-intents-{self.stripe_account_id}"
```
njharman, line 332: Are stripe_account_id strictly numeric (not uuid)?  I'd just lower() the entire key to not have to ever worry about it.

ICYDK, literal strings have methods like lower `f"{email...}-payment...".lower()`

apps/contributions/views.py
```diff
@@ -417,39 +424,118 @@ def partial_update(self, request, pk):
...
+                expand=["payment_method", "invoice.subscription.default_payment_method"],
+            )
+            # we only update cache if able to successfully retrieve PI
+            self.update_subscription_in_cache(
+                email,
+                revenue_program.payment_provider.stripe_account_id,
+                subscription,
+                pi,
+            )
+        except stripe.error.StripeError:
```
njharman, line 459: [Not actionable commentary]  I failed to find a definitive answer if Django cache system will not raise Exception. So, I can't really recommend also catching those (Although I'd really recommend that our CacheProviders catch, log, and suppress Exceptions (which is probably/hopefully what Django is doing)).

Caching being an "add on" should fail gracefully, but loudly (logigng).

apps/contributions/views.py
```diff
@@ -417,39 +424,118 @@ def partial_update(self, request, pk):
...
+            # only need to update cache if we successfully retrieve sub and PI
+            self.update_subscription_in_cache(email, revenue_program.payment_provider.stripe_account_id, sub, pi)
+        except stripe.error.StripeError:
+            # in this case, we only want to log because the subscription has already been canceled and user shouldn't retry
+            logger.exception(
+                "stripe.PaymentIntent.retrieve returned a StripeError after canceling subscription when re-retrieving PI %s",
+                sub.latest_invoice.payment_intent.id,
+            )
+        # TODO: [DEV-2438] return the canceled sub
+        return response
```
njharman, line 527: Isn't this largely the same repeated code from partial_update? It should be DRY.

@x110dc Out of scope but I'm still gonna complain; 
I don't think I like the way we've implemented caching [I might not know enough of our business/use cases and be wrong]. It should be much lower down and more invisible to "users" of storage system (such as this view). I'd probably want to wrap stripe.retrieve/modify/etc with cache (I think those are the "store").  Which would mean the cache always matches "store" and would probably have prevented this tickets issue.

I just see more errors as "users" forget to also update cache and even if they do, requiring too much repeated boiler plate code.

njharman (changes requested):
Great code coverage, good changes, +1 converting 403 to 404 for security greatness!

Some small requests/clarifications and larger DRY one.

apps/contributions/stripe_contributions_provider.py
```diff
@@ -105,9 +105,12 @@ def subscription_id(self):
...
     def card(self):
-        return getattr(self.payment_intent.payment_method, "card", None) or AttrDict(
-            **{"brand": None, "last4": None, "exp_month": None}
-        )
+        default = AttrDict(**{"brand": None, "last4": None, "exp_month": None})
+        if invoice := self.payment_intent.invoice:  # it's a subscription
+            return ((invoice.get("subscription", {}) or {}).get("default_payment_method", {}) or {}).get(
+                "card", {}
+            ) or default
+        return (self.payment_intent.get("payment_method", {}) or {}).get("card", {}) or default
```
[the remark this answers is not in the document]
benjaminEwhite: They're `AttrDict`-like in that Stripe objects behave both like dicts and objects.

The change here is narrowly to make this property after changing things so we update cache. I'm reluctant to rethink the overall existing implementation.

apps/contributions/views.py
benjaminEwhite: Don't disagree, but out of scope for this PR.

apps/contributions/views.py
benjaminEwhite: FWIW, @njharman part of the reason we're not doing a deeper rethink of the caching layer here is there's a good chance it will go away altogether in some currently planned work. I believe we'll be relying more on storing relevant stripe data in Revengine and ingesting it via webhooks.

apps/contributions/views.py
benjaminEwhite: Great point! 6cc54832a0f

apps/contributions/views.py
benjaminEwhite: I'm generally all for DRYing up, but going to push back on that here. 

The first try/except block in the code you cited is not comparable to partial update. In the case of partial_update, we don't need to re-retrieve the subscription because `stripe.Subscription.modify` returns it (and can expand requried fields). 

In destroy, we need to re-retrieve the subscription to get fresh state, as `stripe.Subscription.delete`  does not  allow for expanding fields, and we need those for things to work out with serialization/caching.

That said, the second try/except can be dried up. Did that in 00db11f4522

apps/contributions/stripe_contributions_provider.py
benjaminEwhite: Ended up just lowering as you suggested below when creating `self.key` in respective caches.

9c83555ad0f

apps/contributions/stripe_contributions_provider.py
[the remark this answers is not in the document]
benjaminEwhite: AttrDict and Stripe objects are similar though not identical in behaviors. Shared traits: act like dict, act like object.

Difference is that if you look up a nonexistent property/key in an attrdict, you get back an empty dict. 

If you do that on a Stripe object, you get an AttributeError.

apps/contributions/stripe_contributions_provider.py
benjaminEwhite: agree. this is a better approach. 

9c83555ad0f

apps/contributions/stripe_contributions_provider.py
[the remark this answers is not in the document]
njharman: Yep, ok

apps/contributions/views.py
njharman: fair enough

njharman (approved):
Ship it!

x110dc (approved):

PR #1219, opened 2023-08-25
DEV-3907 | enforce serializer level validation for max length and reduce sentry noise

benjaminEwhite:

#### What's this PR do?

This PR adds max length validation for several fields on the CustomizeAccountSerializer. Until now, these max lengths were only enforce at the db-level, causing uncaught 500 errors being returned to the SPA.

We now validate max length on the following fields at the serializer level

- first_name
- last_name
- organization_name
- fiscal_sponsor_name
- job_title

Also, validation errors are expected, and therefore we log at info level so they don't cause noise in Sentry.

- Part of solving for a critical bug in org dashboard that stops new users from completing full sign up workflow
- Reducing noise in Sentry makes it more valuable.

Yes

Yes

N/A

No

No

- [DEV-3907](https://news-revenue-hub.atlassian.net/browse/DEV-3907)
- [DEV-3911](https://news-revenue-hub.atlassian.net/browse/DEV-3911)

No

Yes. [DEV-3911](https://news-revenue-hub.atlassian.net/browse/DEV-3911)

[DEV-3907]: https://news-revenue-hub.atlassian.net/browse/DEV-3907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


njharman (approved):
approved

apps/organizations/models.py
```diff
@@ -196,6 +204,32 @@ def user_is_member(self, user):
     def user_is_owner(self, user):
         return user in [through.user for through in self.user_set.through.objects.filter(is_owner=True)]
 
+    @classmethod
+    def generate_unique_name(cls, name: str) -> str:
```
benjaminEwhite, line 208: This logic used to live  inline in the `customize_account` view function. Isolated and moved it to model so potentially can be used in other context and to make it more easily testable.

apps/organizations/models.py
```diff
@@ -463,7 +497,7 @@ class RevenueProgramManager(models.Manager):
 
 
 class RevenueProgram(IndexedTimeStampedModel):
-    name = models.CharField(max_length=255)
+    name = models.CharField(max_length=RP_NAME_MAX_LENGTH)
```
benjaminEwhite, line 500: Moved to a constant so can reference in other files (especially tests).

apps/users/permissions.py
```diff
@@ -6,30 +6,28 @@ class UserOwnsUser(permissions.BasePermission):
...
-def update_keys_dont_require_email_verification(request):
+def update_keys_dont_require_email_verification(request) -> bool:
     allowed_update_keys = ["password"]
     return request.method == "PATCH" and set(request.data.keys()).issubset(set(allowed_update_keys))
 
 
-def user_email_is_verified(request):
-    return request.user and request.user.email_verified
+def user_email_is_verified(request) -> bool:
+    return bool(request.user) and request.user.email_verified
```
benjaminEwhite, line 22: Added unit test for this, and in so doing discovered that original implementation returns None instead of boolean when request.user is None, which is not what we want (even if both are falsy).

apps/users/permissions.py
```diff
@@ -43,5 +41,5 @@ class UserHasAcceptedTermsOfService(permissions.BasePermission):
 
     message = "Please accept terms of service."
 
-    def has_permission(self, request, view):
-        return request.user and request.user.accepted_terms_of_service is not None
+    def has_permission(self, request, view) -> bool:
+        return request.user is not None and request.user.accepted_terms_of_service is not None
```
benjaminEwhite, line 45: Also here, wanted to ensure we will get a boolean, not None if no user.

apps/users/serializers.py
```diff
@@ -149,41 +171,104 @@ class Meta:
...
     def validate_fiscal_status(self, value):
         fiscal_sponsor_name = self.initial_data.get("fiscal_sponsor_name")
         if value == FiscalStatusChoices.FISCALLY_SPONSORED and not fiscal_sponsor_name:
-            raise serializers.ValidationError("Please enter the fiscal sponsor name.")
+            raise serializers.ValidationError({"fiscal_sponsor_name": [FISCAL_SPONSOR_NAME_REQUIRED_ERROR_MESSAGE]})
         elif fiscal_sponsor_name and value != FiscalStatusChoices.FISCALLY_SPONSORED:
             raise serializers.ValidationError(
-                "Only fiscally sponsored Revenue Programs can have a fiscal sponsor name."
+                {"fiscal_sponsor_name": [FISCAL_SPONSOR_NAME_NOT_PERMITTED_ERROR_MESSAGE]}
             )
```
benjaminEwhite, line 214: These are field-level errors, and so should return dict with key that's field name to follow pattern elsewhere.

apps/users/serializers.py
```diff
@@ -149,41 +171,108 @@ class Meta:
...
+        # which is 3 less than the max length of the Organization name field. The reason for the 3 less is that we allow for handle_organization_name
+        # to take that already validated value and append up to 3 characters to make it unique. That method is unit-tested elsewhere
+        # to prove that it raises OrgNameNonUniqueError if it gets past `-99` in its attempts to make the name unique.
+        self.validated_data["organization_name"] = name
+        self.validated_data["organization_slug"] = Organization.generate_slug_from_name(name)
+        return super().save(**kwargs)
+
+    @transaction.atomic
+    def create(self, validated_data: Dict[str, Any]) -> CustomizeAccountSerializerReturnValue:
+        """Create an organization, revenue program and role assignment. Also update user with new data
```
benjaminEwhite, line 243: This logic used to live in the view layer, but this is more idiomatic to have here, and also more easily testable.

apps/users/serializers.py
```diff
@@ -149,41 +171,108 @@ class Meta:
...
+        )
+        user = self.context.get("user")
+        user.first_name = validated_data["first_name"]
+        user.last_name = validated_data["last_name"]
+        user.job_title = validated_data["job_title"]
+
+        with reversion.create_revision():
+            # do this with revision history
+            user.save(update_fields={"first_name", "last_name", "job_title", "modified"})
+            reversion.set_comment("CustomizeAccountSerializer.create updated user")
```
benjaminEwhite, line 269: We were't doing this before, but in general, good idea to add revision any time we're modifying a model so there's a "paper" trail.

apps/users/tests/test_permissions.py
benjaminEwhite, line 1: We weren't unit testing these permissions in isolation before, which seemed non-ideal. By unit testing here, it allows us to just test that the customize_account endpoint has the right permissions set, and allows us to avoid having slow tests via view layer that check that permissions (i.e., log in as user who shouldn't have permissions, try to do the thing, assert about response).

apps/users/tests/test_serializers.py
```diff
@@ -4,23 +4,162 @@
...
+            ("truthy", FiscalStatusChoices.FISCALLY_SPONSORED, None),
+            ("", FiscalStatusChoices.FOR_PROFIT, None),
+            (None, FiscalStatusChoices.FOR_PROFIT, None),
+            ("truthy", FiscalStatusChoices.FOR_PROFIT, serializers.FISCAL_SPONSOR_NAME_NOT_PERMITTED_ERROR_MESSAGE),
+            ("", FiscalStatusChoices.NONPROFIT, None),
+            (None, FiscalStatusChoices.NONPROFIT, None),
+            ("truthy", FiscalStatusChoices.NONPROFIT, serializers.FISCAL_SPONSOR_NAME_NOT_PERMITTED_ERROR_MESSAGE),
+        ),
+    )
+    def test_validate_fiscal_status(
```
benjaminEwhite, line 147: Opportunistically added unit test for this method to get coverage on this file to 100%

apps/organizations/migrations/0022_DEV-3907_decrease_org_name_max_length.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("organizations", "0021_DEV-1750_add_uuid_and_stripe_subscription_id_to_org"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="organization",
+            name="name",
+            field=models.CharField(max_length=63, unique=True),
```
njharman, line 16: Just commenting/approving, good choice to use literal and not CONSTANT var here for max_length in DB migration file.

Not doing so has caused me pain.  Migrations are a state change at specific point in time.  A later modification of the  CONSTANT should not alter what this (or any previous) migration did. Instead, a new migration should be added that reflects the modification of constant's value.

apps/users/permissions.py
njharman: fair, I dislike that it's been implemented two different ways, casting here and ` if not None` below. I prefer if not None

apps/organizations/models.py
```diff
@@ -196,6 +204,32 @@ def user_is_member(self, user):
...
+        counter = 1
+        # we limit to 99 because we don't want to have to deal with 3-digit numbers.
+        # also, note that we would never expect to reach this limit and if we do, there's probably something
+        # untoward going on.
+        while counter <= MAX_APPEND_ORG_NAME_ATTEMPTS:
+            appended = f"{name}-{counter}"
+            if not cls.objects.filter(name=appended).exists():
+                return appended
+            else:
+                counter += 1
```
njharman, line 225: Counters are so "C". I'd use the power of Python!

```
for counter in range(1, MAX_APPEND_ORG_NAME_ATTEMPTS):
      appended = f"{name}-{counter}"
            if not cls.objects.filter(name=appended).exists():
                return appended
# after 99 attempts, naturally falls out of loop to log & raise.
```

njharman (changes requested):
Looks good, +1 coverage

Small issue and +1 comment, no need for re-rereview

apps/organizations/migrations/0022_DEV-3907_decrease_org_name_max_length.py
benjaminEwhite: Just the behavior I got by default by setting it to a literal in the model file, but I'll take the complement :) 

That said, I've encountered migration files in past that had hard coded constants and it caused headaches to be sure.

apps/users/permissions.py
benjaminEwhite: 71a9bb9d51b3

apps/organizations/models.py
benjaminEwhite: 29d6ed8b4f2

apps/organizations/models.py
```diff
@@ -196,6 +204,30 @@ def user_is_member(self, user):
...
+    @classmethod
+    def generate_unique_name(cls, name: str) -> str:
+        """Generate a unique organization name based on input name
+
+        Note that this does not guarantee that the name will be otherwise valid in terms of max length.
+        """
+        logger.info("Called with name %s", name)
+        if not cls.objects.filter(name=name).exists():
+            return name
+        counter = 1
```
x110dc, line 216: Does this need to be set when the loop will already set it to 1?

x110dc (approved):

x110dc (changes requested):
One question.

apps/organizations/models.py
benjaminEwhite: No, bad copy pasta after Norm's suggestion.

benjaminEwhite: 98ad4c26aef283071631d895e5d45f59cb3d9ec6

x110dc (approved):

PR #1227, opened 2023-09-05
DEV-3320 |  adding page 1 to slugified rp name pushes it over character limit

benjaminEwhite:

Note: This PR contains both server-side and SPA changes. 

At time of writing this description, I'm still wrapping up unit tests for the SPA-side changes, but didn't want that to block review of the backend changes.

Also, note that this PR solves for a critical bug that prevents org's with long RP names from publishing their pages. In status quo, when a user creates a page, an initial slug value is generated by the SPA behind the scenes, and when they go to publish, a new suggested value appears.

We now initially create the page without a slug value (and there is a change to the page model to allow for this). We also no longer provide a suggested slug to the user in the publish modal. We may decide to circle back on this and devote design resources to this detail of the UX, but we didn't want that to block solving this critical bug.  This is tracked in [DEV-3949](https://news-revenue-hub.atlassian.net/browse/DEV-3949).

#### What's this PR do?

This PR solves for a bug that allows org admins to be able to create new pages that have a slug that exceeds the maximum slug length. When this would happen, there was no indication of what caused the specific error or if there was anything they could do about it. In fact, there wasn't something they could do because the bug was caused by automatic generation of page slugs based on the revenue program's name (which can exceed the length allowed for page slug, which is only 50 chars).

This PR changes how pages are initially created, allowing for a null value for page slug. When a user first publishes a page, they must provide a page slug in the publish modal. 

Additionally, we now display a field-level validation error for the page slug if the server complains about the value they have attempted to set. 

Also of note, in this PR I converted the `PublishModal` component and associated files to typescript from .js and added a storybook story for this component.

Removes a bug that prevents org users from creating and publishing pages

Yes

No

Yes

N/A

No

N/A

- [DEV-3320](https://news-revenue-hub.atlassian.net/browse/DEV-3320)
- [DEV-3949](https://news-revenue-hub.atlassian.net/browse/DEV-3949)

- [DEV-3949](https://news-revenue-hub.atlassian.net/browse/DEV-3949) — note that this is non-blocking.

[DEV-3949]: https://news-revenue-hub.atlassian.net/browse/DEV-3949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/pages/serializers.py
```diff
@@ -305,10 +308,78 @@ def validate_sidebar_element_permissions(self, data):
                 {"sidebar_elements": f"You're not allowed to use the following elements: {', '.join(prohibited)}"}
             )
 
+    def is_valid(self, raise_exception: bool = False):
```
benjaminEwhite, line 311: In terms of UX, if the user provides a non-unique slug for the RP, the way they can fix this is by changing the value for the page slug.

That said, DRF raises this error as a "non-field" error. We cast the non-field error to a field-level validation error to make life easier for the SPA.

apps/pages/serializers.py
```diff
@@ -305,10 +308,78 @@ def validate_sidebar_element_permissions(self, data):
...
+        validation_error = serializers.ValidationError({"slug": "This field is required."})
+        # if it's new and no slug is provided or the sent slug is Falsy
+        if is_new and not sent_slug:
+            raise validation_error
+        # if it's not new and no slug was sent in data and the instance has no slug
+        if not is_new and not in_data and not self.instance.slug:
+            raise validation_error
+        # if it's not new and slug value was sent in data but the sent value is Falsy
+        if not is_new and in_data and not sent_slug:
+            raise validation_error
```
benjaminEwhite, line 342: FWIW, I usually prefer `if any([x, y, z])` but tried that in this block and for some reason I could not run down, the expected short-circuit behavior – of no longer evaluating conditions once a truthy one has been met – was not happening here.

apps/pages/tests/test_serializers.py
```diff
@@ -430,6 +454,115 @@ def test_validate_publish_limit_when_patching(self, plan, hub_admin_user):
             == f"Your organization has reached its limit of {org.plan.publish_limit} published page{'' if org.plan.publish_limit == 1 else 's'}"
         )
 
+    @pytest.mark.parametrize("include_published_date", (True, False))
+    def test_validate(self, include_published_date, hub_admin_user, mocker):
+        """The purpose of this test is just to prove that the validate method calls expected methods,
+        which are tested elsewhere.
+        """
```
njharman, line 461: Hope this isn't too ranty. I do think it's important code structure/design methodology to consider.

I'm -1 on tests like this. Which test the implementation details (what's called, how many functions are split). Rather than testing the "outputs/side effects"

They are fragile, require refactoring when implementation changes. [One major benefit of unit tests is the can be run before/after implementation change to confirm it didn't break existing behavior. Which is lost if you have to refactor tests]

They don't test what we care about. That bad data results in validation error.

Ideally tests should treat the unit under test as a black box. This is not strictly possible due to high side-effect nature of Python. Gotta mock out db calls. Really shouldn't be mocking out calls to methods of the class under test, or mocking out calls that don't do external I/O or have nasty side effects.   The closer you can write code that is testable as a black box, the less spagetti that code will be (depending only on inputs and producing only outputs).

apps/pages/tests/test_serializers.py
```diff
@@ -430,6 +454,115 @@ def test_validate_publish_limit_when_patching(self, plan, hub_admin_user):
...
+        else:
+            mock_validate_publish_limit.assert_not_called()
+        mock_validate_page_element_permissions.assert_called_once_with(data)
+        mock_validate_sidebar_element_permissions.assert_called_once_with(data)
+
+    @pytest_cases.parametrize(
+        "instance,expect",
+        (
+            (pytest_cases.fixture_ref("live_donation_page"), False),
+            (None, True),
```
njharman, line 493: There seems to be three "conditions" in code but only two tests.

1. not self.instance
2. yes self.instance, not self.instance.id
3. yes to both

apps/pages/serializers.py
```diff
@@ -305,10 +308,78 @@ def validate_sidebar_element_permissions(self, data):
...
+        """Ensure that a slug is unique for a given revenue program
+
+        NB: We are not using a UniqueTogetherValidator for this because we that validator
+        requires that the fields are truthy, and we allow nulls for slug.  A given RP can
+        have > 1 page with a null slug.
+
+        NB: we assume rp is already validated at this point, so we can safely access it in case of
+        new page creation.
+
+        and assume if key is there, it's non empty if string
```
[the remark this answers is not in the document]
njharman: def validate_revenue_program is not fully covered.  Don't know if/how that impacts assumption

njharman (changes requested):
Mostly commentary, couple minor fixes. Good coverage.

Maybe add note on slug that the max length 25? is set/validated in SPA.

I think the code allows a slug consisting of only white space (Actually SlugField/slugify probably strips). If whitespace only slugs are disallowed needs a test. I didn't see/look hard so it might exist.

I started to verify all combinations were covered, but found it difficult and got increasingly frustrated I couldn't refactor the code to make it easier.

apps/pages/tests/test_serializers.py
benjaminEwhite: I had a similar feeling about this test (that it's nothing but a test of implementation). On other hand, it offers a guarantee that we're doing some validation that is handled elsewhere. 

We already had tests for some of this,  and I've added some cases to round things out.

- `TestPageViewSet.test_create_when_already_at_page_limit`  proves that we don't allow creation if already at page limit
- `TestPageViewSet.test_create_when_invalid_data` now proves we can't create if unpermitted elements, sidebar elements, or being published but no slug.
-  `TestPageViewSet.test_patch_when_expected_user_with_invalid_data` proves that we can't create a page if prohibited elements or sidebar elements, and now also includes cases for slug + published interaction.


6b4fcff84293

333b8f6280

apps/pages/serializers.py
```diff
@@ -220,6 +224,17 @@ def get_benefit_levels(self, obj):
             serializer = BenefitLevelDetailSerializer(benefit_levels, many=True)
             return serializer.data
 
+    def validate_slug(self, value):
```
benjaminEwhite, line 227: @njharman after a bunch of digging, I determined that the easiest way to handle validation of slug min length is by having it on serializer in an explicit field-level validation. 

Since we're doing explicit validation here, it seems that we don't get the model-level validation that we would otherwise get (specifically, we don't get the `validate_slug_against_denylist` validation which is applied to slug field on page model).

FWIW, we formerly were validating uniqueness of RP + slug with a `rest_framework.validators import UniqueTogetherValidator`, but that validator would require that slug be non-null, and we've enabled null values for that field on the model (so that pages can be initially created with out a slug — the user is prompted to create a slug ahead of publishing a page).

apps/pages/tests/test_models.py
benjaminEwhite, line 1: In trying to fix the slug validation issue, I went down the path of applying a model-field level validator that ensured if value is string, it had min length of 1. That was before I discovered that those validators were not being called as expected via default field level validation in the serializer. I had added a unit test for the page model around this and took the opportunity to make this test suite entirely Pytest based.

I ended up backing out of validating slug min length on model for reasons explained in comment above, but as long as i put in effort to convert these tests to pytest, I kept these changes in.

apps/pages/tests/test_serializers.py
benjaminEwhite: 0ca83384fead16068ec57cb7a43aebfae65074c4

benjaminEwhite: addressed in 0ca83384fead16068

apps/pages/serializers.py
```diff
@@ -305,10 +308,78 @@ def validate_sidebar_element_permissions(self, data):
...
+            logger.debug("Slug is empty, so skipping uniqueness check")
+            return
+        if slug == "":
+            raise serializers.ValidationError({"slug": "This field must be at least 1 character long."})
+        if self.instance and self.instance.id:
+            already_exists = (
+                DonationPage.objects.filter(revenue_program=rp, slug=slug).exclude(id=self.instance.id).exists()
+            )
+        else:
+            already_exists = DonationPage.objects.filter(revenue_program=rp, slug=slug).exists()
```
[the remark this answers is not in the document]
benjaminEwhite: Did know, but didn't think of it here. Incorporated into this commit: cdb85081dba2e6

apps/pages/serializers.py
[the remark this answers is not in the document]
benjaminEwhite: This comment ended up requiring a fair amount of investigation and refactoring. Some of my assumptions about how model-field-level validation happens turned out (in this specific case) not to be true. Not going to point to specific commit because changes spanned several, but there is now extensive doc string explaining oddities.

apps/pages/tests/test_models.py
benjaminEwhite, line 1: In trying to fix the slug validation issue, I went down the path of applying a model-field level validator that ensured if value is string, it had min length of 1. That was before I discovered that those validators were not being called as expected via default field level validation in the serializer. I had added a unit test for the page model around this and took the opportunity to make this test suite entirely Pytest based.

I ended up backing out of validating slug min length on model for reasons explained in comment above, but as long as i put in effort to convert these tests to pytest, I kept these changes in.

apps/pages/tests/test_serializers.py
```diff
@@ -466,3 +594,9 @@ def test_get_used_live(self, published):
         page.styles = StyleFactory(revenue_program=page.revenue_program)
         page.save()
         assert StyleListSerializer(page.styles).data["used_live"] is published
+
+    def test_validate_revenue_program(self):
```
benjaminEwhite, line 598: opportunistically added since wasn't tested before.

spa/cypress/fixtures/pages/create-page-response.json
benjaminEwhite, line 1: Relevant change in this file is changing page slug to `null` to reflect real life value given changes in this PR.

spa/cypress/fixtures/pages/unpublished-page-1.json
benjaminEwhite, line 1: Relevant change here is slug going to `null` to reflect expected state for unpublished pages moving forward. Everything is automatically generated reformatting.

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.styled.ts
```diff
@@ -76,7 +79,11 @@ export const Paper = styled(MuiPaper)`
   }
 `;
 
-export const Icon = styled.div`
+interface IconProps {
```
benjaminEwhite, line 82: This was to get rid of an console.error i was seeing when running jest tests

spa/src/components/pageEditor/editInterface/pageSetup/PageSetup.test.js
```diff
@@ -22,7 +22,7 @@ const page = {
 
 function tree() {
   return render(
-    <PageEditorContext.Provider value={{ errors: {} }}>
+    <PageEditorContext.Provider value={{ errors: [] }}>
```
benjaminEwhite, line 25: This was to get rid of console.error i saw when running jest tests.

spa/cypress/e2e/03-donationPageList.cy.js
```diff
@@ -111,21 +111,7 @@ describe('Add Page modal', () => {
...
-        expect(request.body).to.eql({
-          name: 'Some Rev Program Page 1',
-          revenue_program: 1,
-          slug: 'some-rev-program-page-1'
-        });
-      });
-    });
-
-    it('immediately creates a page with a unique name and slug based on existing pages', () => {
+    it('immediately creates a page with a unique name based on existing pages', () => {
```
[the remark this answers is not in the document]
benjaminEwhite: This functionality is being retired, so removing unit test here. Specifically, we no longer create a temporary slug, and instead page gets created without a slug.

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.styled.ts
```diff
@@ -116,37 +123,56 @@ export const PublishButton = styled(MuiButton)`
...
-export const Input = styled.input`
-  width: 100%;
-  padding: 12px;
-  height: 40px;
-  border: 1px solid ${(props) => props.theme.colors.muiGrey[400]};
+interface InputProps {
+  start?: boolean;
+  center?: boolean;
+  end?: boolean;
+  readOnly?: boolean;
```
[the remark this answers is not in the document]
benjaminEwhite: i was getting console.errors when running jest tests at one point, but I can try in current implementation.

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.styled.ts
benjaminEwhite: Looks like `type="grey"` doesn't do anything. Removed in 7940331

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.styled.ts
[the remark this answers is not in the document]
benjaminEwhite: Without that i get typescript complaint.

> No overload matches this call.
  Overload 1 of 2, '(props: WithOptionalTheme<({ label?: ReactNode; slot?: string | undefined; select?: boolean | undefined; style?: CSSProperties | undefined; title?: string | undefined; ref?: Ref<...> | undefined; ... 282 more ...; end?: boolean | undefined; } & Partial<...>) | ({ ...; } & Partial<...>) | ({ ...; } & Partial<...>), DefaultTheme> & { ...; }): ReactElement<...>', gave the following error.
    Type '{ value: string; readOnly: true; start: true; }' is not assignable to type 'IntrinsicAttributes & (WithOptionalTheme<({ label?: ReactNode; slot?: string | undefined; select?: boolean | undefined; style?: CSSProperties | undefined; ... 284 more ...; end?: boolean | undefined; } & Partial<...>) | ({ ...; } & Partial<...>) | ({ ...; } & Partial<...>), DefaultTheme> & { ...; })'.
      Property 'readOnly' does not exist on type 'IntrinsicAttributes & (WithOptionalTheme<({ label?: ReactNode; slot?: string | undefined; select?: boolean | undefined; style?: CSSProperties | undefined; ... 284 more ...; end?: boolean | undefined; } & Partial<...>) | ({ ...; } & Partial<...>) | ({ ...; } & Partial<...>), DefaultTheme> & { ...; })'.
  Overload 2 of 2, '(props: StyledComponentPropsWithAs<ForwardRefExoticComponent<(Pick<StandardTextFieldProps, "label" | "slot" | "select" | "style" | "title" | "type" | "name" | "key" | "defaultChecked" | "defaultValue" | ... 274 more ... | "SelectProps"> | Pick<...> | Pick<...>) & RefAttributes<...>>, ... 4 more ..., ForwardRefExoticComponent<...>>): ReactElement<...>', gave the following error.
    Type '{ value: string; readOnly: true; start: true; }' is not assignable to type 'IntrinsicAttributes & StyledComponentPropsWithAs<ForwardRefExoticComponent<(Pick<StandardTextFieldProps, "label" | "slot" | "select" | "style" | "title" | "type" | "name" | ... 277 more ... | "SelectProps"> | Pick<...> | Pick<...>) & RefAttributes<...>>, ... 4 more ..., ForwardRefExoticComponent<...>>'.
      Property 'readOnly' does not exist on type 'IntrinsicAttributes & StyledComponentPropsWithAs<ForwardRefExoticComponent<(Pick<StandardTextFieldProps, "label" | "slot" | "select" | "style" | "title" | "type" | "name" | ... 277 more ... | "SelectProps"> | Pick<...> | Pick<...>) & RefAttributes<...>>, ... 4 more ..., ForwardRefExoticComponent<...>>'.ts(2769)

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.tsx
```diff
@@ -68,28 +83,31 @@ const PublishModal = ({ open, onClose, onPublish, page, loading }) => {
               <Label>Page Name</Label>
             </Grid>
             <Grid item xs={4}>
-              <Input value={page?.revenue_program?.slug} start="true" readOnly />
+              <Input value={page?.revenue_program?.slug} readOnly start />
+              <UnderText>*Site name can’t be changed upon publish.</UnderText>
```
nrh-cklimas, line 87: IMHO the layout of this modal has gotten more complicated than it needs to be. I'd like to create a followup story to address this if that's OK with you, Ben.

spa/src/components/common/Button/PublishButton/PublishModal/__mocks__/PublishModal.tsx
```diff
@@ -1,9 +1,12 @@
-const PublishModal = ({ onClose, onPublish, open }) => (
+import { PublishModalProps } from '../PublishModal';
+
+export const PublishModal = ({ onClose, onPublish, open, slugError }: PublishModalProps) => (
   <>
     {open && (
       <div data-testid="mock-publish-modal">
         <button onClick={() => onPublish({ slug: 'mock-slug' })}>onPublish</button>
         <button onClick={onClose}>PublishModal onClose</button>
+        <div>Slug error:{slugError?.[0]}</div>
```
nrh-cklimas, line 9: Style thing: prefer `<div data-testid="slug-error">` because it's easier to locate in a test.

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.styled.ts
```diff
@@ -116,37 +123,56 @@ export const PublishButton = styled(MuiButton)`
...
+}
+
+export const Input = styled(TextField).withConfig({
+  // This is here to avoid downstream complaints about the `start`, `center`, and `end` props
+  // being passed to TextField
+  shouldForwardProp: (prop: any) => !['start', 'center', 'end'].includes(prop)
+})<InputProps>`
```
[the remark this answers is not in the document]
benjaminEwhite: Done in 1f5a687a5

nrh-cklimas (changes requested):

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.tsx
```diff
@@ -68,28 +83,31 @@ const PublishModal = ({ open, onClose, onPublish, page, loading }) => {
               <Label>Page Name</Label>
             </Grid>
             <Grid item xs={4}>
-              <Input value={page?.revenue_program?.slug} start="true" readOnly />
+              <Input value={page?.revenue_program?.slug} readOnly start />
```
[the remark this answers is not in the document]
benjaminEwhite: Done in 1f5a687a5

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.tsx
```diff
@@ -68,28 +83,31 @@ const PublishModal = ({ open, onClose, onPublish, page, loading }) => {
...
+              />
             </Grid>
             <Grid item xs={5}>
               <Input
+                end
+                error={!!slugError?.length}
                 data-testid="page-name-input"
                 placeholder="Ex. contribute, donate, join"
-                end="true"
+                helperText={slugError?.length ? slugError.join('. ') : ''}
```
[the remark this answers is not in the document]
nrh-cklimas: I agree that it doesn't make sense to hide errors. This feels like a followup story to standardize this across the board. TBH I haven't seen multiple errors on a field in practice.

spa/src/components/common/Button/PublishButton/PublishButton.tsx
```diff
@@ -148,7 +151,12 @@ function PublishButton({ className }: PublishButtonProps) {
 
       handleOpenSuccessfulPublishModal();
     } catch (error) {
-      alert.error(GENERIC_ERROR);
+      // If there is a slug error, we pass it to the modal.
+      if ((error as AxiosError).response?.data?.slug) {
+        setSlugError((error as AxiosError).response?.data?.slug);
```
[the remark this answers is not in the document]
benjaminEwhite: I tried that initially, thinking typescript would be okay given line 155, but i don't think it's quite clever enough. Going with suggestion causes:

> Object is possibly 'undefined'.ts(2532)
> (local var) error: unknown
> No quick fixes available

spa/src/components/common/Button/PublishButton/PublishModal/__mocks__/PublishModal.tsx
benjaminEwhite: Fine by me. My "armchair" understanding of these things is that it's less in keeping with testing library's bent toward ["testing the application components in the way the user would use it"](https://testing-library.com/docs/guiding-principles) since user won't inspect data attributes but will find elements based on text.

that said, this is a mock that an end user won't ever see anyways, so that principle seems less important in this context.

And as it turns out, there's not really a need for either the original `Slug error:` or `data-testid` here given current test implementation, so I just removed `Slug error:` without adding test id.

a7836c6b6f6

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.tsx
```diff
@@ -22,27 +22,42 @@ import {
...
+  page: PropTypes.shape(PagePropTypes).isRequired,
+  loading: PropTypes.bool,
+  slugError: PropTypes.arrayOf(PropTypes.string)
+};
+
+export interface PublishModalProps extends InferProps<typeof PublishModalPropTypes> {
+  onClose: () => void;
+  onPublish: ({ slug }: { slug: string }) => void;
+  page: ContributionPage;
+  slugError?: string[] | null;
```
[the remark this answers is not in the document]
benjaminEwhite: Was able to simplify to `slugError: string[]` by setting slugError's initial value to `[]` in parent context. 

bab1a4ea

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.tsx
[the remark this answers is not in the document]
benjaminEwhite: Created follow up ticket: https://news-revenue-hub.atlassian.net/browse/DEV-3973

njharman (approved):

spa/src/components/common/Button/PublishButton/PublishButton.tsx
```diff
@@ -69,6 +70,8 @@ function PublishButton({ className }: PublishButtonProps) {
     handleOpen: handleOpenUnpublishModal
   } = useModal();
   const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
+  const [slugError, setSlugError] = useState<string[] | null>(null);
```
GuiMend, line 73: A cleaner way of doing this is:

```suggestion
  const [slugError, setSlugError] = useState<string[]>();
```

spa/src/components/common/Button/PublishButton/PublishButton.tsx
```diff
@@ -69,6 +70,8 @@ function PublishButton({ className }: PublishButtonProps) {
     handleOpen: handleOpenUnpublishModal
   } = useModal();
   const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
```
GuiMend, line 72: Same here:

```suggestion
  const [anchorEl, setAnchorEl] = useState<HTMLElement>();
```

It will be undefined instead of null, but I don't think this is an issue?

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.styled.ts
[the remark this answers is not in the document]
GuiMend: Got it. 👍 
You can leave it your way

spa/src/components/common/Button/PublishButton/PublishButton.tsx
benjaminEwhite: Don't disagree, but this is otherwise untouched code, so would prefer not to make change.

spa/src/components/common/Button/PublishButton/PublishButton.tsx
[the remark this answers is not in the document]
nrh-cklimas: This works but debatable as to whether it's actually better. I'll leave it up to you.

```tsx
const axiosError = error as AxiosError;

// If there is a slug error, we pass it to the modal.
if (axiosError.response?.data?.slug) {
  setSlugError(axiosError.response.data.slug);
} else {
  alert.error(GENERIC_ERROR);
}
```

spa/src/components/common/Button/PublishButton/PublishModal/PublishModal.tsx
nrh-cklimas: Works for me.

spa/src/components/common/Button/PublishButton/PublishModal/__mocks__/PublishModal.tsx
nrh-cklimas: To me the usefulness of a data-testid is to prevent a false positive, as unlikely as it is. This isn't a blocking issue for me.

spa/src/components/common/Button/PublishButton/PublishButton.tsx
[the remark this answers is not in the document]
benjaminEwhite: no real preference, so will prefer laziness and keep as is 🤷

spa/src/components/common/Button/PublishButton/PublishModal/__mocks__/PublishModal.tsx
benjaminEwhite: I deferred to your implicitly trusted front end preferences :)

nrh-cklimas (approved):

GuiMend (approved):

x110dc (approved):

PR #1236, opened 2023-09-14
ITS-2024: Only run tests when relevant code changes

nrh-cklimas:

n/a

#### What's this PR do?

Adds logic so that unit tests only run when the related code changes.

Faster CI, uses fewer GH Actions minutes.

n/a

No.

n/a, CI-related only

Not needed.

No.

No.

https://news-revenue-hub.atlassian.net/browse/ITS-2024

No.

No.


.github/workflows/test-python.yml
```diff
@@ -6,7 +6,30 @@ on:
...
+        with:
+          fetch-depth: 0
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            django:
+              - 'apps/**'
+              - 'public/**'
+              - 'revengine/**'
```
benjaminEwhite, line 29: I don't know exactly how this works, but does this mean we would not get Python tests running if the following files (located at repo root folder) change?

- conftest.py
- manage.py
- poetry.lock
- pyproject.toml
- pytest.ini

If any of those changes, I think we want to run tests, even if no other python code has changed.

benjaminEwhite (changes requested):
Overall seems good. Think we may be filtering out some files at root level for which we should run Python tests if changes.

.github/workflows/test-python.yml
nrh-cklimas: That's correct--thanks for the feedback! I added those in c0beed5711c3931e1095d77c597fd1e5ec524810.

benjaminEwhite (approved):

x110dc (approved):

PR #1244, opened 2023-09-20
DEV-3882: Create new SPA router for portal, implement magic link login

GuiMend:

n/a
#### What's this PR do?
- Create new router `PortalRouter` component to handle all new portal routes.
- Create usePortal hook to handle all api requests
- Create new Magic Link screen for the new portal
- Freeze storybook date so that it stops flagging "new/modified stories" in github CI when it's just rendering a new date.
- Create helper function to identify when accessing new portal routes (isPortalAppPath)

- Improving the overall user experience for accessing portal.

yes

Yes

yes

New Portal available at ".../portal/" url.

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-3882

no

Parent ticket: https://news-revenue-hub.atlassian.net/browse/DEV-3780


spa/src/components/base/TextField/TextField.tsx
```diff
@@ -37,6 +37,8 @@ const StyledMuiTextField = styled(MuiTextField)`
     }
 
     .NreTextFieldInput {
+      /* Needed for custom styles */
+      background-color: white;
```
nrh-cklimas, line 41: Should use theme.

spa/src/components/DashboardRouter.tsx
```diff
@@ -35,6 +35,8 @@ function DashboardRouter() {
 
   return (
     <RouterSetup>
+      <SentryRoute path={Object.values(ROUTES.PORTAL)} render={() => <TrackPageView component={PortalRouter} />} />
```
nrh-cklimas, line 38: Prefer to move this into MainLayout--doesn't make sense to me at first glance to have this remain a child of DashboardRouter.

spa/src/components/PortalRouter.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return data;
+  } catch (error) {
+    // Log it for Sentry and rethrow, which should cause the generic error
+    // message to appear.
+    console.error(error);
+    throw error;
+  }
+}
+
+function PortalRouter() {
```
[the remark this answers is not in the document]
GuiMend: > If we're repeating SegregatedStyles and DonationPageHeader on each route, let's make a PortalPage component that is responsible for that 

✅ 

> Can we get away without fetching the page here? It's hard for me to tell. I'd like to avoid it because it adds overhead to every pageload, e.g. people will get white screens while the request completes. Would much prefer to delegate it into routes, even if it means showing a loading spinner.

Asked for clarification on slack DM, but I don't think it's possible as we need the page fetched to get custom styles

> Any reason we can't use useContributionPage() to fetch the page?

We are getting the information from `/pages/live-detail/` in this page and in the `useContributionPage` hook we only fetch from `/pages/id` or `/pages/draft-detail`. If you mean update the hook to handle this, yes I can do that. But as it currently is, we cannot just plug&play.

spa/src/components/DashboardRouter.tsx
GuiMend: If that's the case we would need to create an entirely new router as we have with `ContributorRouter`.

I agree that it doesn't sound right to have the `PortalRouter` inside `DashboardRouter`, but it would be true as well for  `ContributorRouter` (which is also inside Dashboard). It would be way simpler to move `ContributorRouter` to MainLayout.

I think this change would probably be outside the scope of this ticket. We could have a new ticket to solve some of the tech debt we have with our Routers or I do it all in here and we have an even larger ticket than what it is.

spa/src/components/PortalRouter.tsx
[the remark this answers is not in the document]
nrh-cklimas: > We are getting the information from /pages/live-detail/ in this page and in the useContributionPage hook we only fetch from /pages/id or /pages/draft-detail. If you mean update the hook to handle this, yes I can do that. But as it currently is, we cannot just plug&play.

I see--I think it would be good in that case to move retrieval to a separate hook that doesn't return updatePage or deletePage, so it's clear that when you're using the unauthenticated endpoint, you can only GET. If you see a clean way to do that by modifying useContributionPage instead, I'm OK with that, but I fear it'd mean adding a lot of conditional statements across the app.

spa/src/components/base/TextField/TextField.tsx
nrh-cklimas: Saw the update but grey4 is #f9f9f9, not white, so I'm confused.

spa/src/components/DashboardRouter.tsx
nrh-cklimas: I agree, I don't think we should touch anything in ContributorRouter in this PR.

I don't fully understand what you mean by "creating an entirely new router". What does that entail, beyond wrapping it in RouterSetup? I'm looking at DonationPageRouter as a model we could use.

spa/src/components/base/TextField/TextField.tsx
GuiMend: My mistake. I was talking to Rachel and grey4 is the background color for the new portal and I added it here as well

spa/src/components/DashboardRouter.tsx
GuiMend: > I don't fully understand what you mean by "creating an entirely new router". What does that entail, beyond wrapping it in RouterSetup? I'm looking at DonationPageRouter as a model we could use.

You are right, I didn't mean it as a lot of work, I just meant that it would keep on following the pattern of having multiple different router components instead of only one (which I was trying to implement when I added the portal routes inside the main router). I'm not against making your suggestion, will do it now 👍

spa/src/components/PortalRouter.tsx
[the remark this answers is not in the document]
GuiMend: Done that 👍

spa/src/components/donationPage/DonationPageHeader/DonationPageHeader.tsx
nrh-cklimas, line 1: I'd like to avoid changing this component because there are effects to live contribution pages and widens our QA scope. What I'd suggest is we conditionally render DonationPageHeader in PortalPage, only after the page has been fetched.

If this causes an ugly layout shift or if the behavior is different than the contribution page--I'm not sure if the placeholder NRE logo is only meant to appear while the page is loading, or if the page has no custom image set--then I think a separate component is better to use here.

spa/src/components/portal/PortalEntry.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <Title>Welcome to the {page?.revenue_program?.name ?? 'RevEngine'} Contributor Portal</Title>
+        {magicLinkIsSuccess ? (
+          <Confirmation>
+            <h2>Email Sent!</h2>
+            <p>
+              An email has been sent to you containing your magic link. Click on your magic link to view your
+              contributions.
+            </p>
+            <p>
+              <br />
```
nrh-cklimas, line 42: This extra spacing doesn't match Figma.

spa/src/components/portal/PortalEntry.tsx
```diff
@@ -0,0 +1,84 @@
...
+                name="email"
+                control={control}
+                rules={{
+                  pattern: {
+                    value: /\S+@\S+\.\S+/,
+                    message: 'Please enter a valid email address'
+                  }
+                }}
+                render={({ field }) => (
+                  <TextField
```
nrh-cklimas, line 63: Let's add `type="email"` to get nicer browser UX.

spa/src/components/portal/PortalEntry.styled.ts
```diff
@@ -0,0 +1,91 @@
...
+  `}
+`;
+
+export const Content = styled.section`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  margin-top: 4rem;
+  padding: 2rem;
+  max-width: 693px;
```
nrh-cklimas, line 40: This doesn't seem to be rendering at 693px in practice--for me in the review app, I got 633px instead.

nrh-cklimas (changes requested):

spa/src/components/donationPage/DonationPageHeader/DonationPageHeader.tsx
GuiMend: From what I understood the default NRE logo should appear if no custom styles.
I removed it from DonationPageHeader 👍

spa/src/components/portal/PortalEntry.styled.ts
GuiMend: The padding was taking space from the 693px. Thanks for catching that.

spa/src/components/portal/PortalEntry.tsx
nrh-cklimas: I see an extra line break in the review app.
Screenshot 2023-09-25 at 5 51 11 PM

nrh-cklimas (approved):

PR #1245, opened 2023-09-20
DEV-3960 | Add language to donation page model modify page limit validation to account for language

benjaminEwhite:

#### What's this PR do?

This PR adds `locale` to the DonationPage model. At the model level we only enforce max length (no enforcement of locale choices from enum).

Then, in DonationPageFullDetailSerializer, we update the logic around validating page and publish limits such that locale is taken into account.

On create/update, we allow pages to have one of two values for locale: "en" or "es". 

Note that we also have changed the style limit on all plans to `UNLIMITED_CEILING`.  

The reason for this is that we are no longer allow users to create styles distinct from pages. With the introduction of page locales, an org can have up to 2x the number of pages in page limit, and 2x the number of published pages in published page limit (if half the pages are english and half Spanish).  We decided to retain UNLIMITED_CEILING limit vs having no limit at all to protect against a bad actor deciding to generate an excessive amount of styles (even though they could only do this via API), while allowing us to reduce complexity around the introduction of page locale.

Better support orgs with non-English and/or bi/multi-lingual readership.

Forthcoming

No

Yes

N/A

No

N/A

- [DEV-3960](https://news-revenue-hub.atlassian.net/browse/DEV-3960)
- [DEV-4021](https://news-revenue-hub.atlassian.net/browse/DEV-4021)

No

No

[DEV-3960]: https://news-revenue-hub.atlassian.net/browse/DEV-3960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/pages/models.py
```diff
@@ -49,39 +49,36 @@ class DonationPage(IndexedTimeStampedModel):
     A DonationPage represents a single instance of a Donation Page.
     """
 
-    name = models.CharField(max_length=PAGE_NAME_MAX_LENGTH)
-    heading = models.CharField(max_length=PAGE_HEADING_MAX_LENGTH, blank=True)
+    elements = models.JSONField(null=True, blank=True, default=defaults.get_default_page_elements)
```
benjaminEwhite, line 52: FYI, i sorted the model attributes alphabetically so easier to find what you're looking for.

The only non-sorting change here is that we add `locale` on line 60.

apps/pages/tests/test_views.py
```diff
@@ -502,39 +508,50 @@ def test_create_page_when_invalid_because_style_not_exist(self, user, style, pag
...
 
-    @pytest_cases.parametrize("plan", (Plans.FREE.value, Plans.CORE.value, Plans.PLUS.value))
-    def test_create_when_already_at_publish_limit(self, plan, hub_admin_user, api_client):
+    # NB we don't test on plus plan because it is meant to have unlimited pages (published or otherwise)
+    @pytest_cases.parametrize("plan", (Plans.FREE.value, Plans.CORE.value))
+    @pytest_cases.parametrize("locale", (SpanishLocale, EnglishLocale))
+    def test_create_when_already_at_publish_limit(self, plan, locale, hub_admin_user, api_client):
         rp = RevenueProgramFactory(organization=OrganizationFactory(plan_name=plan))
-        for i in range((limit := rp.organization.plan.page_limit)):
+        for i in range((limit := rp.organization.plan.publish_limit)):
```
benjaminEwhite, line 536: Looks like the test in the original was actually testing around page limit, not publish limit :( notwithstanding the test name.

apps/pages/serializers.py
```diff
@@ -299,16 +333,19 @@ def validate_publish_limit(self, data):
...
         )
         if DonationPage.objects.filter(
-            published_date__isnull=False, revenue_program__organization=org
+            published_date__isnull=False,
+            revenue_program__organization=org,
+            locale=locale.code,
         ).count() + offset > (pl := org.plan.publish_limit):
             logger.info(
-                "DonationPageFullDetailSerializer.validate_publish_limit raising ValidationError because org (%s) has reached its publish limit",
+                "DonationPageFullDetailSerializer.validate_publish_limit raising ValidationError because org (%s) has reached its publish limit for %s",
```
njharman, line 341: log, not big deal, seems weird grammer?   "... publish limit for English".   

- "...English publish limit"
- "publish limit for English pages"

apps/pages/serializers.py
```diff
@@ -261,19 +280,33 @@ def validate_thank_you_redirect(self, value):
             )
         return value
 
+    def valiate_locale(self, value):
+        if value not in LOCALE_MAP:
+            raise serializers.ValidationError({"locale": f"Invalid locale: {value}"})
+        return value
```
njharman, line 286: Uncovered by unittests

njharman (changes requested):
one uncovered method

apps/pages/serializers.py
```diff
@@ -261,19 +280,33 @@ def validate_thank_you_redirect(self, value):
...
         .validate() below.
         """
+        locale = self._get_locale(data)
         if self.context["request"].method != "POST":
             return
         if DonationPage.objects.filter(
-            revenue_program__organization=(org := data["revenue_program"].organization)
+            revenue_program__organization=(org := data["revenue_program"].organization),
+            locale=(locale.code),
         ).count() + 1 > (pl := org.plan.page_limit):
```
njharman, line 303: We log publish limit, should we not log here too?

apps/pages/serializers.py
benjaminEwhite: Would help if I spelled val**_i_**date correctly 🤦

benjaminEwhite: f6a0df48181

apps/pages/serializers.py
benjaminEwhite: 424d225494a

apps/pages/serializers.py
benjaminEwhite: 9176b71e9150

apps/pages/serializers.py
```diff
@@ -299,16 +340,20 @@ def validate_publish_limit(self, data):
             else 0
         )
         if DonationPage.objects.filter(
-            published_date__isnull=False, revenue_program__organization=org
+            published_date__isnull=False,
+            revenue_program__organization=org,
+            locale=locale.code,
```
njharman, line 345: I don't think it matters.  Mentioning in case you expected otherwise.

This branch is always taken, the if on #342 is always true.

apps/pages/serializers.py
```diff
@@ -261,19 +280,40 @@ def validate_thank_you_redirect(self, value):
             )
         return value
 
+    def validate_locale(self, value):
+        if value not in LOCALE_MAP:
+            raise serializers.ValidationError({"locale": f"Invalid locale: {value}"})
```
njharman, line 285: Uncovered raise. if on #284 was never true.

njharman (approved):
There's two branches and one raise that is uncovered.  I'm not super stressed about that. If you are , you can fix it. Otherwise, ship it!

apps/pages/serializers.py
benjaminEwhite: 56b8d3cc2

x110dc (approved):
LGTM.

PR #1274, opened 2023-10-12
DEV-4130 | mock implementation of new contributor contributions endpoint for new contributor portal

benjaminEwhite:

#### What's this PR do?

This PR adds a new endpoint at `/api/v1/contributors/<contributor-id>/contributions/`.  Eventually, this endpoint will supply real data for the new contributor portal, but in short term, to unblock front end development but allow it to be against actual server, we're configuring this endpoint to return fake data.  Additionally, the endpoint in short term is not authenticated, as authentication would bring magic email links into path here, and point is to do as little as possible to unblock front end, while providing it with representative data.

Allows us to move forward on front end work for new contributor portal while parallel changes to server-side data modeling around payments is completed, so that we don't have to build new contributor portal api on top of soon-to-be-retired caching layer around contributions that's currently used (which would have been annoying/difficult because we'd need to create custom pagination, sorting, etc.).

Yes

No

A minimal test has been added that shows the mocked endpoint stands up and accepts expected query params. Beyond that, we are not testing this functionality and will do that later when we implement the real endpoint.

N/A

No

No

- [DEV-4130](https://news-revenue-hub.atlassian.net/browse/DEV-4130)

No

N/A

[DEV-4130]: https://news-revenue-hub.atlassian.net/browse/DEV-4130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/views.py
```diff
@@ -567,3 +572,99 @@ def _re_retrieve_pi_and_insert_pi_and_sub_into_cache(
...
+    payment_type: str
+    provider_customer_id: str
+    revenue_program: int
+    status: str
+
+
+fake = Faker()
+
+
+def get_mock_contributor_contributions_results(
```
nrh-cklimas, line 599: I appreciate the idea behind this in that it'll give us all kinds of different data, but I think this might be problematic in practice. Because react-query re-fetches at various intervals, during testing we could see the UI change completely on us midstream, because I think every time I hit the endpoint, I'll get totally different results. Same for testing filtering/sorting--each resultset will be internally consistent, but will be confusing to test.

I think just a static set of data would work better here, or if you want to get fancy here, maybe use the contributor ID as the random seed with faker.

nrh-cklimas (changes requested):

apps/contributions/views.py
benjaminEwhite: That makes sense. I stored output from initial implementation as JSON fixture. We now load that fixture and then reserialize it (so we have a better guarantee that our serializer upholds contract).

f2b07581f4e4

nrh-cklimas (approved):

apps/contributions/serializers.py
```diff
@@ -929,3 +929,29 @@ def get_customer_id(self, instance):
 
     def get_payment_type(self, instance):
         return instance.default_payment_method.type
+
+
+class ContributionAgreementSerializer(serializers.Serializer):
```
x110dc, line 934: I'm not a fan of this name. I think it's confusing. But I guess it's just temporary?

x110dc (approved):
Just one comment.

apps/contributions/serializers.py
benjaminEwhite: I would like a way to systematically distinguish between "common parlance" of what a contribution means (which is a single instance of someone giving money) vs. the intention to give on an ongoing basis. The latter is not itself a "contribution" even though that's how we've been referring to it in our system. 

To my mind, this name captures exactly what a user is doing when making a one time or recurring gift. In both cases, they're agreeing to send money to the org, either once or recurrently.

If there's a more appealing way to refer to this, I'm all ears, but I would like to have the name we choose reflect this nuance (in a way that, say, `ContributionSerializer` does not).

PR #1275, opened 2023-10-12
DEV 3107 uwsgi config changes

x110dc:

#### What's this PR do?

- moves the uwsgi configuration from environment variables in the Dockerfile to an INI-style config file
- installs the uwsgi monitoring app uwsgitop
- upgrades uwsgi to the latest release
- adds new uwsgi configuration options: stop logging requests (except 4xx and 5xx); log details about harikari

- it seems clearer to have all the uwsgi configuration in one file in the same format you usually find advise about how to configure it
- this also simplifies the Dockerfile
- uwsgitop may help us diagnose why workers are being killed 
- logging every request pollutes the logs with information that isn't very valuable; this makes the logs easier to read

Not yet.

N/A

N/A

N/A

N/A

N/A

https://news-revenue-hub.atlassian.net/browse/DEV-3107

No.

N/A


docker-entrypoint.sh
```diff
@@ -18,4 +18,4 @@ fi
 
 /bin/sh ./google_sa.sh
 
-exec uwsgi --http=0.0.0.0:$PORT --show-config
+exec uwsgi --ini ./uwsgi.conf --http=0.0.0.0:$PORT --show-config
```
benjaminEwhite, line 21: Looks like `--show-config` here is duplicative of `show-config = true` in uwsgi.conf.

pyproject.toml
```diff
@@ -81,6 +81,7 @@ mailchimp-marketing = "^3.0.80"
 google-cloud-secret-manager = "^2.16.0"
 django-request-id = "^1.0.0"
 pydantic = "^2.2.0"
+uwsgitop = "^0.11"
```
benjaminEwhite, line 84: AFAICT, we don't start uwsgitop in the container. Is the plan to do that in a follow up release, or is that something that will be done manually from terminal?

benjaminEwhite (changes requested):
Left one small suggestion and a question about starting uwsgitop.

docker-entrypoint.sh
x110dc: Good catch. I'll remove here.

x110dc: https://github.com/newsrevenuehub/rev-engine/commit/eb275ef41

benjaminEwhite (approved):

PR #1285, opened 2023-10-18
WIP for DEV-4313

GuiMend:

n/a
#### What's this PR do?
- Adds request_id to celery logs
- Adds configuration to make request_id enabled for services that can’t generate the X-Request-Id header at the web server level (GKE/Google)

- Adds request_id to missing places, helps debugging/tracking issues.

n/a

no

- Yes, but they are currently being skipped as they are only passing locally and failing in CI 

n/a

no

n/a

https://news-revenue-hub.atlassian.net/browse/DEV-3207

Yes, https://news-revenue-hub.atlassian.net/browse/ITS-2125

no


apps/common/__init__.py
```diff
@@ -0,0 +1,4 @@
+from apps.common.logger import setup_logs
+
+
+setup_logs()
```
benjaminEwhite, line 4: If i'm reading `setup_logs` correctly, this function is not about setting up logs in general, but specifically about setting up logging handlers for Celery tasks.

This code gets invoked in apps.common.__init__.py, which means it will run the first time a Python session executes code that imports from apps.common (and not on subsequent runs imports in the sesion).

But since this function is narrowly about handling the case of code running in Celery task (not in web app), I'm not sure we need to be running it in the general case of importing from apps.common.

I wonder if we might instead be able to call this (and possibly define it) in `revengine.celery.py`.  The function name could stay the same (and within that module's name space, it would better align with its behavior of narrowly setting up log handling for celery).

If we call `setup_logs` at module level in `revengine.celery.py`, I believe it would execute when a Celery worker process first starts.

apps/common/logger.py
```diff
@@ -0,0 +1,37 @@
+import contextvars
```
benjaminEwhite, line 1: See above. I think this code maybe should live in `revengine.celery.py`. My guess is that `RequestIdFilter` doesn't have broader applications than celery task logs, but if it does, it could continue to live in this file.

Otherwise, `set_request_id` and `setup_logs` could go in `revengine.celery.py`

apps/common/logger.py
```diff
@@ -0,0 +1,37 @@
...
+        return True
+
+
+@task_prerun.connect
+def set_request_id(*args, **kwargs):
+    req_id = str(uuid.uuid4())[:8]
+    request_id.set(req_id)
+
+
+def setup_logs():
```
benjaminEwhite, line 29: Would be good to have a docstring here detailing what the end result of this setup is (namely, what we're trying to accomplish in our celery task logs). A year from now, it will be hard to suss out that intentionality from just reading through implementation here.

benjaminEwhite (changes requested):

apps/common/logger.py
benjaminEwhite: Also, wouldn't be a bad idea to unit test this function.

revengine/celery.py
```diff
@@ -22,3 +27,51 @@
 @app.task(bind=True)
 def debug_task(self):
     print(f"Request: {self.request!r}")
+
+
+### Setup Request Id in Celery Logs
+
+LOGGING_FORMAT = "%(levelname)s %(request_id)s %(module)s:%(lineno)d - %(message)s"
+
+request_id = contextvars.ContextVar("request_id", default=str(uuid.uuid4())[:8])
```
benjaminEwhite, line 36: Can you give some context for use of `ContextVar` here? What is the lifecycle of `request_id` , and do we really need this module level var?

revengine/tests/test_celery.py
```diff
@@ -1,5 +1,30 @@
...
+        setup_celery_logs()
+
+        # Log a warning message
+        logger.warning("This is a warning")
+
+        # Get the log output captured by sys.stderr
+        log_output = mock_stderr.write.call_args[0][0].strip()
+
+        # Assert if the log output matches the expected format
+        expected_output = f"WARNING [{request_id.get()}]  test_celery:20 - This is a warning"
```
benjaminEwhite, line 29: This should be pytest-based, not unittest.

```python
class TestLoggingConfiguration:
    def test_log_format(self, mocker):
        mock_stderr = mocker.patch("sys.stderr")
        ....

```

Also, might as well set line 23 above's message to a var, reference that var there,  and then can use it in f-string at line 29 instead of hardcoding "This is a warning" here.

revengine/celery.py
GuiMend: I'm not sure how to answer that @benjaminEwhite. I was adapting the implementation we have in switchboard to revengine: https://news-revenue-hub.atlassian.net/browse/DEV-1219

benjaminEwhite: Okay. In that case, let's not block with this. But let's leave this conversation around for when @x110dc does his review, and he can weigh in.

I'd like better understand what problem we're trying to solve by using contextvars.

revengine/tests/test_celery.py
```diff
@@ -1,5 +1,29 @@
...
+
+        # Log a warning message
+        logger.warning(message)
+
+        # Get the log output captured by sys.stderr
+        log_output = mock_stderr.write.call_args[0][0].strip()
+
+        # Assert if the log output matches the expected format
+        expected_output = f"WARNING [{request_id.get()}]  test_celery:22 - [test_log_format] {message}"
+        assert expected_output in log_output
```
benjaminEwhite, line 29:
```suggestion
        assert f"WARNING [{request_id.get()}]  test_celery:22 - [test_log_format] {message}" in mock_stderr.write.call_args[0][0]
```

benjaminEwhite (changes requested):
Had one suggestion for simplifying test.

Also, looks like there's one untested path in `RequestIdFilter.filter`.

Screenshot 2023-10-24 at 6 44 18 PM

revengine/celery.py
```diff
@@ -1,8 +1,13 @@
+import contextvars
+import logging
 import os
+import uuid
 
 from django.conf import settings
 
 from celery import Celery
+from celery.app.log import TaskFormatter
+from celery.signals import task_prerun
```
benjaminEwhite, line 10: I know this is not code you touched, but I read lines 9-10 more closely, and I don't understand what it's doing. 

The Django app won't run if `DJANGO_SETTINGS_MODULE` is not set, and our workers run in a context where they have access to app-level settings. Across local, review apps, standing envs, and prod, we set this env var ahead of building image/ deploying etc.

So this code is reading in DJANGO_SETTINGS_MODULE, but expecting that sometimes it won't be set (I don't think there's a path to that happening in reality), only to then set the environment variable to a new value (revengine.settings.deploy) if it wasn't already set.

Not necessarily your responsibility to solve here, but I wonder if everything works if we remove this.

revengine/celery.py
```diff
@@ -22,3 +27,51 @@
 @app.task(bind=True)
 def debug_task(self):
     print(f"Request: {self.request!r}")
+
```
benjaminEwhite, line 30: Know it's untouched code and not necessarily in scope, but I wonder if we could remove `debug_task`. Only place i see it referenced is in its unit test. It uses `print` instead of logger, which i don't like (though maybe there are good reasons for that).

Seems like we could remove `debug_task` and `test_debug_task`.

revengine/celery.py
```diff
@@ -22,3 +27,51 @@
...
+request_id = contextvars.ContextVar("request_id", default=str(uuid.uuid4())[:8])
+
+
+class RequestIdFilter(logging.Filter):
+    def filter(self, record):
+        record.request_id = ""
+        req_id = request_id.get()
+        if req_id:
+            record.request_id = f"[{req_id}] "
+        return True
```
benjaminEwhite, line 45:
```suggestion
    def filter(self, record):
        req_id = request_id.get()
        record.request_id = f"[{req_id}]" if req_id else ""
        return True
```

revengine/celery.py
GuiMend: From what I understood `TaskFormatter` is formatting the log to display this information:
```python
"%(levelname)s %(request_id)s %(module)s:%(lineno)d - [%(funcName)s] %(message)s"
```

As for `task_prerun`, seems to me that it's setting a new request_id before a new task is started.

Did I get things wrong? 🤔

revengine/celery.py
njharman: @benjaminEwhite

revengine/tests/test_celery.py
```diff
@@ -1,5 +1,41 @@
...
 
+from revengine.celery import request_id, setup_celery_logs
 
-def test_debug_task():
-    debug_task()
+
+logger = get_task_logger("CELERY_TEST_LOGGER")
+
+
+# Tests works locally but fail in CI, so we skip them for now
```
benjaminEwhite, line 10: **blocking**

@GuiMend would like to understand more about why this is the case. Let's do that in Slack or Zoom because I have a feeling that will be easier. You can reach out to me when you're working on this.

revengine/tests/test_celery.py
```diff
@@ -1,5 +1,41 @@
...
+class TestLoggingConfiguration:
```
benjaminEwhite, line 11: **blocking, but trivial**

The PR description section about unit tests says N/A, but in fact unit tests are added but skipped. PR description should reflect this.

benjaminEwhite (changes requested):

GuiMend:
Celery logs will be addressed in https://news-revenue-hub.atlassian.net/browse/DEV-4313
AC1 of DEV-3207 will be done in https://github.com/newsrevenuehub/rev-engine/pull/1342

PR #1303, opened 2023-11-02
DEV-3959: add Spanish strings

nrh-cklimas:

#### What's this PR do?

1. Adds Spanish translations. ⚠️ There are some English strings in the file because we decided not to translate errors.
2. Adds a new component, `ContributionPageI18nProvider`, which initializes i18next with the page locale. See comments there on why we need a separate component and can't do it globally.
3. Wraps `DonationPage` with `ContributionPageI18nProvider` in `PageEditor` and `PublishedDonationPage`, which is a refactored version of `LiveDonationPageContainer` that puts page retrieval logic into a separate hook, `usePublishedPage()`.
4. Makes changes to contribution page components that were pointing at the global i18n instance to instead use the one provided by `useTranslation()`. I moved almost everything out of `parseFrequency` to make this easier. The one function left there is used by non-contribution page components.
5. Moves the mock i18n instance that global into an export from test-utils, so that it can be used in specific tests.
6. Removes `LiveLoading` which never was actually visible when a contribution page was loading and displayed placeholder text.

Completes the Spanish translation work.

Yes.

Need to document the language changes. Won't affect existing pages.

Yes.

Documentation.

I had to create somewhat duplicative code in Cypress (`expectedPaymentUrl`) to avoid needing to work with i18n as well there. I think this is a necessary evil because otherwise our E2E is going to be even more tangled up with the main code.

I also added duplicate code in `test-utils.tsx` because some modules need a i18n instance outside of React now. I tried importing the mock instance there in `setupTests.js`, but this caused an unrelated test to fail. I think this is because `test-utils` imports a bunch of other modules, specifically `useAnalyticsContext`, which in some tests we want to mock.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3959

No.

No.


GuiMend (approved):
Nice changes @nrh-cklimas 💯

PR #1320, opened 2023-11-16
DEV-3884: Implement new contributions list

nrh-cklimas:

#### What's this PR do?

1. Adds a list of contributions in the new contributor portal. Lots of functionality yet to come; see Jira for out of scope stuff.
2. Adds a new context, PortalAuthContext, and a component to manage it.
3. Uses a new component, TokenVerification, instead of the older ContributorVerify.
4. Adds a new hook, usePortalContributionList, to retrieve contributions from the new API endpoint. Existing hook is marked deprecated.
5. Adds a CircularProgress base component.
6. Migrates `formatCurrencyAmount` to TypeScript and adds parameter to optionally omit the cents in an amount if it's a round number.
7. Changes the name of the `PORTAL.DASHBOARD` route to `PORTAL.CONTRIBUTIONS`.

1. Makes progress toward new portal.
2. The existing approach on auth in the contributor portal relied on storing context in local storage. The new context stores it there, but provides a typed interface to consumers who need it (i.e. the contributions list needs to know the contributor ID). This will increase reliability/dev speed since consumers no longer need to use local storage directly and handle possible errors every time they access it.
3. TokenVerification uses the new context in point 2.
4. New hook keeps business logic separate/more testable. I marked the existing one deprecated to avoid confusion since the names are pretty similar.
5. It's a better UI experience to see part of the page show a spinner instead of the entire page graying out. This CircularProgress component allows that.
6. The new contributions list omits cents in round amounts.
7. I thought it was confusing to have a route in the contributor portal named "dashboard" when we are calling the admin area "the dashboard".

Yes.

Need updated docs around the new portal, but even merged, users will continue to be directed to the existing portal.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3884

No.

See rest of epic.

### DIFF
--- a/spa/src/hooks/usePortalAuth.tsx
+++ b/spa/src/hooks/usePortalAuth.tsx
@@ -1,0 +1,128 @@
+import PropTypes, { InferProps } from 'prop-types';
+import { createContext, useCallback, useContext, useEffect, useState } from 'react';
+import axios from 'ajax/axios';
+import { VERIFY_TOKEN } from 'ajax/endpoints';
+import { LS_CONTRIBUTOR, LS_CSRF_TOKEN } from 'appSettings';
+
+/**
+ * A contributor who has logged into the portal.
+ */
+export interface Contributor {
+  /**
+   * Timestamp when the contributor was created.
+   */
+  created: string;
+  /**
+   * Contributor's email address.
+   */
+  email: string;
+  /**
+   * Internal ID for the contributor.
+   */
+  id: number;
+  /**
+   * Timestamp when the contributor was modified.
+   */
+  modified: string;
+  /**
+   * Unique ID for the contributor
+   */
+  uuid: string;
+}
+
+/**
+ * Properties provided to consumers of usePortalAuthContext.
+ */
+export interface PortalAuthContextResult {
+  /**
+   * The current logged-in contributor. If undefined, the user isn't
+   * authenticated.
+   */
+  contributor?: Contributor;
+  /**
+   * Exchanges a token sent via magic link email for a JWT. If undefined, the
+   * user is already authenticated. If this succeeds, contributor is set in
+   * context and the function resolves. If it fails, it rejects with either an
+   * Axios error (if there were problems with the request itself) or an error
+   * with the response from the server.
+   */
+  verifyToken?: (email: string, token: string) => Promise<void>;
+}
+
+/**
+ * Expected API response when verifying a token.
+ */
+export interface VerifyTokenResponse {
+  contributor?: Contributor;
+  csrftoken?: string;
+  detail: string;
+}
+
+export const PortalAuthContext = createContext<PortalAuthContextResult>({});
+
+export const usePortalAuthContext = () => useContext(PortalAuthContext);
+
+const PortalAuthContextProviderProps = {
+  children: PropTypes.node.isRequired
+};
+
+export function PortalAuthContextProvider({ children }: InferProps<typeof PortalAuthContextProviderProps>) {
+  const [contributor, setContributor] = useState<Contributor>();
+  const verifyToken = useCallback(async (email: string, token: string) => {
+    const { data } = await axios.post<VerifyTokenResponse>(VERIFY_TOKEN, { email, token });
+
+    if (data.detail !== 'success') {
+      throw new Error(data.detail);
+    }
+
+    if (!data.contributor) {
+      throw new Error('No contributor in token verification response');
+    }
+
+    if (!data.csrftoken) {
+      throw new Error('No CSRF in token verification response');
+    }
+
+    // Set values in context and in local storage for later usage. The local
+    // storage key is also needed for backwards compat with isAuthenticated() in
+    // utilities/, used by ProtectedRoute.
+
+    setContributor(data.contributor);
+    localStorage.setItem(LS_CONTRIBUTOR, JSON.stringify(data.contributor));
+    localStorage.setItem(LS_CSRF_TOKEN, data.csrftoken);
+  }, []);
+
+  // Try to initally set the contributor based on local storage.
+
+  useEffect(() => {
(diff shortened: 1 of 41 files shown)


spa/src/hooks/usePortalContributionList.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export type CardBrand = 'amex' | 'diners' | 'discover' | 'jcb' | 'mastercard' | 'unionpay' | 'visa' | 'unknown';
+
+export interface PortalContribution {
+  /**
+   * Amount **in cents** of the contribution.
+   */
+  amount: number;
+  /**
+   * What payment card was used on the contribution.
```
benjaminEwhite, line 15: FWIW, this value is likely to be whatever the current payment method is, regardless of what was used on any given payment for a contribution.

spa/src/hooks/usePortalContributionList.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  is_modifiable: boolean;
+  /**
+   * Last four digits of the payment card used on the contribution.
+   */
+  last4: string;
+  /**
+   * Timestamp of when the last related payment occurred.
+   */
+  last_payment_date: string;
```
benjaminEwhite, line 46: Seems like this might be null sometimes. For instance, IIRC imported legacy subscriptions will initially be in a state where our system doesn't see a last payment date and there hasn't been a charge in new system.

spa/src/hooks/usePortalContributionList.ts
```diff
@@ -0,0 +1,98 @@
...
+  card_brand: CardBrand;
+  /**
+   * Timestamp of when the contribution was created.
+   */
+  created: string;
+  /**
+   * When the credit card used for the contribution will expire.
+   * @example "4/2024"
+   */
+  credit_card_expiration_date: string;
```
benjaminEwhite, line 26: This may be a red herring, but we should see if we ever expect to see contributions that are by check or otherwise don't have a cc.

spa/src/hooks/usePortalContributionList.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  last4: string;
+  /**
+   * Timestamp of when the last related payment occurred.
+   */
+  last_payment_date: string;
+  /**
+   * Timestamp of when the next related payment will occur.
+   */
+  next_payment_date: string;
```
benjaminEwhite, line 50: This can be null in case of one-time or canceled recurring.

spa/src/hooks/usePortalContributionList.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Internal ID of the payment in the payment processor, e.g. Stripe.
+   */
+  payment_provider_id: string;
+  /**
+   * How was the payment made?
+   */
+  payment_type: string;
```
benjaminEwhite, line 58: Would maybe expect an enumerated set of literals here.

spa/src/hooks/usePortalAuth.tsx
```diff
@@ -0,0 +1,128 @@
...
+      throw new Error('No CSRF in token verification response');
+    }
+
+    // Set values in context and in local storage for later usage. The local
+    // storage key is also needed for backwards compat with isAuthenticated() in
+    // utilities/, used by ProtectedRoute.
+
+    setContributor(data.contributor);
+    localStorage.setItem(LS_CONTRIBUTOR, JSON.stringify(data.contributor));
+    localStorage.setItem(LS_CSRF_TOKEN, data.csrftoken);
```
benjaminEwhite, line 92: Might make sense to key off of contributor ID, perhaps for both values here. reason being i think that in status quo the localstorage + auth settings is sometimes a source of oddities in when you log in as an org user and then separately log in as a hubadmin/superuser. Since we're keying off a static var name here, can get stale.

benjaminEwhite (changes requested):

spa/src/hooks/usePortalContributionList.ts
nrh-cklimas: There are 30 possible ones :( https://stripe.com/docs/api/payment_methods/object#payment_method_object-type

Do you think it's plausible to make these the ones we expect to see? I think that would be 

- card
- link
- us_bank_account?

spa/src/hooks/usePortalAuth.tsx
nrh-cklimas: This reminds me I needed to create a followup story. https://news-revenue-hub.atlassian.net/browse/DEV-4252

Because of this, I'll request leniency on this point as it maintains backwards compat.

spa/src/hooks/usePortalContributionList.ts
nrh-cklimas: If that happened, would this be null or an empty string (or something else)?

spa/src/hooks/usePortalContributionList.ts
nrh-cklimas: Updated comment in 3c9788004f2b0801618f8d98a786f5e21d654e51

spa/src/hooks/usePortalContributionList.ts
benjaminEwhite: I believe it would be `null`. It's not a case I've thought a lot about. Maybe we just worry about it later.

spa/src/hooks/usePortalContributionList.ts
benjaminEwhite: 30s a lot!

I'm not gung ho about a partial list, TBH.

I think we can leave as is and if we have any issues arise around payment type, then we can get more restrictive at that point.

benjaminEwhite (approved):

PR #1323, opened 2023-11-18
DEV-4213: pin deps, unpin dev deps

leo-dcfa:

N/A

#### What's this PR do?

- Pin regular dependencies
- Unpin dev dependencies
- Fix one file that was failing on pre-commit
- Fix test failures after deps updates - Stripe tests were generating dates with microseconds whilst datetime.datetime.fromtimestamp wasn't - leading to test failures, since the two values would be different. Setting the microseconds to 0 did the trick. Additionally, pydantic_core is no longer imported from pydantic, but as a standalone import.

This will make it so we can do a regular poetry update; pre-commit autoupdate once per sprint to keep deps up-to-date

Yes

N/A

N/A

N/A

N/A

N/A

[DEV-4213]

N/A

N/A

[DEV-4213]: https://news-revenue-hub.atlassian.net/browse/DEV-4213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/tests/test_serializers.py
```diff
@@ -7,6 +7,7 @@
 from django.utils import timezone
 
 import pydantic
+import pydantic_core
```
benjaminEwhite, line 10: **Question**

This change seems out of scope. What's reasoning?

apps/contributions/tests/test_stripe_contributions_provider.py
```diff
@@ -233,7 +233,7 @@ def test_card(self, payment_method, expected, pi_for_valid_one_time_factory, moc
     def test_last_payment_date_when_no_invoice(self, pi_without_invoice):
         assert StripePaymentIntent(pi_without_invoice).last_payment_date == datetime.datetime.fromtimestamp(
             int(pi_without_invoice.created), tz=datetime.timezone.utc
-        )
+        ).replace(microsecond=0)
```
benjaminEwhite, line 236: Unclear why this change and similar ones below required.

benjaminEwhite (commented):
Have a couple questions of small questions (the answer to which I'm assuming is that changes were required to make pre-commit happy, but want to understand).

apps/contributions/tests/test_serializers.py
leo-dcfa: I was also surprised with this one - when I ran `poetry update` the test that references `pydantic.pydantic_core.Url` started failing. I noticed that pydantic_core got bumped from `2.10.1`to `2.14.3`, check the documentation and imported directly instead.

apps/contributions/tests/test_stripe_contributions_provider.py
benjaminEwhite: **blocking request**

worth adding a code comment about why this is required.


**non-blocking question**

do you know why the test started failing now?

benjaminEwhite (approved):

apps/contributions/tests/test_stripe_contributions_provider.py
leo-dcfa: Comment added. 

Concerning why, I don't know, some dependency update led to it

pyproject.toml
```diff
@@ -90,7 +90,7 @@ bandit = "1.7.5"
 black = "23.9.1"
 colorama = "*"
 coverage = "*"
-dateparser = "^1.1.8"
+dateparser = "1.1.8"
```
x110dc, line 93: Why is this pinned and not set to `*`?

x110dc (changes requested):
One dep doesn't look correct. 

Adding more detail to the "what does this PR do" section I think could've avoided some of the back-and-forth for the non-obvious changes here.

x110dc (approved):

PR #1333, opened 2023-12-05
DEV-4253: Maintain Style Name in sync with page name

GuiMend:

n/a

#### What's this PR do?
- Changes the Update Style mutation. Instead of just passing the style changes, also added `page.name`. This updates the style name on page updates that contain style changes (avoiding any duplicate name issue)
- Create an if statement to check if style name and page name aren't in sync when updating the page. If not in sync, update style name, if in sync continue page update as usual.

- Solves issue with creating several pages and styles and eventually getting to a state where a style name is duplicated and the user is unable to save style and/or page updates.

yes

no

yes

n/a

no

n/a

https://news-revenue-hub.atlassian.net/browse/DEV-4253

no

no


spa/src/hooks/useStyleList.tsx
```diff
@@ -76,21 +76,28 @@ function useStyleList(): UseStyleListResult {
   };
 
   const updateStyleMutation = useMutation(
-    (styles: Partial<Style>) => {
+    ({ styles, page }: { styles: Partial<Style>; page?: ContributionPage }) => {
```
nrh-cklimas, line 79: Instead of throwing an error here, let's make `page` required.

spa/src/hooks/useStyleList.tsx
```diff
@@ -76,21 +76,28 @@ function useStyleList(): UseStyleListResult {
...
+      });
     },
     {
       onError: onSaveStylesError
     }
   );
 
-  const updateStyle = async (styles: Partial<Style>) => {
-    return await updateStyleMutation.mutateAsync(styles);
+  const updateStyle = async (styles: Partial<Style>, page?: ContributionPage) => {
```
nrh-cklimas, line 99: Same note about optionality of `page`.

spa/src/hooks/useContributionPage/useContributionPage.ts
```diff
@@ -158,12 +158,22 @@ export function useContributionPage(revenueProgramSlugOrPageId?: number | string
       }
 
       let styles: Style | undefined;
+      // Always pass in the updated page name
+      const latestPageName = { ...page, name: data?.name || page.name };
+
+      if (page.styles?.id && !data.styles && latestPageName.name !== page.styles?.name) {
```
nrh-cklimas, line 164: It looks like this is potentially doing two `updateStyle` calls. Can we combine with the one on line 173?

GuiMend: It will never run both updates as one if required `data.styles` to be **true** and the other requires it to be **false**.

nrh-cklimas: I see now. How about we make that explicit?

```tsx
// Always pass in the updated page name
const latestPageName = { ...page, name: data.name ?? page.name };

if (data.styles) {
  // Page style has changed. Either create a new one or update the existing.

  const mutator = data.styles?.id ? updateStyle : createStyle;
  const { data: response } = await updateStyle(data.styles, latestPageName);

  styles = response;
} else if (page.styles?.id && latestPageName.name !== page.styles?.name) {
  // Page style hasn't changed, but the page name has. We need to update the
  // style name to match it.

  const { data: response } = await updateStyle(page.styles, latestPageName);

  styles = response;
}
```

nrh-cklimas (approved):

PR #1341, opened 2023-12-12
DEV-3886: Add detail view to contribution list

nrh-cklimas:

n/a

#### What's this PR do?

1. Implements a read-only view of contributions in the new portal.
2. Adds trailing slashes to routes that weren't missed in previous work.
3. Adds `status` to the contribution detail endpoint mock.
 
Continues implementation of the new portal.

Yes.

Documentation around the new portal.

Yes.

n/a

I tried a couple different approaches before settling on `useDetailAnchor` to position the detail view. I'm still not fully convinced it's the best way.

- Can't use [Popper component](https://v4.mui.com/api/popper/) because a) it's tricky to manage because we only want the positioning in desktop viewports b) we want the detail view to be in the main page layout so that the page height is correct. e.g. if it were positioned absolutely, the footer would get potentially obscured.
- Can't use pure CSS grid because the height of the detail view may not align exactly with the grid.
- Can't float the detail view right because we may want it to be vertically above the item it's next to.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3886

No.

See rest of epic.


GuiMend (commented):
@nrh-cklimas we have a really similar arrow icon to spa/src/components/portal/ContributionsList/ContributionItem/selected-arrow.svg in the icon library:
> @material-design-icons/svg/round/arrow_left.svg
image

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <Amount $status={contribution.status} data-testid="amount">
+          {formatCurrencyAmount(contribution.amount, true)}
+        </Amount>
+      </Root>
+    );
+  }
+);
+
+// FIXME
+(ContributionItem as any).propTypes = ContributionItemPropTypes;
```
GuiMend, line 99: Was this forgotten @nrh-cklimas ?

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.styled.ts
```diff
@@ -1,29 +1,69 @@
...
-    grid-template-columns: [icon] 35px [date] 1fr [amount] minmax(100px, 1fr);
-    grid-template:
-      'status status status'
-      'icon date amount';
-    padding: 12px;
+/**
+ * Before the mobile breakpoint, we need to hide one column to maintain the
+ * layout. This is the breakpoint where this occurs.
+ */
+const COMPRESS_COLUMNS_BREAKPOINT = '1200px';
```
GuiMend, line 10: Shouldn't we add this to our theme?

spa/src/components/portal/ContributionsList/ContributionDetail/useDetailAnchor.ts
```diff
@@ -0,0 +1,47 @@
...
+ * - Handle resizing between mobile and larger viewports.
+ * - Keep the detail inside its parent container, both top and bottom.
+ * - To the extent possible, vertically align detail's top with the item's top.
+ * - Maintain the detail in layout (i.e. not position it absolutely) so that it
+ *   affects the total page height.
+ *
+ * In practice, the vertical position of the anchor never varies once the page
+ * is rendered, so we don't need to listen for viewport resizes.
+ */
+export function useDetailAnchor(itemEl: OptionalHTMLElement, detailEl: OptionalHTMLElement) {
```
GuiMend, line 19: Crazy hook 😅 . Good job making this work

spa/src/components/portal/ContributionsList/ContributionDetail/useDetailAnchor.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  describe('When both parent and item element are defined', () => {
+    it('shows the detail element', () => {
+      renderHook(() => useDetailAnchor(item, detail));
+      expect(detail.style.visibility).toBe('visible');
+    });
+
+    // Testing of positioning doesn't appear possible in jsdom. But if we could,
+    // we'd want to test that:
+    // - The detail is moved downward to match the item's offset if the parent permits
+    // - The detail is moved to the bottom of the parent if it can't be moved exactly in place
```
GuiMend, line 38: Can we test for the existence of `--two-column-vertical-offset` prop and if the value is what we expect?

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.tsx
nrh-cklimas: Ahh I did forget to resolve this. Maybe you have an idea on this... without the any cast, it says:

```
Type '{ contribution: PropTypes.Validator<object>; replaceHistory: PropTypes.Requireable<boolean>; selected: PropTypes.Requireable<boolean>; }' is not assignable to type 'WeakValidationMap<ContributionItemProps & RefAttributes<HTMLAnchorElement>>'.
  Types of property 'contribution' are incompatible.
    Type 'Validator<object>' is not assignable to type 'Validator<PortalContribution>'.
      Type '{}' is missing the following properties from type 'PortalContribution': amount, card_brand, created, credit_card_expiration_date, and 11 more.ts(2322)
```
I feel like I'm making a basic mistake here but I'm not seeing it. A fix is to remove `contribution` from the prop types (but not the TypeScript) but that doesn't seem ideal. Have you seen something like this before?

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.styled.ts
nrh-cklimas: Interesting point. I'm not sure what we'd call it... `largeishLaptop`? :) It feels situational to me so I hesitate to put it into the theme, but it's not a strong opinion on my part.

spa/src/components/portal/ContributionsList/ContributionDetail/useDetailAnchor.test.ts
nrh-cklimas: The issue I hit in testing was that in jsdom, I could never get the `offsetTop`, `offsetHeight`, etc values for test elements to be anything but 0. I tried assigning a height in inline style IIRC and also deleting the property directly, but jsdom flagged them as readonly properties (which makes sense in general but blocks us here). But yeah, if we can figure out a way around this issue, we should be able to test on the resulting CSS variable.

nrh-cklimas:
> @nrh-cklimas we have a really similar arrow icon to spa/src/components/portal/ContributionsList/ContributionItem/selected-arrow.svg in the icon library:
> 
> > @material-design-icons/svg/round/arrow_left.svg

Plugged this in with Rachel's feedback in 2c79da395a2f05f50f6de70343dbb6a5255c1bf3

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.styled.ts
GuiMend: I'm ok leaving it out of the theme for now and waiting to add it until we use it more frequently.

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.tsx
GuiMend: One approach I used to fix this is by omitting the PropType from the TS:
```
export interface ContributionItemProps extends Omit<InferProps<typeof ContributionItemPropTypes>, "contribution"> {
  contribution: PortalContribution;
}
```

I am not sure the syntax is correct in my suggestion, but it ignores the previous definition for contribution, and you can override it.

nrh-cklimas: Weirdly, this doesn't work either...

```
  Types of property 'contribution' are incompatible.
    Type 'Validator<object>' is not assignable to type 'Validator<PortalContribution>'.
      Type '{}' is missing the following properties from type 'PortalContribution': amount, card_brand, created, credit_card_expiration_date, and 11 more.ts(2322)
```

Anyway, I took `contribution` out of the prop types entirely in b726f535cde54221af2a879dad8275c452b1187e. It's not an ideal solution but it feels like the best we can do.

GuiMend: [suggestion] We can leave it the way you did in the last commit, but how about we try adding the prop back with `PropTypes.any.isRequired` ?

GuiMend (approved):
Left one last suggestion. If it's decided not to do it, you can resolve the comment 👍

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.tsx
nrh-cklimas: Sounds good! Added in b340e37b2f108d90060845db76fe75c301cd7696

PR #620, opened 2022-08-26
Dev 2205 deactivate contributions for early adopters

harsha09:
#### What's this PR do?
Limit access to Contributions dashboard for selected users by allowing access to everyone.

#### Why are we doing this? How does it help us?
Need a mechanism to disable Contributions dashboard access for few Orgs / RP users but should be accessible to all other users.

#### How should this be manually tested? Please include detailed step-by-step instructions.
To test this feature a waffle flag `spa-contributions-section-deny` need to be created using Django Admin.

How to create Waffle flag? Login to Django Admin portal -> Flags Section -> Add
Flag name should be `spa-contributions-section-deny`
1. Can enable this flag to everyone (No user should be able to access contributions dashboard).
2. Can also enable this flag by specifying users where the specified users will not be able to access the contributions dashboard.
3. `spa-contributions-section-deny` will take precedence over `spa-contributions-section-access`, you can play around with the combinations.

After creating the Waffle Flag login to Contributions with the user for whom `spa-contributions-section-deny` is enabled, Contributions tab should not be visible and navigation to /contributions redirects to 404 page.

Users for whom `spa-contributions-section-deny` is disabled and `spa-contributions-section-access` is enabled Contributions tab should be visible and be able to list contributions.

#### Did this PR make changes to the user interface that may require changes to the user-facing docs?
No changes to UI but enabling and disabling Contributions for few set of users.

#### Have automated unit tests been added? If not, why?
Yes

#### How should this change be communicated to end users?
Probably user manual.

#### Are there any smells or added technical debt to note?
NA

#### Has this been documented? If so, where?
Not yet

#### What are the relevant tickets? Add a link to any relevant ones.
[DEV-2205]

#### Do any changes need to be made before deployment to production (adding environment variables, for example)?
NA

#### Are there next steps? If so, what? Have tickets been opened for them? List next-step tickets here:
NA

[DEV-2205]: https://news-revenue-hub.atlassian.net/browse/DEV-2205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1346, opened 2023-12-15
DEV-4032 | implement new contributor facing contributions api endpoints

benjaminEwhite:

Initially, this PR has DEV-4045-listen-for-webhooks-to-create-update-payment-objects as its base because that has not yet been merged into main and in some places it was better for code in this branch to presuppose those changes.

After DEV-4045 has been merged, base should be set to `main`.

#### What's this PR do?

This PR adds API support for the new contributor portal. It does that by creating a new viewset and new serializers that are specifically for use by contributors to interact with their own contributions.

Overall, the API is meant to provide parity with API used by status quo portal. That means allow reads of list and detail, and ability to update payment method and/or cancel recurring subscriptions.

It solves for the AC in the following 4 tickets:

- [DEV-4032](https://news-revenue-hub.atlassian.net/browse/DEV-4032)
- [DEV-3885](https://news-revenue-hub.atlassian.net/browse/DEV-3885)
- [DEV-3887](https://news-revenue-hub.atlassian.net/browse/DEV-3887)
- [DEV-3888](https://news-revenue-hub.atlassian.net/browse/DEV-3888)

This PR adds:

- A read-only list view at GET `api/v1/contributors/<contributor_id>/contributions`. This endpoint conforms [to the API schema captured in this new schema file](https://github.com/newsrevenuehub/rev-engine/blob/DEV-4032-implement-new-contributor-facing-contributions-api-endpoint/apps/api/schemas/portal-contribution-list.openapi.yml) that we'll track in source code from now on.
- A read, (patch) update, and delete endpoint at `api/v1/contributors/<contributor_id>/contributions/<contribution_id>` conforming to [API spec tracked here](https://github.com/newsrevenuehub/rev-engine/blob/DEV-4032-implement-new-contributor-facing-contributions-api-endpoint/apps/api/schemas/portal-contribution-detail.openapi.yml).
- We also add a new column to the payment table: `transaction_time`. This became necessary in order to support ordering payments in contribution serializer. `created` is not a reliable value because when the entity gets created in our db is somewhat arbitrary, whereas Stripe's balance transaction created time is the definitive time at which the transaction occurred.

Required to ship new portal.

Yes. Each of the 4 associated tickets have detailed testing instructions.

No.

Yes.

N/A

We might want to reconsider how we're handling payment method update if feasible. I don't love that we allow client to send provider_payment_method_id to backend. Would prefer that SPA initiates action with stripe which causes webhook we listen for that causes necessary state changes on backend. See follow up ticket [DEV-4342](https://news-revenue-hub.atlassian.net/browse/DEV-4342).

- [DEV-4032](https://news-revenue-hub.atlassian.net/browse/DEV-4032)
- [DEV-3885](https://news-revenue-hub.atlassian.net/browse/DEV-3885)
- [DEV-3887](https://news-revenue-hub.atlassian.net/browse/DEV-3887)
- [DEV-3888](https://news-revenue-hub.atlassian.net/browse/DEV-3888)
- [DEV-4387](https://news-revenue-hub.atlassian.net/browse/DEV-4387)
- [ITS-2211](https://news-revenue-hub.atlassian.net/browse/ITS-2211)
- [DEV-2287](https://news-revenue-hub.atlassian.net/browse/DEV-2287)
- [DEV-4391](https://news-revenue-hub.atlassian.net/browse/DEV-4391)
- [DEV-4403](https://news-revenue-hub.atlassian.net/browse/DEV-4403)

- [DEV-4343](https://news-revenue-hub.atlassian.net/browse/DEV-4343)
- [DEV-4342](https://news-revenue-hub.atlassian.net/browse/DEV-4342)
- [DEV-4333](https://news-revenue-hub.atlassian.net/browse/DEV-4333)
- [ITS-2211](https://news-revenue-hub.atlassian.net/browse/ITS-2211)
- [DEV-2287](https://news-revenue-hub.atlassian.net/browse/DEV-2287)
- [DEV-4403](https://news-revenue-hub.atlassian.net/browse/DEV-4403)

[DEV-4032]: https://news-revenue-hub.atlassian.net/browse/DEV-4032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DEV-4342]: https://news-revenue-hub.atlassian.net/browse/DEV-4342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DEV-3885]: https://news-revenue-hub.atlassian.net/browse/DEV-3885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/models.py
```diff
@@ -223,6 +224,15 @@ class Meta:
...
         return f"{self.formatted_amount}, {self.created.strftime('%Y-%m-%d %H:%M:%S')}"
 
+    @property
+    def next_payment_date(self) -> datetime.datetime | None:
+        if self.interval == ContributionInterval.ONE_TIME:
+            return None
+        if not self.stripe_subscription:
+            logger.warning("Expected a retrievable stripe subscription on contribution %s but none was found", self.id)
+            return None
+        return datetime.datetime.fromtimestamp(self.stripe_subscription.current_period_end, tz=pytz.UTC)
```
benjaminEwhite, line 234: **Question**

What will this be in case of canceled subscription? Is there an edge case to handle there?

apps/contributions/models.py
```diff
@@ -609,7 +619,74 @@ def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
...
+    @property
+    def card_last_4(self) -> str | None:
+        return self.card.last4 if self.card else None
+
+    @property
+    def is_cancelable(self) -> bool:
+        return self.status in (ContributionStatus.PROCESSING, ContributionStatus.FLAGGED)
+
+    @property
+    def is_modifiable(self) -> bool:
```
benjaminEwhite, line 658: **Question** 

Is this the right way to determine. It turns out the status quo way of determining is_modifiable seems to reference non-existent statuses (as well as one existent one).  See https://github.com/newsrevenuehub/rev-engine/blob/1d114f81e3acbb0b971cd58adddf84b8e89d0d14/apps/contributions/serializers.py#L801

Best i can tell, we would want to allow modificaiton of recurring if it's status is either paid (implying first payment has been processed and there's potential for future ones) or if status is failed, in which case the user couldl update their CC info to make work. 

Does this seem right?

apps/contributions/views.py
```diff
@@ -565,60 +564,146 @@ def _re_retrieve_pi_and_insert_pi_and_sub_into_cache(
...
+        try:
+            subscription = stripe.Subscription.retrieve(
+                serializer.instance.provider_subscription_id,
+                stripe_account=serializer.instance.revenue_program.payment_provider.stripe_account_id,
+                expand=["customer"],
+            )
+        except stripe.error.StripeError:
+            logger.exception("stripe.Subscription.retrieve returned a StripeError")
+            return Response({"detail": "subscription not found"}, status=status.HTTP_404_NOT_FOUND)
+        if (email := request.user.email.lower()) != subscription.customer.email.lower():
```
benjaminEwhite, line 658: **Question**

Can this be removed? I copied this business logic from the status quo Stripe-cache based approach to showing users their contributions.  Maybe this is not a concern here? And if it is, probably better handled more explicitly  and in an easily reproducible way (perhaps with a decorator).

njharman: Seems to be covered by  `contribution = contributor.contribution_set.get(pk=contribution_id)` in contribution_detail(), right?  Or can user with valid contribution, twiddle the http payload/url to set provider_subscription_id to unrelated one one they don't own. If that's not non-sensical question/case should be unit test for it.

I worry a little that that contribution_set() "check" is outside these handle_ methods (some future dev will call them without similar check).  Maybe rename handle_* to _handle_* to give hint these are internal/depend on internal implementation/use with caution.
 
Seems issue prone and "sad" (because it can change, is out of our control) to verify user by email. I remember complaining before and it has to be this way cause we don't have User records for contributors, just whatever is passed via magic link which is email based, so email **is** the best id.

If this logic is kept, wouldn't it be better to pass the contributor (retrieved line #620) and using that instead of pulling out of request.  Cut request and pass contributor and provider_payment_id

apps/contributions/views.py
```diff
@@ -568,60 +567,146 @@ def _re_retrieve_pi_and_insert_pi_and_sub_into_cache(
...
+
+    def handle_delete(self, contribution):
+        """If subscription found, cancel it in Stripe.
+
+        NB: we don't do anything to update NRE contribution status here and instead rely on ensuing webhooks to do that.
+        """
+        if contribution.interval == ContributionInterval.ONE_TIME:
+            logger.warning("Request was made to cancel one-time contribution %s", contribution.id)
+            return Response({"detail": "Cannot cancel one-time contribution"}, status=status.HTTP_400_BAD_REQUEST)
+        try:
```
njharman, line 700: Shouldn't we be checking it is in cancellable state?  maybe via is_cancelable()

apps/contributions/views.py
```diff
@@ -568,60 +567,146 @@ def _re_retrieve_pi_and_insert_pi_and_sub_into_cache(
...
-                    ]
-                    return Response(result, status=status.HTTP_200_OK)
-    return Response({"detail": f"No contribution exists with ID {contribution_id}"}, status=status.HTTP_404_NOT_FOUND)
+        match request.method:
+            case "GET":
+                pass
+            case "PATCH":
+                serializer.is_valid(raise_exception=True)
+                return self.handle_patch(serializer=serializer, request=request)
+            case "DELETE":
```
njharman, line 636: Coverage says DELETE never matched. I'm not sure I believe that.

njharman (changes requested):

apps/api/permissions.py
```diff
@@ -88,6 +88,14 @@ def has_permission(self, request, view):
         return request.method == "PATCH"
 
 
+class UserIsContributor(permissions.BasePermission):
```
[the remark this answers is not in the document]
benjaminEwhite: 709250bfebe3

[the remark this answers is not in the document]
benjaminEwhite: 8d5a1b2 and 082ca1a8c

apps/api/permissions.py
```diff
@@ -88,6 +88,14 @@ def has_permission(self, request, view):
...
+    def has_object_permission(self, request, view, obj):
+        has_permission = obj == request.user and isinstance(obj, Contributor)
```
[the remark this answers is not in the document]
benjaminEwhite: Good catch!  8d5a1b2c

apps/api/permissions.py
```diff
@@ -88,6 +88,14 @@ def has_permission(self, request, view):
...
+        logger.info("Temp log showing object permission in UserIsContributor: %s", has_permission)
```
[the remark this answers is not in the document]
benjaminEwhite: 709250bf

apps/contributions/models.py
```diff
@@ -14,6 +14,7 @@
 from django.template.loader import render_to_string
 from django.utils.safestring import SafeString, mark_safe
 
+import pytz
```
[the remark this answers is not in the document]
benjaminEwhite: Good to know:

ffaa2f0218

apps/contributions/models.py
```diff
@@ -607,7 +620,70 @@ def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
...
+            expand=["default_source"],
+        )
+        return customer.default_source if customer.default_source and customer.default_source.object == "card" else None
+
+    @property
+    def card_brand(self) -> str:
+        return self.card.brand if self.card else ""
+
+    @property
+    def card_expiration_date(self) -> str:
```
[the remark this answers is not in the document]
benjaminEwhite: Ultimately, I'm liking having "neat" serializers that have the client-facing names we want (no need to have `_str`  ending for the client) and don't require additional configuration (that is to say, we just name the fields we want in `PORTAL_CONTRIBUTION_BASE_SERIALIZER_FIELDS` and don't have to explicitly declare those fields with `foo = serializers. SerializerMethodField(source="foo_str")`

apps/contributions/models.py
```diff
@@ -966,6 +1042,8 @@ class Payment(IndexedTimeStampedModel):
     gross_amount_paid = models.IntegerField()
     amount_refunded = models.IntegerField()
     stripe_balance_transaction_id = models.CharField(max_length=255, unique=True)
+    # add index on transaction time since wwill need to sort by for last pasyment
```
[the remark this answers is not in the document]
benjaminEwhite: Also turned this into a JIRA todo

c3f889d8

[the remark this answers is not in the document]
benjaminEwhite: Turns out that also requires a data migration so we can manually retrieve the balance_transaction for the Payment (which is source for payment.transaction_time).  Would prefer to do that in a follow up ticket, as that adds scope here in short term, and would like to move this through.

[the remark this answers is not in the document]
benjaminEwhite: also turned this into a JIRA todo

c3f889d8

benjaminEwhite:
Closing and reopening because migrations changed and need to build from scratch.

apps/contributions/models.py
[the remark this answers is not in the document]
benjaminEwhite: Ended up adding now, which requires allowing transaction_time to be nullable in short term, and an additional data migration to backfill. Once this PR is merged, we can make non-nullable in DEV-4379 (which is. TODO).

apps/contributions/serializers.py
```diff
@@ -830,27 +831,83 @@ def get_payment_type(self, instance):
...
+
+
+class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
+    payments = PortalContributionPaymentSerializer(many=True, read_only=True, source="payment_set")
+
+    class Meta:
+        model = Contribution
+        fields = PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS
+        read_only_fields = [
+            K for K in PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS if K != "provider_payment_method_id"
```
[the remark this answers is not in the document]
benjaminEwhite: Another good catch. This is a vestige of earlier iteration of schema. No longer needed. Removed in f69b1d2a4c

apps/contributions/views.py
benjaminEwhite: @njharman yeah i'm not sure what's going on here. we definitely test this path (you can put in a debugger and run `test_contribution_detail_delete_happy_path`, for instance).

apps/contributions/views.py
```diff
@@ -568,60 +567,146 @@ def _re_retrieve_pi_and_insert_pi_and_sub_into_cache(
...
+                return self.handle_delete(contribution)
+
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
benjaminEwhite: See 4397e94a9

apps/contributions/models.py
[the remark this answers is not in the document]
benjaminEwhite: This is outdated

apps/contributions/models.py
benjaminEwhite: 9a452fc536

apps/contributions/views.py
benjaminEwhite: Good catch. I did this, and also removed the check on contribution interval, since the critical thing here is the status of the stripe subscription (if any).   Did the same in handle_patch (but with is_modifiable).

9a452fc536

apps/contributions/views.py
benjaminEwhite: @njharman implementation you suggest makes sense, but want to hear from @x110dc if he thinks this check can/should be removed altogether.

njharman (approved):

apps/contributions/migrations/0009_DEV-4032_transaction_time_data_migration.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def set_transaction_time(apps, schema_editor):
+    Payment = apps.get_model("contributions", "Payment")
+    for payment in Payment.objects.filter(transaction_time__isnull=True):
+        bt = stripe.BalanceTransaction.retrieve(
+            payment.stripe_balance_transaction_id, stripe_account=payment.contribution.stripe_account_id
+        )
+        payment.transaction_time = bt.created
+        payment.save(update_fields={"modified", "transaction_time"})
```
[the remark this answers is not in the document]
benjaminEwhite: Thinking about this more, I wonder if this is an argument to not do this as a data migration, in favor of porting that set_transaction_time function into a management command or one off script. Going to leave this conversation thread open to get @x110dc 's thoughts on best way to handle.

apps/api/schemas/portal-contribution-detail.openapi.yml
x110dc, line 1: How are these schemas used? How with they be kept in sync with the code?

apps/contributions/migrations/0009_DEV-4032_transaction_time_data_migration.py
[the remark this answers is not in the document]
x110dc: Yes, I think it should be a management command and not part of the migration. Since there's no batch way to do this it could take quite some time and that will halt/slow deployment.

apps/contributions/models.py
```diff
@@ -607,7 +633,85 @@ def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
...
+    def card(self) -> stripe.Card | None:
+        if not (cust_id := self.provider_customer_id):
+            return None
+        customer = stripe.Customer.retrieve(
+            cust_id,
+            stripe_account=self.donation_page.revenue_program.payment_provider.stripe_account_id,
+            expand=["default_source"],
+        )
+        return customer.default_source if customer.default_source and customer.default_source.object == "card" else None
+
```
x110dc, line 646: The card could also be on `customer.invoice_settings.default_payment_method` -- it may not for RevEngine but for other sources (migrated to RevEngine, for example) it will be.

apps/contributions/models.py
```diff
@@ -607,7 +633,85 @@ def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
...
+    def is_modifiable(self) -> bool:
+        pi = self._expanded_pi_for_cancelable_modifiable
+        return (
+            pi.invoice.subscription.status in self.MODIFIABLE_SUBSCRIPTION_STATUSES
+            if pi and pi.invoice and pi.invoice.subscription
+            else False
+        )
+
+    @property
+    # TODO: [DEV-4333] Update this to be ._last_payment_date when no longer in conflict with db model field
```
x110dc, line 690: remove underscore in comment?

apps/contributions/models.py
```diff
@@ -1109,6 +1215,9 @@ def _handle_create_payment(
                 "net_amount_paid": balance_transaction.net,
                 "amount_refunded": amount_refunded,
                 "gross_amount_paid": balance_transaction.amount,
+                "transaction_time": datetime.datetime.fromtimestamp(
+                    balance_transaction.created, tz=datetime.timezone.utc
```
x110dc, line 1219: Why use the date from the balance transaction and not the payment intent or invoice? The datetimes should be close, but it's not technically the same thing.

apps/contributions/views.py
x110dc: I don't recall why this logic is there. Can we look at the previous PR/commit that added it to see if there's a clue there? If that doesn't help answer the question I'm okay with removing it.

x110dc (changes requested):
A few questions/comments.

apps/contributions/models.py
benjaminEwhite: I'd prefer to have a single stable referent here and not have conditionality around where transaction time comes from. Whether it's a one-time charge, a recurrence, or a refund, we'll always have a balance transaction.  

We could rename this field to `balance_transaction_created` or something that makes source explicit if that's important, or add a docstring if that suffices.

apps/contributions/models.py
benjaminEwhite: 69e16efaab

apps/api/schemas/portal-contribution-detail.openapi.yml
benjaminEwhite: I considered writing a test, but I think that's maybe overkill. I think I'll remove from source, as this was initially for coordinating simultaneous SPA and server-side work around not-yet-created interfaces.

In principle, I like the idea of committing schema to source code and then having tests that prove that API implements the schema. I created a followup task around that here: https://news-revenue-hub.atlassian.net/browse/DEV-4387

apps/contributions/migrations/0009_DEV-4032_transaction_time_data_migration.py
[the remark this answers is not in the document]
benjaminEwhite: 7217d0de

apps/contributions/models.py
benjaminEwhite: 4588933102

apps/contributions/views.py
benjaminEwhite: I looked back a few commits and didn't see any obvious reason we were doing that. 

Best I can come up with is that it may have been defensive programming re: how we're storing data in cache (which keys off of email).  

I've removed this code inclusive of the todo comment (for which i've added a note to ticket and PR description).

9501630bc7

apps/contributions/management/commands/sync_payment_transaction_time.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Command(BaseCommand):
+    """Try to populate null values for payment.transaction_time in db."""
+
+    def handle(self, *args, **options):
+        self.stdout.write(self.style.HTTP_INFO("Running `sync_payment_transaction_time`"))
+        queryset = Payment.objects.filter(transaction_time__isnull=True)
+        self.stdout.write(self.style.HTTP_INFO(f"Found {queryset.count()} payments to sync"))
+        for payment in queryset.all():
+            bt = stripe.BalanceTransaction.retrieve(
```
x110dc, line 18: It can be addressed in another ticket/PR but I think we'll find that some of these will fail. For example, for accounts that were connected but no longer are. So, we'd want a way to skip those and continue.

apps/contributions/models.py
x110dc: Either remedy is okay by me.

x110dc (approved):
Approved.

apps/contributions/management/commands/sync_payment_transaction_time.py
benjaminEwhite: d4f65817

apps/contributions/models.py
benjaminEwhite: a84413b

apps/contributions/models.py
```diff
@@ -633,33 +633,103 @@ def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
             )
             return None
 
-    @cached_property
-    def card(self) -> stripe.Card | None:
```
benjaminEwhite: @x110dc wanted to get your feedback on this. In QA, Agustin found that card was not populated. I thought it would work with original implementation here, but ended up needing to mimic implementation in status quo stripe/cache based approach (where there's attempt to get card from pi).  

Wondering if you have any insight into why customer approach didn't work out.

PR #1359, opened 2024-01-04
DEV-4220: Add cancel and payment method update notification emails

nrh-cklimas:

n/a

#### What's this PR do?

Adds two new emails: when a recurring contribution's payment method is changed and when it is cancelled. These are triggered by Stripe webhook events so that if these changes happen externally, the contributor still gets the email.

Changes recurring email reminders to use a shared method for sending contribution-related emails.

Increases user confidence that changes went through.

Simplifies code around emails.

Yes.

Might want to mention these new emails.

Yes.

n/a

I'm not super happy with this code to be honest. It was a lot of copy and paste work from other existing emails. I'm very open to suggestions on how we can refactor this to avoid code duplication.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-4220

No.

No.


apps/contributions/models.py
```diff
@@ -528,6 +528,118 @@ def handle_thank_you_email(self):
...
+            logger.exception(
+                "Something went wrong retrieving Stripe customer for contribution with ID %s",
+                self.id,
+            )
+            return
+
+        token = str(ContributorRefreshToken.for_contributor(self.contributor.uuid).short_lived_access_token)
+        data = {
+            # Wedging the date into the template--it's now, not the date of the contribution.
+            "contribution_date": datetime.datetime.today().strftime("%m/%d/%Y"),
```
benjaminEwhite, line 561: I don't love that we're overloading `contribution_date` presumably to get some common formatting.

Would prefer to see var name be more generic, or possibly use two separate var names.

apps/contributions/models.py
```diff
@@ -528,6 +528,118 @@ def handle_thank_you_email(self):
...
+            "contribution_amount": self.formatted_amount,
+            "contribution_interval_display_value": self.interval,
+            "contributor_name": customer.name,
+            "non_profit": self.donation_page.revenue_program.non_profit,
+            "contributor_email": self.contributor.email,
+            "tax_id": self.donation_page.revenue_program.tax_id,
+            "fiscal_status": self.donation_page.revenue_program.fiscal_status,
+            "fiscal_sponsor_name": self.donation_page.revenue_program.fiscal_sponsor_name,
+            "magic_link": mark_safe(
+                f"https://{construct_rp_domain(self.donation_page.revenue_program.slug)}/{settings.CONTRIBUTOR_VERIFY_URL}"
```
benjaminEwhite, line 571: Should be able to use existing `Contributor.create_magic_link` static method for this.  https://github.com/newsrevenuehub/rev-engine/blob/main/apps/contributions/models.py#L104

apps/contributions/models.py
```diff
@@ -528,6 +528,118 @@ def handle_thank_you_email(self):
...
+            self.contributor.email,
+            "Canceled contribution",
+            render_to_string("recurring-contribution-canceled.txt", data),
+            render_to_string("recurring-contribution-canceled.html", data),
+        )
+
+    def send_recurring_contribution_payment_updated_email(self) -> None:
+        from apps.api.views import construct_rp_domain
+        from apps.emails.tasks import send_templated_email
+
```
benjaminEwhite, line 593: Ditto re imports

apps/contributions/models.py
```diff
@@ -528,6 +528,118 @@ def handle_thank_you_email(self):
...
+            "contribution_date": datetime.datetime.today().strftime("%m/%d/%Y"),
+            "contribution_amount": self.formatted_amount,
+            "contribution_interval_display_value": self.interval,
+            "non_profit": self.donation_page.revenue_program.non_profit,
+            "contributor_email": self.contributor.email,
+            "contributor_name": customer.name,
+            "tax_id": self.donation_page.revenue_program.tax_id,
+            "fiscal_status": self.donation_page.revenue_program.fiscal_status,
+            "fiscal_sponsor_name": self.donation_page.revenue_program.fiscal_sponsor_name,
+            "magic_link": mark_safe(
```
benjaminEwhite, line 629: Ditto re: using static method

apps/contributions/models.py
```diff
@@ -528,6 +528,118 @@ def handle_thank_you_email(self):
...
+            )
+        except StripeError:
+            logger.exception(
+                "Something went wrong retrieving Stripe customer for contribution with ID %s",
+                self.id,
+            )
+            return
+
+        token = str(ContributorRefreshToken.for_contributor(self.contributor.uuid).short_lived_access_token)
+        data = {
```
benjaminEwhite, line 559: Looks like this is largely overlapping with data creation in send_recurring_contribution_payment_updated_email so probably could dry up.

apps/contributions/tests/test_models.py
```diff
@@ -1143,6 +1143,330 @@ def __init__(self, *args, **kwargs):
...
+            ContributorRefreshToken, "for_contributor", lambda *args, **kwargs: MockForContributorReturn()
+        )
+        customer_name = "Fake Customer Name"
+        mocker.patch("stripe.Customer.retrieve", return_value=AttrDict({"name": customer_name}))
+        settings.EMAIL_BACKEND = "django.core.mail.backends.locmem.EmailBackend"
+        settings.CELERY_ALWAYS_EAGER = True
+        contribution.send_recurring_contribution_payment_updated_email()
+        magic_link = mark_safe(
+            f"https://{construct_rp_domain(contribution.donation_page.revenue_program.slug)}/{settings.CONTRIBUTOR_VERIFY_URL}"
+            f"?token={token}&email={quote_plus(contribution.contributor.email)}"
```
benjaminEwhite, line 1183: Is there any reason we need to construct a realistic URL in this test, or could this just as simply be an empty string or some dummy value?

apps/contributions/tests/test_models.py
```diff
@@ -1143,6 +1143,330 @@ def __init__(self, *args, **kwargs):
...
+                    "This receipt may be used for tax purposes.",
+                    f"{contribution.donation_page.revenue_program.name} is a 501(c)(3) nonprofit organization",
+                    f"with a Federal Tax ID #{tax_id}" if tax_id else "",
+                ]
+            )
+        else:
+            email_expectations.append(
+                f"Contributions to {contribution.donation_page.revenue_program.name} are not deductible as charitable donations."
+            )
+        for x in email_expectations:
```
benjaminEwhite, line 1215: This is totally optional, but I like using match/case for situation like this where conditional is on same condition for each branch.

```
match revenue_program.fiscal_status:

    case FiscalStatusChoices.FISCALLY_SPONSORED:
        // do stuff

    case FiscalStatusChoices.NONPROFIT:
        // do other stuff

    case _:
        // do default stuff

File: apps/contributions/tests/test_models.py [hunk 8]
```diff
@@ -1143,6 +1143,330 @@ def __init__(self, *args, **kwargs):
...
+        "has_default_donation_page",
+        (False, True),
+    )
+    def test_send_recurring_contribution_payment_updated_email_styles(
+        self, revenue_program, has_default_donation_page, monkeypatch, mocker, settings
+    ):
+        with patch("apps.contributions.models.Contribution.fetch_stripe_payment_method", return_value=None):
+            contribution = ContributionFactory(
+                interval=ContributionInterval.YEARLY, provider_customer_id="test-customer-id"
+            )
```
benjaminEwhite, line 1241:
```suggestion
    def test_send_recurring_contribution_payment_updated_email_styles(
        self, revenue_program, has_default_donation_page, monkeypatch, mocker, settings
    ):
       mocker.patch("apps.contributions.models.Contribution.fetch_stripe_payment_method", return_value=None)
        contribution = ContributionFactory(
            interval=ContributionInterval.YEARLY, provider_customer_id="test-customer-id"
        )
```

benjaminEwhite: Just going to note this here, but ^^ would be preferred style throughout.

apps/emails/views.py
```diff
@@ -30,11 +34,80 @@ def preview_recurring_contribution_reminder(request):
...
+        "contribution_date": datetime.now(),
+        "contribution_amount": "$123.45",
+        "contributor_name": "Contributor Name",
+        "copyright_year": datetime.now().year,
+        "contributor_email": "nobody@fundjournalism.org",
+        "magic_link": "https://magic-link",
+        "rp_name": rp.name,
+        "style": style,
+    }
+    if rp.default_donation_page:
```
benjaminEwhite, line 59: Looks like this path doesn't get tested (according to coverage report).  Should test this path (which will get to 100%).

apps/emails/views.py
```diff
@@ -8,10 +8,14 @@
 
 
 def preview_recurring_contribution_reminder(request):
+    """
+    Previews a recurring contribution reminder in the browser. Only accessible
+    if DEBUG is set (see revengine/urls.py).
```
benjaminEwhite, line 13: This is probably a good guard against leaking sensitive data, but otherwise, this endpoint does not have access control. Could it be limited to superusers or hub admins (look for usage of `@permission_classes` decorator in existing code).

This is true for other view as well.

apps/emails/views.py
```diff
@@ -8,10 +8,14 @@
 
 
 def preview_recurring_contribution_reminder(request):
```
benjaminEwhite, line 10: Think it might be better to have a single view function that takes the name of an email template as one of its params (in addition to current ones).  With current design, we have to add a distinct URL to urls.py and distinct view here for any additional emails. 

Below is not totally worked out, but gives a single view function we can use for all email preview needs.

```python
from dataclasses import asdict

from django.http import HttpResponse
from django.template.loader import get_template
from django.template import TemplateDoesNotExist

PERMITTED_TEMPLATES = (
    "recurring-contribution-email-reminder.html",
    "recurring-contribution-canceled.html",
    "recurring-contribution-email-reminder.txt",
    "recurring-contribution-canceled.txt",
    "nrh-default-contribution-confirmation-email.html"
    "nrh-default-contribution-confirmation-email.txt"
)

def preview_email_template(request):
    template_name = request.GET.get('email_template', None)
    if template_name not in PERMITTED_TEMPLATES:
        return HttpResponse("Template not permitted.", status=400)
    
    rp_id = request.GET.get('rp_id', None)
    if rp_id is None:
        return HttpResponse("`rp_id` is a required field.", status=400)
    try:
        rp = RevenueProgram.objects.get(id=rp_id)
        return HttpResponse("RevenueProgram does not exist.", status=404)
    style = (rp_style:=asdict(rp.transactional_email_style) | {"logo_url": (url:=request.GET.get('logo_url', rp_style["logo_url"]))})
    
    try:
        template = get_template(template_name)
    except TemplateDoesNotExist:
        return HttpResponse("Template does not exist.", status=404)

    data = {
        "contribution_interval_display_value": "year",
        "contribution_date": datetime.now(),
        "contribution_amount": "$123.45",
        "contributor_name": "Contributor Name",
        "copyright_year": datetime.now().year,
        "contributor_email": "nobody@fundjournalism.org",
        "magic_link": "https://magic-link",
        "rp_name": rp.name,
        "style": style,
        "default_contribution_page_url": rp.default_donation_page.page_url if rp.default_donation_page else None,
    }
    return HttpResponse(
        template.render(
            data,
            request,
        ),
        content_type=f"text/{'plain' if template_name.endswith("txt") else 'html'}"
    )

```

benjaminEwhite: This approach should also allow the tests to be dried up and simplified. Should be able to have one test that uses pytest.mark.parametrize calls to vary on logo url and default donation page, and to iterate over the permitted templates.

Then would have a separate tests for:

- when unpermitted name used
- when template not found
- when rp id not included
- when rp not exist

benjaminEwhite (changes requested):
Left some inline comments and requests for changes.

apps/contributions/models.py
```diff
@@ -528,6 +528,118 @@ def handle_thank_you_email(self):
                 self.id,
             )
 
+    def send_recurring_contribution_canceled_email(self) -> None:
+        from apps.api.views import construct_rp_domain
+        from apps.emails.tasks import send_templated_email
```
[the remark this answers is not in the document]
nrh-cklimas: Removed in 1794655befcbbd033ef92582d84869c6ab578890

apps/contributions/models.py
nrh-cklimas: Removed in 1794655befcbbd033ef92582d84869c6ab578890

apps/contributions/models.py
nrh-cklimas: Added in 1794655befcbbd033ef92582d84869c6ab578890

apps/contributions/tests/test_models.py
nrh-cklimas: Updated in 41e41d828832662e61ddf5309fe6be9a1d8997d2

apps/contributions/tests/test_models.py
nrh-cklimas: 👍 Updated in 41e41d828832662e61ddf5309fe6be9a1d8997d2

apps/contributions/tests/test_models.py
nrh-cklimas: This is used to validate that the email is sending a link with the correct URL--used the method to generate this in cd447c8a5c61209168861ba1e5e07cde87e7a68b

apps/emails/views.py
```diff
@@ -30,11 +34,80 @@ def preview_recurring_contribution_reminder(request):
...
+def preview_recurring_contribution_canceled(request):
+    """
+    Previews a recurring contribution cancellation notification in the browser.
+    To view the plain text version, use a `text` querystring param. Only
+    accessible if DEBUG is set (see revengine/urls.py).
+    """
+    rp = RevenueProgram.objects.get(id=request.GET.get("rp"))
+    style = asdict(rp.transactional_email_style)
+    if request.GET.get("logo"):
+        style["logo_url"] = request.GET.get("logo")
```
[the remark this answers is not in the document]
nrh-cklimas: This is overriding the logo that is already set in `rp.transactional_email_style`, so I don't think we can change this.

apps/contributions/models.py
nrh-cklimas: It hit me that the two methods, `send_recurring_contribution_canceled_email` and `send_recurring_contribution_payment_updated_email`, are almost identical except for

- subject line
- which template to use

Maybe they should just be one method like `send_recurring_contribution_change_email` that takes one arg as to the type of change? What do you think?

apps/emails/views.py
nrh-cklimas: Going to defer on this until we agree on an approach on https://github.com/newsrevenuehub/rev-engine/pull/1359#discussion_r1457930561, because I think this is essentially the same question.

apps/contributions/models.py
nrh-cklimas: Going to defer until we resolve https://github.com/newsrevenuehub/rev-engine/pull/1359#discussion_r1457930561 because it might affect approach here.

apps/emails/views.py
nrh-cklimas: Added in 457dacb27f51d8df827994b2ae9f28fff7c14fa0

apps/emails/views.py
nrh-cklimas: Added in eb5bd97d76300bdbd9efc61ded888b9be45a23ce

benjaminEwhite (commented):
Reviewed again. Left some comments @nrh-cklimas re: function names and shared code.

apps/contributions/models.py
nrh-cklimas: I like this idea a lot--done in 0a1e6370b850cdff507ffac7388333665752a86e. I didn't add unit tests directly on the underlying method but I could if you think it's a good idea. We have coverage indirectly.

apps/contributions/models.py
nrh-cklimas: Changed this to `timestamp` in 34f77746f95ff9410a0070980e9e1bc36b31a7e1 and updated QA instructions because this affects confirmation email potentially.

apps/emails/views.py
nrh-cklimas: Took a stab at this in e7cd024fc89909489a31a536fbcc3fa6895fc33a.

benjaminEwhite (commented):
@nrh-cklimas left a couple comments here and then (as discussed) ended up doing a PR with a set of suggested changes into DEV-4220 branch. Thought we could huddle and talk through them so I can impart some of the not-yet documented reasoning behind some of the approaches and conventions.

Will link to that PR in the main ticket on JIRA.

benjaminEwhite (approved):
Left one comment you can resolve on your own @nrh-cklimas re: adding docstring.

Also worth adding to PR description "What's this do?" section a bullet about the existing `send_recurring_contribution_email_reminder` implementation shifting to use the new abstracted method for recurring contribution emails.

I'm approving, as above does not need re-review.

apps/contributions/models.py
```diff
@@ -554,41 +558,68 @@ def handle_thank_you_email(self):
                 self.id,
             )
 
-    def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.date) -> None:
-        # vs. circular import
-        from apps.api.views import construct_rp_domain
-        from apps.emails.tasks import send_templated_email
-
+    def send_recurring_contribution_change_email(self, subject_line: str, template_name: str) -> None:
```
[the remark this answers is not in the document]
nrh-cklimas: Done in 760e2ad6ba6ae42d45cacb03b6bc3993a99572b0

benjaminEwhite (approved):

PR #1372, opened 2024-01-15
DEV-2945: Fix IntegrationError: Invalid value for `(): elements should have a mounted Payment Element or Pay 

GuiMend:

n/a
#### What's this PR do?
Memoize stripe's PaymentElement `options` so that it doesn't trigger re-renders that might break stripe create payment function.

Fix issue that avoids users from donating

n/a
no
No, no changes in functionality were made. 

n/a
no
n/a
https://news-revenue-hub.atlassian.net/browse/DEV-2945
no
no

### DIFF
--- a/spa/src/components/donationPage/FinishPaymentModal/StripePaymentForm.tsx
+++ b/spa/src/components/donationPage/FinishPaymentModal/StripePaymentForm.tsx
@@ -132,18 +132,21 @@ export function StripePaymentForm({ payment, locale }: StripePaymentFormProps) {
 
   // for full options, see: https://stripe.com/docs/js/elements_object/create_payment_element
   // notably, can control fonts
-  const paymentElementOptions: StripePaymentElementOptions = {
-    fields: {
-      // we collected name, email, etc. in previous form, so we don't need to display these
-      // inputs in Stripe's payment element which defaults to including them. This means
-      // we need to add that data when we call stripe.confirmPayment.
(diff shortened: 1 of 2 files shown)


spa/src/components/donationPage/FinishPaymentModal/StripePaymentForm.tsx
```diff
@@ -132,18 +132,21 @@ export function StripePaymentForm({ payment, locale }: StripePaymentFormProps) {
...
-      // inputs in Stripe's payment element which defaults to including them. This means
-      // we need to add that data when we call stripe.confirmPayment.
-      billingDetails: 'never'
-    },
-    // This removes legal agreements that Stripe may display
-    terms: {
-      card: 'never'
-    }
-  };
+  const paymentElementOptions: StripePaymentElementOptions = useMemo(
```
nrh-cklimas, line 135: Let's move this outside the component itself, since the values here never change.

nrh-cklimas (changes requested):
I have a suggestion on how to unit test this. I think it is worth doing since if we don't memoize, we cause problems for end users.

1. Mock `PaymentElement`.
2. Render the component, then use [rerender](https://testing-library.com/docs/react-testing-library/api#rerender) to trigger an update--the prop change could be something trivial like changing `locale`.
3. Expect that the `PaymentElement` mock is called twice but with the same options object (e.g. use `toBe`).

PR #1357, opened 2024-01-03
DEV-4343 | update changed property names in in new portal contributions list and detail view

benjaminEwhite:
Please fill out each section even if it's just with "N/A"

#### Plan? (if this draft/incomplete indicate what you intend to do and how)

Initially, this PR takes DEV-4032... as its base, as that is the PR/ticket that changes API schema/field names, with which this PR aligns the SPA.   DEV-4032 has not yet been merged to main. Once that has happened, base should change to main.

This is also to say that this PR is blocked for merge by DEV-4032.

#### What's this PR do?

In working on DEV-4032, I discovered some fields in portal contributions API endpoint schema that could be retired, and others where renaming made sense.  

This PR aligns the SPA with the API schema of backend, as described [here in this API schema file](https://github.com/newsrevenuehub/rev-engine/blob/f69b1d2a4c3272392916b9208fb632b2d2f62de9/apps/api/schemas/portal-contribution-list.openapi.yml).

Specifically:

- `credit_card_expiration_date` becomes `card_expiration_date` (to standardize all CC-related fields as being prefixed with `card_`)
- `last4` becomes `card_last_4`
- `id` is added as a new field
- `payment_provider_id` is removed, as it is not needed
- `provider_customer_id` is removed, as it is not needed

Additionally, this PR opportunistically solves for [DEV-4427](https://news-revenue-hub.atlassian.net/browse/DEV-4427)

#### Why are we doing this? How does it help us?

Part of shipping new portal.

#### Are there detailed, specific, step-by-step testing instructions in the associated ticket?

Yes.

#### Did this PR make changes to the user interface that may require changes to the user-facing docs?

No.

#### Have automated unit tests been added? If not, why?

No, but existing ones have been updated to reflect new schema and updated type.

#### How should this change be communicated to end users?

N/A

#### Are there any smells or added technical debt to note?

No

#### Has this been documented? If so, where?

Sort of. The API schema referenced in description above documents the schema of data returned by endpoint (though that happens in a different PR).

#### What are the relevant tickets? Add a link to any relevant ones.

- [DEV-4343](https://news-revenue-hub.atlassian.net/browse/DEV-4343)
- [DEV-4032](https://news-revenue-hub.atlassian.net/browse/DEV-4032)
- [DEV-4427](https://news-revenue-hub.atlassian.net/browse/DEV-4427)

#### Do any changes need to be made before deployment to production (adding environment variables, for example)? If so, open a ticket and link it to the ticket for this PR and list it here:

No

[...]

PR #1382, opened 2024-01-22
DEV-3869: Implement sorting dropdown on contribution list

GuiMend:

- Wait for https://github.com/newsrevenuehub/rev-engine/pull/1357 to be merged

#### What's this PR do?
- Creates Sort component
- Implement Sorting functionality in contribution list page 

- Facilitate visibility of contributions
- yes
- yes
- yes
- Sorting enabled in contributions list
- no
- no
https://news-revenue-hub.atlassian.net/browse/DEV-3869
- no
Epic: https://news-revenue-hub.atlassian.net/browse/DEV-3782


spa/src/components/base/Select/Select.tsx
nrh-cklimas, line 1: Can you talk through advantages of using this over `<TextField select>`? [According to the docs](https://v4.mui.com/components/text-fields/#select), that prop causes the component to use `Select`, so it feels possibly redundant.

spa/src/components/portal/ContributionsList/ContributionsList.test.tsx
```diff
@@ -14,6 +15,8 @@ jest.mock('./ContributionDetail/ContributionDetail');
 jest.mock('./ContributionItem/ContributionItem');
 jest.mock('./ContributionFetchError');
 
+const refetch = jest.fn();
```
nrh-cklimas, line 18: If we move this to a global in the module, I think we need to call `refetch.mockClear()` in `beforeEach` so that test runs don't potentially pollute each other. I personally prefer to use a specific mock in a test so there's no possibility for this to get missed, but it's a code style concern.

spa/src/components/portal/ContributionsList/ContributionsList.tsx
```diff
@@ -21,6 +29,12 @@ export function ContributionsList() {
   const [selectedContributionEl, setSelectedContributionEl] = useState<HTMLAnchorElement | null>(null);
   let content: ReactChild;
 
+  useEffect(() => {
```
nrh-cklimas, line 32: Instead of using an effect, let's change `usePortalContributionList` to put the query string into the [React query key](https://tanstack.com/query/latest/docs/react/guides/query-keys). This should trigger a refetch for free and means that any other consumers of the hook don't have to think about it.

spa/src/components/common/Sort/Sort.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const [value, setValue] = useState(options[0].value);
+
+  const handleChange = (event: ChangeEvent<{ value: unknown }>) => {
+    setValue(event.target.value as string);
+    if (typeof onChange === 'function') onChange(event.target.value as string);
+  };
+
+  return (
+    <SortWrapper className={className!}>
+      <Typography variant="body2" color="textSecondary">
```
nrh-cklimas, line 45: Right now there's no relationship between this label and the field. This doesn't get flagged by axe because the select's role is `button`, but I'd suggest either using the `labelId` prop here, or `<TextField>` because that handles labelling for you.

spa/src/components/common/Sort/Sort.tsx
```diff
@@ -0,0 +1,67 @@
...
+    ),
+    selectedLabel: 'Amount',
+    value: 'amount'
+  }
+];
+
+const Sort = ({ options, onChange, className = '' }: SortProps) => {
+  const [value, setValue] = useState(options[0].value);
+
+  const handleChange = (event: ChangeEvent<{ value: unknown }>) => {
```
nrh-cklimas, line 38: `value` will always be `string` in a change event, won't it?

spa/src/components/common/Sort/Sort.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { SortWrapper } from './Sort.styled';
+
+import { Typography } from '@material-ui/core';
+import { Select } from 'components/base';
+
+export interface SortProps extends InferProps<typeof SortPropTypes> {
+  onChange?: (value: string) => void;
+}
+
+export const CONTRIBUTION_SORT_OPTIONS = [
```
nrh-cklimas, line 13: I feel like this belongs more in ContributionsList, unless you see this being used elsewhere.

nrh-cklimas (changes requested):
Two general things:

- I think right now there's too much vertical space between the menu and the contribution list. Looking in Figma, I think it should be 25px.
- Would it be a lot of work to change the mock to honor the sorting query params? Right now, it's going to be hard to QA because changing the sort doesn't cause anything to change.

spa/src/components/base/Select/Select.tsx
GuiMend: Now that you mentioned it, I don't think there are any differences. I checked the html and they seem the same. I guess MUI just made them separate to make it easier to glance at?
Would you rather I removed this `Select` component and added it's implementation directly to the `Sort` component?

spa/src/components/common/Sort/Sort.tsx
GuiMend: I believe so, but the `Select` TS only accepts onChange with `value: unkown`
image

spa/src/components/common/Sort/Sort.tsx
```diff
@@ -0,0 +1,67 @@
...
+    value: 'amount'
+  }
+];
+
+const Sort = ({ options, onChange, className = '' }: SortProps) => {
+  const [value, setValue] = useState(options[0].value);
+
+  const handleChange = (event: ChangeEvent<{ value: unknown }>) => {
+    setValue(event.target.value as string);
+    if (typeof onChange === 'function') onChange(event.target.value as string);
```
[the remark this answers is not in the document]
GuiMend: Made it simpler using optional chaining:
```js
onChange?.(event.target.value as string);
```

spa/src/components/common/Sort/Sort.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { Select } from 'components/base';
+
+export interface SortProps extends InferProps<typeof SortPropTypes> {
+  onChange?: (value: string) => void;
+}
+
+export const CONTRIBUTION_SORT_OPTIONS = [
+  {
+    label: (
+      <p style={{ margin: 0 }}>
```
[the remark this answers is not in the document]
GuiMend: Used `span`

spa/src/components/base/Select/Select.tsx
nrh-cklimas: I think so, if it's a drop-in replacement here.

spa/src/components/common/Sort/Sort.tsx
```diff
@@ -10,42 +10,20 @@ export interface SortProps extends InferProps<typeof SortPropTypes> {
...
   const handleChange = (event: ChangeEvent<{ value: unknown }>) => {
     setValue(event.target.value as string);
-    if (typeof onChange === 'function') onChange(event.target.value as string);
+    onChange?.(event.target.value as string);
   };
 
   return (
     <SortWrapper className={className!}>
-      <Typography variant="body2" color="textSecondary">
+      <Typography variant="body2" color="textSecondary" id="sort">
```
nrh-cklimas, line 23: Not sure if this will still be needed with a switch to TextField, but the id should be a prop--otherwise two Sorts on the page will conflict with each other.

spa/src/components/common/Sort/Sort.tsx
nrh-cklimas: Let's revisit with the change to TextField.

GuiMend: Fixed with the TextField 👍

nrh-cklimas (approved):

PR #1393, opened 2024-01-30
DEV-3888 | API for updating payment method on recurring contribution in new portal

benjaminEwhite:

### DIFF
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -635,14 +635,16 @@ class Contribution(IndexedTimeStampedModel):
 
     @cached_property
     def card(self) -> stripe.Card | None:
-        if not (cust_id := self.provider_customer_id):
+        """Return card object for contribution, if one exists"""
+        # TODO: wed this to metadata version, and also determine conditions under which updates would occur for contributions where this is only available on the customer
+        # potentially adding a "can be updated" property or something like that.
+        if not (pm := self.stripe_payment_method):
+            logger.warning(
+                "`Contribution.card` called on contribution with ID %s that can't be linked to a stripe payment method",
+                self.id,
+            )
             return None
-        customer = stripe.Customer.retrieve(
-            cust_id,
-            stripe_account=self.donation_page.revenue_program.payment_provider.stripe_account_id,
-            expand=["default_source"],
-        )
-        return customer.default_source if customer.default_source and customer.default_source.object == "card" else None
+        return pm.card if pm.type == "card" else None
 
     @property
     def card_brand(self) -> str:
@@ -654,7 +656,8 @@ class Contribution(IndexedTimeStampedModel):
 
     @property
     def card_owner_name(self) -> str:
-        return self.card.name if self.card else ""
+        pm = self.stripe_payment_method
+        return pm.billing_details.name if pm and pm.billing_details and pm.type == "card" else ""
 
     @property
     def card_last_4(self) -> str:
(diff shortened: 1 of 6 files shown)


apps/contributions/models.py
```diff
@@ -635,14 +635,14 @@ def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
 
     @cached_property
     def card(self) -> stripe.Card | None:
-        if not (cust_id := self.provider_customer_id):
+        """Return card object for contribution, if one exists"""
```
benjaminEwhite, line 638: The former approach didn't take advantage of affordances of tracking `provider_payment_method_id` and keeping that in sync via webhooks. 

If we assume that that value will usually be synced to our model, then that gives us a single stable referent for where to look for card (on the contribution's corresponding payment method.

apps/contributions/models.py
```diff
@@ -633,32 +634,33 @@ def stripe_payment_intent(self) -> stripe.PaymentIntent | None:
             )
             return None
 
-    @cached_property
-    def card(self) -> stripe.Card | None:
```
benjaminEwhite, line 637: Turns out there's a much simpler way to get card data. We already have a contribution.provider_payment_method_details property for tracking PM data.  As long as we don't override that with card data instead of PM data (which sometimes happens in status, quo, if i'm not mistaken), webhooks should keep this property in sync with relevant pm in stripe.

apps/contributions/serializers.py
```diff
@@ -850,8 +851,12 @@ def get_payment_type(self, instance):
 
 
 class PortalContributionBaseSerializer(serializers.ModelSerializer):
+    card_brand = serializers.CharField(read_only=True, allow_blank=True)
+    card_expiration_date = serializers.CharField(read_only=True, allow_blank=True)
+    card_last_4 = serializers.CharField(read_only=True, allow_blank=True)
+    last_payment_date = serializers.DateTimeField(source="_last_payment_date", read_only=True, allow_null=True)
```
benjaminEwhite, line 857: Have to explicitly add these with "allow_blank=True" because otherwise DRF will omit the key from the response object when serializing.

apps/contributions/webhooks.py
```diff
@@ -232,12 +225,6 @@ def handle_subscription_canceled(self):
             "`StripeWebhookProcessor.handle_subscription_canceled` updated contribution",
         )
 
-    def handle_payment_method_attached(self):
```
x110dc, line 235: I understand that we shouldn't handle this if we're saving the wrong detail. But I think we should refactor it do it correctly instead of removing it completely. There are ways that the payment method could change outside of the contributor doing it in our portal (for example, Stripe has arrangements with card issuers where they sometimes replace a card when a cardholder has a card replaced, or the expiration is updated) and if we don't capture this (via webhooks) we'll have (and show) out-of-date info to contributors. It's okay by me if that's a separate ticket. Or if I'm missing something here let me know.

x110dc (changes requested):
One request/question.

apps/contributions/webhooks.py
benjaminEwhite: Added a TODO in dbb60bd8111.

Also updated PR description, and removed ITS ticket around retiring the webhook from Stripe envs.

In short term, we'll retain the handler, but handler just returns with no other action. In DEV-4450, we'll re-implement correctly.

x110dc (approved):

njharman (changes requested):

apps/contributions/serializers.py
```diff
@@ -897,13 +902,28 @@ class Meta:
...
         read_only_fields = PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS
 
+    def update(self, instance, validated_data) -> Contribution:
+        provider_payment_method_id = validated_data.get("provider_payment_method_id")
+        if provider_payment_method_id:
+            instance.update_payment_method_for_subscription(
+                provider_payment_method_id=provider_payment_method_id,
+            )
+        for key, value in validated_data.items():
+            setattr(instance, key, value)
```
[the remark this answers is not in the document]
benjaminEwhite: Added a test in  a8f223bb that demonstrates this, but if you send empty string for `provider_payment_method_id`, the request is valid, but there will not be a dict item in validated data.  

See also default values for allow_blank indicated: https://www.django-rest-framework.org/api-guide/fields/#charfield


> Should it not be skipped over in the for k,v loop?

We want to update it in the db, so should not be skipped. It's just we want to ensure the side effect here ahead of updating in db.

apps/contributions/tests/test_models.py
```diff
@@ -1659,6 +1683,29 @@ def test__expanded_pi_for_cancelable_modifiable(self, provider_payment_id, mocke
...
+    def test_card_expiration_date(self, payment_data, expected):
+        pass
+
+    @pytest.mark.parametrize("payment_data,expected", ((None, ""),))
+    def test_card_owner_name(self, payment_data, expected):
+        pass
+
+    @pytest.mark.parametrize("payment_data,expected", ((None, ""),))
+    def test_card_last_4(self, payment_data, expected):
+        pass
```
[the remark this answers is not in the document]
benjaminEwhite: I missed this. Meant to come back to stubs, and then visually scanning later, thought this was actual implementation (and test coverage didn't convince me otherwise, since this code gets touched in other tests).

See 6a84e48b

njharman (approved):

PR #1396, opened 2024-02-01
DEV-4452 | spike our implementation of double submit csrf token strategy is broken making it less secure

benjaminEwhite:

### DIFF
--- a/spa/src/ajax/axios.js
+++ b/spa/src/ajax/axios.js
@@ -24,14 +24,13 @@ function getCsrfTokenFromCookie() {
   if (parts.length === 2) return parts.pop().split(';').shift();
 }
 
-Axios.interceptors.request.use(
-  (request) => {
-    const csrfToken = getCsrfTokenFromCookie();
-    if (csrfToken) request.headers[CSRF_HEADER] = csrfToken;
-    return request;
-  },
-  (error) => Promise.reject(error)
-);
+const sendCookieCsrfAsHeader = (request) => {
+  const csrfToken = getCsrfTokenFromCookie();
+  if (csrfToken) request.headers[CSRF_HEADER] = csrfToken;
+  return request;
+};
+
+Axios.interceptors.request.use(sendCookieCsrfAsHeader, (error) => Promise.reject(error));
 
 /**
  * A Response interceptor.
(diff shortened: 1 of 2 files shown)


apps/api/tests/test_views.py
```diff
@@ -173,8 +171,6 @@ def test_post_happy_path(self, user, expected_orgs, many_orgs, many_rps, api_cli
         else:
             assert _user["role_type"] is None
 
-        assert response.json()["csrftoken"]
-        assert response.cookies[settings.CSRF_COOKIE_NAME]
```
benjaminEwhite, line 177: It no longer makes sense to test for presence of this cookie in this test because the tokenobtainpaircookieview does not have any permissions (inclusive of our new double submit csrf permission), and csrf is not enforced for this endpoint.

In actual practice, the user probably will have a csrf cookie set at this point, but it will likely have been set when they went to magic link request page, which makes an api request to /api/v1/contrib-email/ which causes a csrf cookie (if none was previously attached).

apps/api/tests/test_views.py
```diff
@@ -474,14 +470,12 @@ def test_response_when_email_token_mismatch(self):
     def test_response_with_valid_token(self):
         response = self.client.post(self.url, {"email": self.contributor.email, "token": self.valid_token})
         self.assertEqual(response.status_code, 200)
-        self.assertIn("csrftoken", response.data)
         self.assertEqual(response.data["detail"], "success")
         self.assertEqual(response.data["contributor"]["id"], self.contributor.pk)
 
-    def test_response_sets_token_and_csrf_cookies(self):
```
benjaminEwhite, line 481: This view is actually marked as `csrf_exempt` when referenced in urls.py. So we should not in fact expect it to have csrf set.

apps/api/views.py
```diff
@@ -223,9 +221,6 @@ def post(self, request, *args, **kwargs):
         # Generate long-lived token
         long_lived_token = str(refresh.long_lived_access_token)
 
-        # Get anti-CSRF token
-        csrf_token = csrf.get_token(self.request)
```
benjaminEwhite, line 227: cookie-level csrf generation is already happening. if we create a separate csrf token here, it's guaranteed to be out of sync with the one in request cookies.

apps/organizations/views.py
```diff
@@ -280,7 +283,6 @@ def get_stripe_account_link_return_url(request):
 
 @reversion.create_revision()
 @api_view(["POST"])
-@permission_classes([IsAuthenticated, HasRoleAssignment])
```
benjaminEwhite, line 283: These are the default permissions set in base.py

apps/api/views.py
```diff
@@ -106,7 +105,6 @@ def post(self, request, *args, **kwargs):
         response = Response(
             {
                 "detail": "success",
-                "csrftoken": csrf.get_token(self.request),
```
benjaminEwhite, line 109: See next comment

PR #1407, opened 2024-02-09
DEV-4467: Portal pre-release refinements

GuiMend:

n/a
#### What's this PR do?
- Align the loading, error, empty contribution components in the correct place on screen
- Refetches data 15 seconds after update or cancel contribution
- Redirects to magic link page if not logged in
- Remove sentry alert from being triggered if stripe returns "card declined" error
- Refining new portal for launch
- yes
- no
- yes
- n/a
- no
- no
- https://news-revenue-hub.atlassian.net/browse/DEV-4467
- no
- New Portal UI Updates Epic: https://news-revenue-hub.atlassian.net/browse/DEV-3780

### DIFF
--- a/spa/src/ajax/portal-axios.ts
+++ b/spa/src/ajax/portal-axios.ts
@@ -1,0 +1,15 @@
+import { PORTAL } from 'routes';
+import Axios from './axios';
+import { AxiosError } from 'axios';
+
+export default Axios;
+
+/**
+ * Overrides response interceptor from base Axios instance
+ */
+Axios.interceptors.response.use((success) => success, handleResponseError);
+
+function handleResponseError(error: AxiosError) {
+  if (error?.name === 'AuthenticationError') return window.location.assign(PORTAL.ENTRY);
+  return Promise.reject(error);
+}
--- a/spa/src/ajax/portal-axios.ts
+++ b/spa/src/ajax/portal-axios.ts
@@ -1,15 +1,47 @@
+import { CSRF_HEADER, LS_CSRF_TOKEN, REVENGINE_API_VERSION } from 'appSettings';
+import axios, { AxiosError } from 'axios';
 import { PORTAL } from 'routes';
-import Axios from './axios';
-import { AxiosError } from 'axios';
+
+export const apiBaseUrl = `/api/${REVENGINE_API_VERSION}/`;
+
+const Axios = axios.create({
+  baseURL: apiBaseUrl,
+  timeout: 15000,
+  withCredentials: true // allow setting/passing cookies
+});
 
 export default Axios;
+
+/**
+ * A Request interceptor.
+ * first callback intercepts successfully formed requests
+ * second callback handles errors, so pass through
+ */
+Axios.interceptors.request.use(
+  (request) => {
+    const csrfToken = localStorage.getItem(LS_CSRF_TOKEN);
+    if (csrfToken) request.headers[CSRF_HEADER] = csrfToken;
+    return request;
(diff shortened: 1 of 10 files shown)


nrh-cklimas:
@GuiMend how would you feel about creating a function similar to `ajax/axios.js` but tailored for use in the portal? There is similar logic there for 401 errors and the advantage to me is we don't have to think about it in each query. I think potentially the Axios interceptor could do the redirect to the portal login for us instead of passing back a particular error, but I haven't thought this through fully.

spa/src/ajax/portal-axios.ts
```diff
@@ -0,0 +1,15 @@
+import { PORTAL } from 'routes';
+import Axios from './axios';
```
nrh-cklimas, line 2: I think we need a new Axios instance here; otherwise we're getting the interceptors used in the admin dash, and our interceptors will fire in the admin dash also.

spa/src/ajax/portal-axios.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Overrides response interceptor from base Axios instance
+ */
+Axios.interceptors.response.use((success) => success, handleResponseError);
+
+function handleResponseError(error: AxiosError) {
+  if (error?.name === 'AuthenticationError') return window.location.assign(PORTAL.ENTRY);
+  return Promise.reject(error);
+}
```
nrh-cklimas, line 15:
```suggestion
function handleResponseError(error: AxiosError) {
  if (error?.name === 'AuthenticationError') {
    window.location.assign(PORTAL.ENTRY);
  }
  
  return Promise.reject(error);
}
```

I'm fuzzy on whether JS code would continue to execute after `location.assign()` is called, but it seems like it's a good idea to always return the error.

spa/src/hooks/usePortalContribution.test.tsx
```diff
@@ -111,6 +115,20 @@ describe('usePortalContribution', () => {
       await waitFor(() => expect(axiosMock.history.get.length).toBe(1));
       expect(result.current.contribution).toBeUndefined();
     });
+
+    it('redirects user to magic link (login) page if GET error = Authentication Error', async () => {
```
nrh-cklimas, line 119: I'd prefer to move these tests (and those in usePortalContributionList) into a new test file for `portal-axios.ts`, since they're testing the behavior over there.

spa/src/hooks/usePortalContribution.tsx
```diff
@@ -107,7 +114,6 @@ export function usePortalContribution(contributorId: number, contributionId: num
         });
       },
       onError: (error: AxiosError) => {
-        console.error('[usePortalContribution:cancelContribution] ', error);
```
nrh-cklimas, line 110: Curious, why remove this line? Will the error get logged elsewhere?

nrh-cklimas (changes requested):

spa/src/hooks/usePortalContribution.test.tsx
GuiMend: Sure. I had left them there because they were already implemented

spa/src/hooks/usePortalContribution.tsx
GuiMend: Shouldn't have. Re-added deleted line

nrh-cklimas (approved):

benjaminEwhite (approved):
Backend changes LGTM.

PR #1429, opened 2024-02-23
DEV-4482: Don't let contributors see rejected or processing contributions

nrh-cklimas:

n/a

#### What's this PR do?

Filters out rejected and processing contributions from new endpoints. They don't appear in contribution lists and trying interact with detail (GET, PATCH, DELETE) returns a 404.

These shouldn't be visible to contributors.

Yes.

No..?

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-4482

No.

No.

### DIFF
--- a/apps/contributions/tests/test_views.py
+++ b/apps/contributions/tests/test_views.py
@@ -1744,6 +1744,23 @@ class TestPortalContributorsViewSet:
         response = api_client.get(reverse("portal-contributor-contributions-list", args=(contributor.id,)))
         assert response.status_code == status.HTTP_403_FORBIDDEN
 
+    @pytest.mark.parametrize("status", (ContributionStatus.FAILED, ContributionStatus.PROCESSING))
+    def test_contributions_list_hides_statuses(
+        self,
+        status,
+        api_client,
+        monthly_contribution,
+        one_time_contribution,
+        portal_contributor_with_multiple_contributions,
+    ):
+        contributor = portal_contributor_with_multiple_contributions[0]
+        one_time_contribution.status = status
+        one_time_contribution.save()
+        api_client.force_authenticate(contributor)
+        response = api_client.get(reverse("portal-contributor-contributions-list", args=(contributor.id,)))
+        assert len(response.json()["results"]) == 1
+        assert response.json()["results"][0]["id"] == monthly_contribution.id
+
     def test_contribution_detail_get_happy_path(
         self,
         api_client,
@@ -1852,6 +1869,20 @@ class TestPortalContributorsViewSet:
         assert response.status_code == status.HTTP_404_NOT_FOUND
         assert response.json() == {"detail": "Contribution not found"}
 
+    @pytest.mark.parametrize("contribution_status", (ContributionStatus.FAILED, ContributionStatus.PROCESSING))
+    def test_contributions_detail_when_hidden_status(
+        self, contribution_status, api_client, one_time_contribution, portal_contributor_with_multiple_contributions
+    ):
+        contributor = portal_contributor_with_multiple_contributions[0]
+        one_time_contribution.status = contribution_status
+        one_time_contribution.save()
+        api_client.force_authenticate(contributor)
+        response = api_client.get(
+            reverse("portal-contributor-contribution-detail", args=(contributor.id, one_time_contribution.id))
+        )
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+        assert response.json() == {"detail": "Contribution not found"}
+
     @pytest.mark.parametrize("request_data", ({}, {"provider_payment_method_id": "pm_123"}))
     def test_contribution_detail_patch_happy_path(
         self,
@@ -1955,6 +1986,26 @@ class TestPortalContributorsViewSet:
         assert response.json() == {"detail": "Problem updating contribution"}
         mock_pm_attach.assert_called_once()
 
+    @pytest.mark.parametrize("contribution_status", (ContributionStatus.FAILED, ContributionStatus.PROCESSING))
+    def test_contributions_detail_patch_when_hidden_status(
+        self, contribution_status, api_client, one_time_contribution, portal_contributor_with_multiple_contributions
+    ):
+        contributor = portal_contributor_with_multiple_contributions[0]
+        one_time_contribution.status = contribution_status
+        one_time_contribution.save()
+        api_client.force_authenticate(contributor)
+        response = api_client.patch(
+            reverse(
+                "portal-contributor-contribution-detail",
+                args=(
+                    contributor.id,
+                    one_time_contribution.id,
+                ),
+            )
+        )
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+        assert response.json() == {"detail": "Contribution not found"}
+
     def test_contribution_detail_delete_happy_path(
         self, api_client, portal_contributor_with_multiple_contributions, mocker
     ):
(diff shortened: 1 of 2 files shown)


apps/contributions/views.py
```diff
@@ -577,6 +578,10 @@ class PortalContributorsViewSet(viewsets.GenericViewSet):
         "status",
     ]
 
+    # Contributors should never see contributions with these statuses, or
+    # interact with them (e.g. delete or patch them).
+    HIDDEN_STATUSES = [ContributionStatus.FAILED, ContributionStatus.PROCESSING]
```
benjaminEwhite, line 583: I think we also want `REJECTED` and `FLAGGED` here.

apps/contributions/views.py
```diff
@@ -587,6 +592,12 @@ def _get_contributor_and_check_permissions(self, request, contributor_id):
         self.check_object_permissions(request, contributor)
         return contributor
 
+    def _prune_contributions_queryset(self, queryset: ContributionQuerySet) -> ContributionQuerySet:
+        """Removes contributions from a contribution queryset that should never be visible to a contributor."""
+        # This is needed bcecause the queryset property of this class is tied to
+        # contributors, not contributions.
+        return queryset.exclude(status__in=self.HIDDEN_STATUSES)
+
```
benjaminEwhite, line 600: Notwithstanding my advice when we huddled @nrh-cklimas i think this doesn't really need its own separate method. I didn't realize that it would be a one liner/this easy. In that case, might as well just do it inline at line 623. In this case, I think the "native" django code is self-documenting.

apps/contributions/tests/test_views.py
```diff
@@ -2034,6 +2085,26 @@ def test_contribution_detail_delete_when_subscription_not_cancelable(
         assert response.status_code == status.HTTP_400_BAD_REQUEST
         assert response.json() == {"detail": "Cannot cancel contribution"}
 
+    @pytest.mark.parametrize("contribution_status", (ContributionStatus.FAILED, ContributionStatus.PROCESSING))
```
benjaminEwhite, line 2088: See note below, but think we should also have rejected and flagged here.

benjaminEwhite (changes requested):
Looks good. Think we need to include REJECTED and FLAGGED statuses. And probably can just inline some of the logic (in spit of what i advised when we huddled).

apps/contributions/views.py
nrh-cklimas: Updated in 396e562c9356b2a04efeef513f24ab32b70da47f

apps/contributions/tests/test_views.py
nrh-cklimas: Updated in db58f0c93f53169becaee0b2d5e1f34955b4827c

apps/contributions/views.py
nrh-cklimas: Updated in db58f0c93f53169becaee0b2d5e1f34955b4827c

apps/contributions/tests/test_views.py
```diff
@@ -1744,6 +1750,31 @@ def test_contributions_list_when_im_not_owning_contributor(
...
 
+    @pytest.mark.parametrize(
+        "status",
+        (
+            ContributionStatus.FAILED,
+            ContributionStatus.FLAGGED,
+            ContributionStatus.PROCESSING,
+            ContributionStatus.REJECTED,
+        ),
+    )
```
benjaminEwhite, line 1761: **Optional suggestion**

Chris, this is more a "just so you know". Since these same statuses get used across several tests, another approach would be to create a fixture called something like "excluded_contributor_status" and then add that as a parameter for the test functions that need it.

```
@pytest.fixture(params=[ContributionStatus.FAILED, ..., ContributionStatus.REJECTED]
def excluded_contributor_status(self, request):
    return request.param
```

if you include `excluded_contributor_status` as param (instead of current `status`), pytest will run the test once for each of the statuses in the fixture params.

apps/contributions/tests/test_views.py
```diff
@@ -1852,6 +1883,28 @@ def test_contribution_detail_get_when_not_own_contribution(
...
+    @pytest.mark.parametrize(
+        "contribution_status",
+        (
+            ContributionStatus.FAILED,
+            ContributionStatus.FLAGGED,
+            ContributionStatus.PROCESSING,
+            ContributionStatus.REJECTED,
+        ),
+    )
+    def test_contributions_detail_when_hidden_status(
```
benjaminEwhite, line 1895: **Optional suggestion**

Looks like for the different detail methods, the test code is pretty much the same. Could combine into a single test, and parametrize the HTTP method (which will give you cross product of status params + method params -- also true if you went with the fixture approach mentioned above).

```
    @pytest.mark.parametrize(
        "contribution_status",
        (
            ContributionStatus.FAILED,
            ContributionStatus.FLAGGED,
            ContributionStatus.PROCESSING,
            ContributionStatus.REJECTED,
        ),
    )
    @pytest.mark.parametrize("http_method", ("delete", "get", "patch"))
  def test_contributions_detail_view_when_hidden_status(
        self, contribution_status, http_method, api_client, one_time_contribution, portal_contributor_with_multiple_contributions
    ):
        contributor = portal_contributor_with_multiple_contributions[0]
        one_time_contribution.status = contribution_status
        one_time_contribution.save()
        api_client.force_authenticate(contributor)
        response = getattr(api_client, http_method)(
            reverse(
                "portal-contributor-contribution-detail",
                args=(
                    contributor.id,
                    one_time_contribution.id,
                ),
            )
        )
        assert response.status_code == status.HTTP_404_NOT_FOUND
        assert response.json() == {"detail": "Contribution not found"}
```

apps/contributions/views.py
```diff
@@ -577,6 +578,15 @@ class PortalContributorsViewSet(viewsets.GenericViewSet):
...
 
+    # Contributors should never see contributions with these statuses, or
+    # interact with them (e.g. delete or patch them).
+    HIDDEN_STATUSES = [
+        ContributionStatus.FAILED,
+        ContributionStatus.FLAGGED,
+        ContributionStatus.PROCESSING,
+        ContributionStatus.REJECTED,
+    ]
+
```
benjaminEwhite, line 589: @nrh-cklimas I missed this first time around, but I don't think we want to excluded "failed". That status can happen (among other reasons) if the payment method on a recurring contribution expires.

Also don't see "failed" mention in AC for ticket.

benjaminEwhite (changes requested):
@nrh-cklimas I missed on first time review that "FAILED" is one of the excluded statuses in this PR, and as commented in line, I don't think that's one we should be excluding.

Other than that, things look good. I made some totally optional suggestions for drying up tests.

Only other request is to update PR description to reflect the final state of excluded contribution statuses.

apps/contributions/tests/test_views.py
nrh-cklimas: Curious what your opinion would be of importing HIDDEN_STATUSES and using that as a basis for the fixture. We sometimes do similar things on the SPA side but there's always the danger of the test being vacuously true--e.g. if HIDDEN_STATUSES has incorrect values, tests still pass.

apps/contributions/views.py
nrh-cklimas: Updated in 2ff011339f5e1417c308344bbd71437ed3d6eee4

apps/contributions/tests/test_views.py
benjaminEwhite: I like that idea (but leave it up to you whether or not you want to act on it).

benjaminEwhite (approved):
LGTM. Approved.

apps/contributions/tests/test_views.py
nrh-cklimas: I got a little lost in details trying to do this so I think I'll try this next time and ask for help.

PR #1440, opened 2024-03-04
DEV-4518 | Remove double submit CSRF pattern

benjaminEwhite:

This is good to do in its own right, but is also a blocker for [DEV-4131](https://news-revenue-hub.atlassian.net/browse/DEV-4131) (Django upgrade) because our custom CSRF implementation is effected by breaking changes in Django 4.2.

#### What's this PR do?

This PR removes the current double submit CSRF pattern.  Changes touch both the frontend and backend.

_Server-side changes_

- Removes code that sent second CSRF token in response body when authing
- Removes flawed implementation of double submit (which was intended to compare cookie and headers token sent back by clients, but did not). As it turns out, Django has double-submit pattern by default, and the attempted redundant implementation was a red herring.
- Adds a new method decorator `apps.common.views.csrf_protect_json` which allows a DRF-involved view method (or class) to be decorated such that if valid CSRF token not included in request headers, the request will fail with 403 and the response object will be JSON. Before, we were relying on standard enforce_csrf which was causing SPA ajax requests to sometimes get an HTML response.  This decorator is used on `PaymentViewset` since that viewset of necessity is not protected by JWT authentication. By using this decorator there, CSRF requests to the payment endpoint (for delete and create) are cut off.
- Opportunistically moves `AuthorizedContributorRequestsTest` to use Pytest since that code was touched, and renames to `TestAuthorizedContributorRequests`
- Removes CSRF exemption from contributor-verify URL since it was redundant (DRF-involved views do not have CSRF protection, by default -- you need to opt in to that).
- Moves `@method_decorator(ensure_csrf_cookie)` to `ReactAppView`. That way any time the SPA is loaded, it will be guaranteed to get CSRF token in response cookies.  That's where it should have been all along, and likely that occasional bugs resulted from it being on live detail view.
- Updates the `api_client` test fixture so that CSRF checks are enforced. By default, they are not enforced in tests, and we actually want our API-involved tests to factor in CSRF and tell us if there are problems.

_SPA-side changes_

- Refactors code that was sending CSRF token from local storage in requests to API. This was a flawed implementation. Now  we correctly pull cookie value for CSRF token and if found, include in request headers.
- Removes CSRF token removal from  local storage from logout function
- Removes code that set CSRF token in local storage after successful login
- Removes code that set CSRF token in contributor verify
- Removes setting CSRF header from some hooks. This gets set in Axios interceptors.
- Updates tests and test fixtures in line with above.
- Adds [js-cookie](https://github.com/js-cookie/js-cookie) as a dependency. We use `react-cookie` for cookie management elsewhere in the SPA, but that's not feasible in Axios configuration as that doesn't happen inside hooks.

- Remove tech debt / flawed implementation in favor of idiomatic CSRF approach
- Unblock Django 4.2 upgrade

There are instructions but not detailed because the instructions are to go through well known QA flows involving authentication to ensure no regressions.

No.

Yes (mainly existing tests being altered based on changes)

N/A

No. Removes a lot of tech debt.

In progress. I'm writing up a [detailed Confluence page](https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2702737433/RevEngine+Authentication+and+CSRF+Protection+Guide) that describes ins and outs of JWT and CSRF tokens. 

- [DEV-4518](https://news-revenue-hub.atlassian.net/browse/DEV-4518)
- [DEV-4131](https://news-revenue-hub.atlassian.net/browse/DEV-4131)

No.

No.

[DEV-4131]: https://news-revenue-hub.atlassian.net/browse/DEV-4131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-4518]: https://news-revenue-hub.atlassian.net/browse/DEV-4518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/api/authentication.py
```diff
@@ -75,7 +58,7 @@ def get_user(self, validated_token):
             return super().get_user(validated_token)
 
 
-class MagicLinkTokenAuthenticationBase(JWTHttpOnlyCookieAuthentication):
+class ShortLivedTokenAuthentication(JWTHttpOnlyCookieAuthentication):
```
benjaminEwhite, line 61: The only user of `MagicLinkTokenAuthenticationBase` is `ShortLivedTokenAuthentication` so no need for this indirection.

apps/api/permissions.py
```diff
@@ -148,7 +148,9 @@ def __init__(self, *args, **kwargs):
         Flag = get_waffle_flag_model()
         self.flag = Flag.objects.filter(name=MAILCHIMP_INTEGRATION_ACCESS_FLAG_NAME).first()
         if not self.flag:
-            raise ApiConfigurationError()
+            raise ApiConfigurationError(
```
benjaminEwhite, line 151: There were some hard to run down failed tests that I encountered in making these changes, and this was the cause. Added better messaging here so that will be clear in the future if happens.

apps/api/permissions.py
```diff
@@ -162,4 +164,6 @@ def __init__(self, *args, **kwargs):
         Flag = get_waffle_flag_model()
         self.flag = Flag.objects.filter(name=CONTRIBUTIONS_API_ENDPOINT_ACCESS_FLAG_NAME).first()
         if not self.flag:
-            raise ApiConfigurationError()
+            raise ApiConfigurationError(
```
benjaminEwhite, line 167: See comment above at other `ApiConfigurationError` instance.

apps/api/tests/test_views.py
```diff
@@ -509,49 +501,43 @@ def test_response_when_using_noncontributor_token(self):
...
-            )
-
-        self.client.cookies["csrftoken"] = csrf._get_new_csrf_token()
-        return self.client.get(self.contributions_url, data={"rp": self.org1_rp1.slug})
-
-    def test_contributor_request_when_token_valid(self):
-        response = self._make_request()
-        self.assertEqual(response.status_code, 200)
+@pytest.mark.django_db
+@pytest.mark.usefixtures("default_feature_flags")
```
benjaminEwhite, line 505: Opportunistically moved to pytest since was touching this code anyways.

spa/src/components/contributor/ContributorVerify.jsx
```diff
@@ -11,7 +11,7 @@ import { useHistory, useLocation } from 'react-router-dom';
 // AJAX
```
nrh-cklimas, line 11: We're probably going to have merge conflicts with https://news-revenue-hub.atlassian.net/browse/DEV-4500--not sure if there is anything we can do at this point about it, though.

spa/src/components/donationPage/DonationPage.jsx
```diff
@@ -159,7 +159,7 @@ function DonationPage({ page, live = false }, ref) {
       } else if (error.response?.status === 400 && error.response?.data) {
         setErrors(error.response.data);
       } else {
-        // 403s happen if csrf token expired, or if user not authorized by api. We do want
+        // 403s happen if  user not authorized by api. We do want
```
nrh-cklimas, line 162: Just double-checking that no code change was needed here, and this is just reflecting changed(?) behavior in the backend. (I've accidentally updated a comment but not updated code myself in the past.)

spa/src/components/authentication/util.js
nrh-cklimas, line 1: Since this file is so tiny, consider converting to TypeScript and/or adding unit tests.

apps/api/authentication.py
```diff
@@ -60,8 +45,6 @@ def authenticate(self, request):
         if not validated_user and not validated_token:
             return None
 
-        enforce_csrf(request)
```
benjaminEwhite, line 63: There's no good reason for CSRF and JWT to be entangled. We use Django's off the shelf CSRF protection middleware, which gets applied to all non-DRF-involved views by default. 

Most endpoints in our API require JWT-based authentication. There are some notable places where we require an unauthed endpoint but want to guard against CSRF attacks, and in those cases, we manually decorate the relevant views to enforce CSRF (see below).

That's context for why this check here is removed.

apps/api/authentication.py
```diff
@@ -75,7 +58,7 @@ def get_user(self, validated_token):
             return super().get_user(validated_token)
 
 
-class MagicLinkTokenAuthenticationBase(JWTHttpOnlyCookieAuthentication):
+class ShortLivedMagicLinkTokenAuthentication(JWTHttpOnlyCookieAuthentication):
```
benjaminEwhite, line 61: Renamed `ShortLivedTokenAuthentication` to `ShortLivedMagicLinkTokenAuthentication` to make usage/meaning clearer, since this class is not meant for link authentication in general, but has specifically to do with authentication related to contributors and magic links.

I also removed `MagicLinkTokenAuthenticationBase` and moved its logic directly into `ShortLivedMagicLinkTokenAuthentication` which is the only code that uses that logic.

apps/api/tests/test_auth.py
```diff
@@ -1,100 +1,16 @@
...
 
-from apps.api.authentication import JWTHttpOnlyCookieAuthentication
 from apps.api.permissions import IsContributor
 from apps.contributions.tests.factories import ContributorFactory
 
 
 user_model = get_user_model()
 
 
-class JWTCookieAuthenticationTest(APITestCase):
```
benjaminEwhite, line 21: This test class is retired. We test basics of JWT-authentication all over the place in our test of api-level views.  

CSRF-related protections are narrowly tested on specific API views that require CSRF (see below).

apps/api/urls.py
```diff
@@ -23,7 +22,7 @@
     path("v1/public/", include(public_urlpatterns)),
     path("v1/token/", TokenObtainPairCookieView.as_view(), name="token-obtain-pair"),
     path("v1/contrib-email/", RequestContributorTokenEmailView.as_view(), name="contributor-token-request"),
-    path("v1/contrib-verify/", csrf_exempt(VerifyContributorTokenView.as_view()), name="contributor-verify-token"),
```
benjaminEwhite, line 26: `csrf_exempt` here was redundant here since DRF views do not enforce CSRF by default.

apps/api/views.py
```diff
@@ -104,12 +103,7 @@ def post(self, request, *args, **kwargs):
             raise exceptions.InvalidToken(e.args[0])
 
         response = Response(
-            {
-                "detail": "success",
-                "csrftoken": csrf.get_token(self.request),
```
benjaminEwhite, line 109: We no longer send this because the SPA should be using the cookie CSRF token and sending it in request header.

apps/contributions/views.py
```diff
@@ -156,14 +156,14 @@ def process_stripe_webhook(request):
     return Response(status=status.HTTP_200_OK)
 
 
-@method_decorator(csrf_protect, name="dispatch")
```
benjaminEwhite, line 159: Before, although this would enforce CSRF protection, in case of rejecting request, a HTML response would be sent back instead of JSON, whereas the SPA making the request expected JSON response.

apps/contributions/views.py
```diff
@@ -156,14 +156,14 @@ def process_stripe_webhook(request):
...
+@method_decorator(csrf_protect_json, name="dispatch")
 class PaymentViewset(mixins.CreateModelMixin, mixins.DestroyModelMixin, viewsets.GenericViewSet):
     permission_classes = []
     lookup_field = "uuid"
     queryset = Contribution.objects.all()
 
     def get_serializer_class(self):
-        if (interval := self.request.data["interval"]) == ContributionInterval.ONE_TIME:
+        if (interval := self.request.data.get("interval", None)) == ContributionInterval.ONE_TIME:
```
benjaminEwhite, line 166: Made this change because discovered when refactoring a test that if request doesn't include this value in body, it was causing an uncaught error.

apps/pages/views.py
```diff
@@ -144,7 +142,6 @@ def get_serializer_class(self):
         if self.action in ("partial_update", "create", "retrieve"):
             return serializers.DonationPageFullDetailSerializer
 
-    @method_decorator(ensure_csrf_cookie)
```
benjaminEwhite, line 147: Turns out this was the wrong place to be doing this. Moved to view that serves index.html for SPA.

spa/src/ajax/axios.js
```diff
@@ -14,13 +17,15 @@ export default Axios;
...
  * A Request interceptor.
- * first callback intercepts successfully formed requests
+ * first callback inserts the CSRF token into the request headers
  * second callback handles errors, so pass through
  */
 Axios.interceptors.request.use(
   (request) => {
-    const csrfToken = localStorage.getItem(LS_CSRF_TOKEN);
-    if (csrfToken) request.headers[CSRF_HEADER] = csrfToken;
+    const token = Cookies.get(CSRF_COOKIE);
```
benjaminEwhite, line 25: @nrh-cklimas this is updated with correct flow.

spa/src/components/donationPage/DonationPage.jsx
benjaminEwhite: See updated comment in 4c3a13ce as well as broader changes to SPA.

spa/src/components/authentication/util.js
benjaminEwhite: I'm trying to limit scope here so will pass on this (I know you said "consider" but just acknowledging and explaining).

spa/src/components/contributor/ContributorVerify.jsx
benjaminEwhite: Those came through and dealt with.

apps/contributions/tests/test_views.py
```diff
@@ -1382,17 +1382,32 @@ def test_when_called_with_unexpected_interval(self, minimally_valid_contribution
         assert response.status_code == status.HTTP_400_BAD_REQUEST
         assert response.json() == {"interval": "The provided value for interval is not permitted"}
 
-    def test_when_no_csrf(self):
-        """Show that view is inaccessible if no CSRF token is included in request.
+    def test_payment_creation_csrf_protection(self, minimally_valid_contribution_form_data, mocker):
```
benjaminEwhite, line 1385: I've broken out two separate tests around CSRF token. one for creations, one for deletion. In both cases, we prove that request fails without token and succeeds with it.

spa/src/ajax/portal-axios.ts
```diff
@@ -14,11 +15,17 @@ export default Axios;
 
 /**
  * A Request interceptor.
- * first callback intercepts successfully formed requests
+ * first callback insets CSRF token from cookie into request headersx
```
nrh-cklimas, line 18:
```suggestion
 * first callback inserts CSRF token from cookie into request headersx
```

spa/src/ajax/axios.js
nrh-cklimas, line 1: It'd be nice to convert this to TypeScript and add tests similar to portal-axios.ts. If you'd prefer to do that outside this PR, let me know and I'll add it to the story re: harmonizing js-cookie and react-cookie.

nrh-cklimas (approved):

spa/src/ajax/axios.js
benjaminEwhite: I'd prefer to handle this separately.  I created this follow up ticket: https://news-revenue-hub.atlassian.net/browse/DEV-4551

apps/api/permissions.py
```diff
@@ -148,7 +148,9 @@ def __init__(self, *args, **kwargs):
         Flag = get_waffle_flag_model()
         self.flag = Flag.objects.filter(name=MAILCHIMP_INTEGRATION_ACCESS_FLAG_NAME).first()
         if not self.flag:
-            raise ApiConfigurationError()
+            raise ApiConfigurationError(
+                f"{MAILCHIMP_INTEGRATION_ACCESS_FLAG_NAME} flag  not found in database. Check configuration."
```
njharman, line 152: super minor; double space typo after "flag  "

apps/api/views.py
```diff
@@ -200,42 +194,28 @@ def post(self, request, *args, **kwargs):
...
 
-    authentication_classes = [ShortLivedTokenAuthentication]
+    authentication_classes = [ShortLivedMagicLinkTokenAuthentication]
     permission_classes = []
 
     def post(self, request, *args, **kwargs):
         logger.info("[VerifyContributorTokenView][post] Request received for user (%s)", request.user)
         response = Response(status=status.HTTP_200_OK)
-
-        # Serializer contributor for response
```
njharman, line 217: +1

apps/common/views.py
```diff
@@ -1,4 +1,36 @@
...
+    not what we want.
+
+    This decorator allows us to continue to lean on Django's CSRF protection, but return a JSON response instead of an HTML response.
+
+    """
+
+    @wraps(view_func)
+    def wrapped_view(request, *args, **kwargs):
+        reason = CsrfViewMiddleware().process_view(request, None, (), {})
+        # If there's a reason returned, it indicates a CSRF failure
```
njharman, line 26: Looking at code, it returns None if accepted and returns settings.CSRF_FAILURE_VIEW on rejection.

I assume View()s will always be Truthy. Otoh, assuming things in security code makes me nervous. So, maybe if reason is None retrun view_func else JsonResponse?

Pretty minor,

apps/contributions/tests/test_views.py
njharman: Should we not add test "With token, but token is invalid"? Or, would that just be testing 3rd party(Django) code and none of ours?

apps/contributions/views.py
```diff
@@ -156,14 +156,17 @@ def process_stripe_webhook(request):
     return Response(status=status.HTTP_200_OK)
 
 
-@method_decorator(csrf_protect, name="dispatch")
+# This view allows for creation and cancelation (via destroy) of contributions. It is used in contribution checkout flow.
+# We need it to be unauthed, but we want a minimal level of protection against malicious actors, so we use csrf_protect_json
```
njharman, line 160: add? see https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/2702737433/RevEngine+Authentication+and+CSRF+Protection+Guide

Is it obvious must be unauthed?  Maybe worth brief comment or pointer to "confluence explanation" for future refactorer. 

I'm -1 on links to external docs vs comment. External things are so much more easily getting out of sync with code.

conftest.py
```diff
@@ -88,8 +88,17 @@
...
 @pytest.fixture
 def api_client():
-    """A DRF test API client that can be used to make API-level requests"""
-    return APIClient()
+    """A DRF test API client that can be used to make API-level requests
+
+    Note that we set `enforce_csrf_checks` to `True` to ensure that CSRF checks are enforced by default in tests.
+    This is so that our tests will reveal any issues with CSRF protection that might otherwise go unnoticed.
+
+    This is desirable because although we are using Django's default CSRF middleware, our API uses DjangoRestFramewor,
```
njharman, line 96: typo missing 'k',   "DjangoRestFramewor,"

njharman (approved):

apps/common/views.py
```diff
@@ -1,4 +1,36 @@
...
+        return (
+            JsonResponse({"detail": "CSRF token missing or incorrect."}, status=status.HTTP_403_FORBIDDEN)
```
njharman, line 28: middle ware does some logging thing on reject django/utils/log.py::log_response() """Log errors based on HttpResponse status..."""

Should we log something?  To give us warning someone is attempting concerted attack? Meh?

apps/api/permissions.py
benjaminEwhite: 9f2ebda2

apps/common/views.py
benjaminEwhite: 7aaa5580a

conftest.py
benjaminEwhite: c303f7aaed

apps/common/views.py
benjaminEwhite: Lookiing at the implementation on `logresponse` used in Django's `CsrfViewMiddleware._reject` function, I believe we should get a warning log on this automatically.

```
def _reject(self, request, reason):
        response = _get_failure_view()(request, reason=reason)
        log_response(
            'Forbidden (%s): %s', reason, request.path,
            response=response,
            request=request,
            logger=logger,
        )
        return response
```

and 

```
def log_response(message, *args, response=None, request=None, logger=request_logger, level=None, exc_info=None):
    """
    Log errors based on HttpResponse status.

    Log 5xx responses as errors and 4xx responses as warnings (unless a level
    is given as a keyword argument). The HttpResponse status_code and the
    request are passed to the logger's extra parameter.
    """
    # Check if the response has already been logged. Multiple requests to log
    # the same response can be received in some cases, e.g., when the
    # response is the result of an exception and is logged at the time the
    # exception is caught so that the exc_info can be recorded.
    if getattr(response, '_has_been_logged', False):
        return

    if level is None:
        if response.status_code >= 500:
            level = 'error'
        elif response.status_code >= 400:
            level = 'warning'
        else:
            level = 'info'
....etc
```

`._reject` isn't sending in any level arg, so get default behavior which is a warning.
```

apps/contributions/tests/test_views.py
benjaminEwhite: IMO that would be testing 3rd party.  Ideally, we'd spy on the internal use of `CsrfViewMiddleware.process_view` inside `csrf_protect_json` from this test, and it would be enough to know that we're using Django's code internally.

That said, there's not an elegant way to spy on the instance of CsrfViewMiddleware that's initialized inside `csrf_protect_json`, and I'm not keen to mock that entirely.

TLDR; i think this means we probably want a dedicated unit test for `csrf_protect_json` where we can assert that it calls CSRFViewMiddleware internally and show that it returns view when no reason, and returns 403 JSON response when there is a reason.

df64b6b2

apps/contributions/views.py
benjaminEwhite: Gave more detail in comment, and linked to confluence for further context. 79e4d64568

PR #1446, opened 2024-03-05
DEV-4500-convert-local-storage-based-auth-to-session-storage-instead

GuiMend:

n/a
#### What's this PR do?
Swaps new portal authentication from localStorage to sessionStorage so that the user "logs out" when the tab is closed
Implementing a logout to new portal
yes
no
yes
Closing the tab logs the donors out of the new portal
no
no
https://news-revenue-hub.atlassian.net/browse/DEV-4500
no
no

### DIFF
--- a/spa/src/hooks/usePortalAuth.test.tsx
+++ b/spa/src/hooks/usePortalAuth.test.tsx
@@ -2,7 +2,7 @@ import * as Sentry from '@sentry/react';
 import MockAdapter from 'axios-mock-adapter';
 import { useState } from 'react';
 import Axios from 'ajax/axios';
-import { LS_CONTRIBUTOR, LS_CSRF_TOKEN } from 'appSettings';
+import { SS_CONTRIBUTOR, LS_CSRF_TOKEN } from 'appSettings';
 import { fireEvent, render, screen, waitFor } from 'test-utils';
 import { PortalAuthContextProvider, usePortalAuthContext } from './usePortalAuth';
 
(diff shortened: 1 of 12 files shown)


spa/src/components/contributor/ContributorVerify.jsx
```diff
@@ -30,7 +30,7 @@ function ContributorVerify() {
   const handleVerifySuccess = useCallback(
     (response) => {
       if (response.status !== 200) throw new Error(`Unexpected non-failure response code: ${response.status}`);
-      localStorage.setItem(LS_CONTRIBUTOR, JSON.stringify(response.data.contributor));
+      sessionStorage.setItem(LS_CONTRIBUTOR, JSON.stringify(response.data.contributor));
```
nrh-cklimas, line 33: I think this will cause problems for the existing portal because ContributorVerify is used by both workflows. I think the best we can do right now is to set both here, but I kind of like the idea of using a different constant than `LS_CONTRIBUTOR` for the session storage key (could even be the same value, just a different constant) to make it obvious if we accidentally use the wrong key in the future.

While we're here... do you think it would expand scope a lot to convert this component to TypeScript and/or add tests? I think unit tests are more valuable if I had to pick between the two.

spa/src/utilities/isAuthenticated.ts
```diff
@@ -6,7 +6,7 @@ import { LS_CONTRIBUTOR, LS_USER } from 'appSettings';
  */
 function isAuthenticated(forContributor?: boolean): boolean {
   if (forContributor) {
-    return window.localStorage.getItem(LS_CONTRIBUTOR) !== null;
+    return window.sessionStorage.getItem(LS_CONTRIBUTOR) !== null;
```
nrh-cklimas, line 9: Either:

- Need a check here for either local or session storage, so that we keep auth working the same in the existing portal.
- We create a new component like ProtectedRoute that looks at session storage only.

spa/src/hooks/usePortalAuth.test.tsx
nrh-cklimas, line 1: We should update the test descriptions to say "session storage" instead of "local storage" to avoid confusion.

nrh-cklimas (changes requested):

nrh-cklimas (changes requested):
Changes in code look good! There are still some lingering mentions of local storage in comments that should be cleaned up in usePortalAuth.ts.

- Line 91: should rewrite the comment to explain auth changes made here to ProtectedRoute
- Line 127: "their local storage" -> "their session storage"

GuiMend:
> Changes in code look good! There are still some lingering mentions of local storage in comments that should be cleaned up in usePortalAuth.ts.
> 
> * Line 91: should rewrite the comment to explain auth changes made here to ProtectedRoute
> * Line 127: "their local storage" -> "their session storage"

@nrh-cklimas , the only mention left in `usePortalAuth` is supposed to continue as local storage.

```js
  it('saves the CSRF token to local storage', async () => {
        tree();
        expect(window.localStorage.getItem(LS_CSRF_TOKEN)).toBe(null);
```

Not sure about what you mean by on line 91: 
image

nrh-cklimas:
> @nrh-cklimas , the only mention left in `usePortalAuth` is supposed to continue as local storage.

I meant in usePortalAuth.tsx, not usePortalAuth.test.tsx.

nrh-cklimas (approved):

PR #1450, opened 2024-03-13
DEV-4477 | Add filtering by revenue program to portal contributor contributions endpoint

benjaminEwhite:

#### What's this PR do?

Adds filtering by revenue program ID for portal contributions list view: `PortalContributionFilter`

The filtering logic was getting more complicated, so moved out of view and into a separate filter class that's responsible for handling non-ordering related filtering.

Without it, new portal will return all contributions irrespective of revenue program to a contributor visiting portal.

Yes

No

Yes

N/A

No

No

- [DEV-4477](https://news-revenue-hub.atlassian.net/browse/DEV-4477)

N/A

[DEV-4477]: https://news-revenue-hub.atlassian.net/browse/DEV-4477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/tests/test_views.py
```diff
@@ -267,7 +267,7 @@ def test_list_when_expected_non_contributor_user(self, non_contributor_user, api
         api_client.force_authenticate(non_contributor_user)
         # superuser and hub admin can retrieve all:
         if non_contributor_user.is_superuser or non_contributor_user.roleassignment.role_type == Roles.HUB_ADMIN:
-            ContributionFactory.create_batch(size=2)
+            ContributionFactory.create_batch(size=2, status=ContributionStatus.PAID)
```
benjaminEwhite, line 270: Added these changes around status because they were causing test failures when factory created non-viewable statuses.

apps/contributions/views.py
```diff
@@ -598,19 +597,25 @@ def _get_contributor_and_check_permissions(self, request, contributor_id):
...
-
     def get_serializer_class(self):
         return (
             serializers.PortalContributionDetailSerializer
             if self.action == "contribution_detail"
             else serializers.PortalContributionListSerializer
         )
 
+    def filter_contributions_queryset(self, request, queryset):
+        return PortalContributionFilter().filter_queryset(request, queryset)
```
njharman, line 608: Is this part of DRF "machinery"?  I don't see this method being called anywhere in our code.

It's uncovered by unittests. So must be deletable ;)

apps/contributions/views.py
```diff
@@ -598,19 +597,25 @@ def _get_contributor_and_check_permissions(self, request, contributor_id):
...
+
+    def handle_ordering(self, request, queryset):
```
njharman, line 610: Can we not use DRF filter_backends and ordering_fields? Not work with view_sets? 

https://www.django-rest-framework.org/api-guide/filtering/#specifying-which-fields-may-be-ordered-against

apps/contributions/views.py
```diff
@@ -576,20 +575,20 @@ class PortalContributorsViewSet(viewsets.GenericViewSet):
...
-
     # Contributors should never see contributions with these statuses, or
     # interact with them (e.g. delete or patch them).
     HIDDEN_STATUSES = [
         ContributionStatus.FLAGGED,
         ContributionStatus.PROCESSING,
         ContributionStatus.REJECTED,
     ]
-
+    # NB: This view is about returning contributor.contributions and never returns contributors, but
```
njharman, line 585: This is out of scope, probably ignorable. [without looking deeply there seems a fair bit of code to work around DRF expectations. I like less code.]

Why is this not Contributions.objects.all() and instead of `contributor.contribution_set`, filter viewsets query .filter(contributor=self.._get_contributor...(request, pk). 

Could then make` queryset = Contributions.objects.exclude(status__in=HIDDEN_STATUSES)` and not have to remember to use exclude_hidden_statuses method

apps/contributions/views.py
```diff
@@ -621,22 +626,15 @@ def get_serializer_class(self):
...
-            ordering_filter.filter_queryset(request, qs, self)
-            if request.query_params.get("ordering")
-            else qs.order_by(*self.DEFAULT_ORDERING_FIELDS)
+        qs = PortalContributionFilter().filter_queryset(
+            request, self.exclude_hidden_statuses(contributor.contribution_set)
         )
+        qs = self.handle_ordering(request, qs)
+        return self.paginate_results(qs, request)
+
+    def paginate_results(self, queryset, request):
```
njharman, line 635: Don't like that request and queryset params are reversed compared to handle_ordering().  Are these overriding DRF methods which dictate parameter order? 

[edit oh, because paginator.paginate_queryset() param order?  Even if I'd make it match the order of this Viewsetclass (hide implementation detail of paginator.paginate_queryset) from this "layer".]

request is often first param in Django/DRF methods I believe.

njharman, line 635: It annoys me that this method is between the @action methods.  Would like all the @action to be right next to each other.  As they are the mental "entry points" to this view set. Other methods are in support of the them.

apps/contributions/filters.py
```diff
@@ -47,3 +49,24 @@ class Meta:
...
+            Q(donation_page__revenue_program__id=revenue_program_id)
+            | Q(contribution_metadata__contains={"revenue_program_id": revenue_program_id})
+        )
+
+    def filter_queryset(self, request, queryset):
+        filters = {}
+        for field in [x for x in self.ALLOWED_FILTER_FIELDS if x != "revenue_program"]:
+            value = request.query_params.get(field)
+            if value is not None:
+                filters[field] = value
```
njharman, line 68: Does (value := in dictionary comprension) work this way? too tricky?

filters = {x: value for x in self.ALLOWED_FILTER_FIELDS if x != "revenue_program" and (value :=  request.query_params.get(field)}

apps/contributions/filters.py
```diff
@@ -47,3 +49,24 @@ class Meta:
...
+    def filter_queryset_by_rp(self, queryset, revenue_program_id: str):
+        return queryset.filter(
+            Q(donation_page__revenue_program__id=revenue_program_id)
+            | Q(contribution_metadata__contains={"revenue_program_id": revenue_program_id})
+        )
+
+    def filter_queryset(self, request, queryset):
+        filters = {}
+        for field in [x for x in self.ALLOWED_FILTER_FIELDS if x != "revenue_program"]:
+            value = request.query_params.get(field)
```
njharman, line 66: you explicitly set default below `.get("revenue_program", None)`.  Is default not require here too?

njharman (changes requested):

apps/contributions/filters.py
benjaminEwhite: Good catch!  0d48262a98

apps/contributions/filters.py
benjaminEwhite: Think that gets a little too tricky (for me) to read, so would like to keep as is.

apps/contributions/views.py
benjaminEwhite: Yeah, missed removing this.   3b17018

apps/contributions/views.py
benjaminEwhite: 1f65c7d88

benjaminEwhite: 89cf6173e

apps/contributions/views.py
njharman: Fair enough. But I'm suggesting making initial queryset model be Contributions.  and using the reverse FK Contributions.filter(where contributor_id == contributor)

But sure out of scope.

apps/contributions/views.py
njharman: Well did the opposite, but it is consistent so ok.

njharman (approved):

x110dc (approved):

PR #1453, opened 2024-03-15
[WIP] DEV-4110 | Upgrade Django to 4.2.

benjaminEwhite:


apps/api/authentication.py
```diff
@@ -18,8 +17,7 @@ def enforce_csrf(request):
     """
     Enforce CSRF validation. From drf source, authentication.py
     """
-    check = CSRFCheck()
```
benjaminEwhite, line 21: This change was necessary because DRF's `CSRFCheck` inherits from `django.middleware.csrf.CsrfViewMiddleware` which in turn inherits from `MiddlewareMixin`, and in Django 4.2, that means it needs an argument for `get_response` param.

apps/api/tests/test_auth.py
```diff
@@ -33,7 +33,7 @@ def setUp(self):
         valid_token_obj = AccessToken().for_user(self.user)
         self.valid_jwt = jwt.encode(valid_token_obj.payload, settings.SECRET_KEY, algorithm="HS256")
         self.invalid_jwt = "notavalidhmac"
-        self.csrf_token = csrf._get_new_csrf_token()
```
benjaminEwhite, line 36: `_.get_new_csrf_token()` does not exist in Django 4.2, but we can use `._get_new_csrf_string`.

Sidenote: I don't love that we're depending on a dundered, undocumented Django method here, but addressing that is not in scope.

apps/contributions/tests/test_views.py
```diff
@@ -1,6 +1,7 @@
 import json
 from datetime import datetime, timedelta
 from unittest import mock
+from zoneinfo import ZoneInfo
```
benjaminEwhite, line 4: pytz was deprecated in Django 4.0 in favor of `ZoneInfo`.

apps/contributions/tests/test_views.py
```diff
@@ -208,9 +208,23 @@ def test_retrieve_when_expected_non_contributor_user(self, non_contributor_user,
         api_client.force_authenticate(non_contributor_user)
         new_rp = RevenueProgramFactory(organization=OrganizationFactory(name="new-org"), name="new rp")
         if non_contributor_user.is_superuser or non_contributor_user.roleassignment.role_type == Roles.HUB_ADMIN:
-            ContributionFactory(**{"one_time": True, "donation_page__revenue_program": new_rp})
-            ContributionFactory(**{"annual_subscription": True, "donation_page__revenue_program": new_rp})
-            ContributionFactory(**{"monthly_subscription": True, "donation_page__revenue_program": new_rp})
+            ContributionFactory(
+                **{"status": ContributionStatus.PAID, "one_time": True, "donation_page__revenue_program": new_rp}
```
benjaminEwhite, line 212: Changes around contributions status in factories from here down in this test are not directly related to Django 4.2 upgrade, but the ContributionFactory by default chooses randomly from available statuses. We have tests that assert that returned contribution counts, and if contributions with unviewable statuses are created (which can happen randomly), that will cause test failure.

apps/pages/signals.py
```diff
@@ -11,7 +11,7 @@
 logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
 
 
-page_published = Signal(providing_args=["instance"])
```
benjaminEwhite, line 14: This param was removed in Django 4.0.

apps/users/admin.py
```diff
@@ -1,7 +1,7 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.db.models import Q
-from django.utils.translation import ugettext_lazy as _
```
benjaminEwhite, line 4: This was removed in Django 4.0.

apps/users/models.py
```diff
@@ -69,13 +69,14 @@ def role_type(self) -> RoleTypeReturnValues:
         ra = self.get_role_assignment()
         return (ra.role_type, ra.get_role_type_display()) if ra else None
 
-    def validate_unique(self, exclude) -> None:
+    def validate_unique(self, exclude: set) -> None:
```
benjaminEwhite, line 72: In Django 4.1, it's recommended to start passing a set for `exclude` to validate_unique. https://docs.djangoproject.com/en/5.0/releases/4.1/#miscellaneous

revengine/urls.py
```diff
@@ -1,10 +1,9 @@
 import os
 
 from django.conf import settings
-from django.conf.urls import include
 from django.contrib import admin
 from django.contrib.auth import views as auth_views
-from django.urls import path, re_path
```
benjaminEwhite, line 7: This change was not, strictly speaking, necessary for upgrade. That said, in Django's documentation (apparently since 2.0), the norm is to import from django.urls not django.conf.urls.

apps/common/tests/test_views.py
```diff
@@ -21,7 +21,7 @@ def setUp(self):
         self.revenue_program = RevenueProgramFactory(name="My Test", slug="my-test")
 
     def test_page_title_includes_rev_program_name_when_subdomain(self):
-        response = self.client.get(reverse("index"), HTTP_HOST=f"{self.revenue_program.slug}.test-domain.com")
```
benjaminEwhite, line 24: This and other changes around headers stem from running django-upgrade and seeing it's recommended changes. Apparently in Django 2.2, best practice for configuring headers changed.

See https://github.com/adamchainz/django-upgrade?tab=readme-ov-file#django-22

apps/organizations/admin.py
```diff
@@ -77,14 +77,12 @@ def clean_tax_id(self):
...
     model = BenefitLevel
     verbose_name = "Benefit level"
-    verbose_name_plural = "Benefit levels"
     extra = 0
 
 
 class BenefitLevelBenefit(NoRelatedInlineAddEditAdminMixin, ReadOnlyOrgLimitedTabularInlineMixin):
     model = BenefitLevel.benefits.through
     verbose_name = "Benefit"
-    verbose_name_plural = "Benefits"
```
benjaminEwhite, line 87: In Django 4.0, InlineModelAdmin (which is a base for this) will automatically set verbose_plural_name to verbose_name + s.  See https://docs.djangoproject.com/en/4.2/releases/4.0/#django-contrib-admin

apps/api/authentication.py
```diff
@@ -18,8 +17,7 @@ def enforce_csrf(request):
...
-    check.process_request(request)
```
benjaminEwhite, line 22: it's been a while, and I can't recall reasoning behind removing this. will need to run down if removing this makes sense, or if can stay.

PR #1460, opened 2024-03-21
DEV-4484: Add canceled_at property to Contribution

GuiMend:

n/a
#### What's this PR do?
Adds "canceled_at" property to Contribution so that we can display it in the portal when the cancelation was made.
Adding additional contribution information for users
yes
no
yes
n/a
no
no
https://news-revenue-hub.atlassian.net/browse/DEV-4484
no
https://news-revenue-hub.atlassian.net/browse/DEV-4451


apps/contributions/tests/test_models.py
```diff
@@ -1593,6 +1594,29 @@ def test_next_payment_date_when_no_subscription(self):
         assert contribution.stripe_subscription is None
         assert contribution.next_payment_date is None
 
+    def test_cancelled_at_date_when_no_subscription(self):
```
benjaminEwhite, line 1597: Should keep spelling same here as in method (i.e., canceled, not cancelled).

apps/contributions/models.py
```diff
@@ -248,6 +248,16 @@ def next_payment_date(self) -> datetime.datetime | None:
         next_date = self.stripe_subscription.current_period_end
         return datetime.datetime.fromtimestamp(next_date, tz=ZoneInfo("UTC")) if next_date else None
 
+    @property
+    def canceled_at(self) -> datetime.datetime | None:
+        if self.status != ContributionStatus.CANCELED:
+            return None
```
benjaminEwhite, line 254: I think maybe we shouldn't check for the NRE contribution status here. It has the possibility of being out of sync with what's in Stripe, and Stripe is ultimate source of truth for cancellation date.  So even if our status is not canceled, if Stripe has a cancelation date, I would think we'd want to show it.

apps/contributions/serializers.py
```diff
@@ -844,6 +844,7 @@ def get_payment_type(self, instance):
     "is_modifiable",
     "last_payment_date",
     "next_payment_date",
+    "canceled_at",
```
benjaminEwhite, line 847: AC call for this being on detail view, not list view, so think this should be added to `PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS` below.

apps/contributions/tests/test_models.py
```diff
@@ -1593,6 +1594,29 @@ def test_next_payment_date_when_no_subscription(self):
...
+        contribution.provider_subscription_id = "something"
+        contribution.status = ContributionStatus.CANCELED
+        contribution.save()
+        (provider := contribution.donation_page.revenue_program.payment_provider).stripe_account_id = "else"
+        provider.save()
+        monkeypatch.setattr("stripe.Subscription.retrieve", lambda *args, **kwargs: return_value)
+        assert contribution.stripe_subscription == return_value
+        assert contribution.canceled_at == datetime.datetime.fromtimestamp(
+            return_value["canceled_at"], tz=ZoneInfo("UTC")
+        )
```
benjaminEwhite, line 1618: No need to use `monkeypatch`.  Also, can re-use an existing `subscription_factory` fixture from conftest.py

```suggestion
    def test_cancelled_at_date_when_subscription(self, contribution, subscription_factory, mocker):
        contribution.provider_subscription_id = "something"
        contribution.status = ContributionStatus.CANCELED
        contribution.save()
        (provider := contribution.donation_page.revenue_program.payment_provider).stripe_account_id = "else"
        provider.save()
        mocker.patch(
            "stripe.Subscription.retrieve",
            return_value=(sub := subscription_factory.get(canceled_at=datetime.datetime.now().timestamp())),
        )
        assert contribution.stripe_subscription == sub
        assert contribution.canceled_at == datetime.datetime.fromtimestamp(sub.canceled_at, tz=ZoneInfo("UTC"))
```

apps/contributions/models.py
```diff
@@ -248,6 +248,16 @@ def next_payment_date(self) -> datetime.datetime | None:
...
+        if not self.stripe_subscription:
+            logger.warning("Expected a retrievable stripe subscription on contribution %s but none was found", self.id)
+            return None
+        canceled_at = self.stripe_subscription["canceled_at"]
```
benjaminEwhite, line 258: Stripe objects are weird in that they allow for both dict and dot access, but I'm trying to consistently use dot access throughout revengine, so let's change this to `self.stripe_subscription.canceled_at`

apps/contributions/models.py
benjaminEwhite: Also, looks like this path is untested. Non-issue if this check gets removed, but if decide not to remove, then would want to test.

benjaminEwhite (dismissed):

njharman (approved):

PR #1476, opened 2024-04-04
DEV-3870: Add contact phone to RP model

GuiMend:

n/a
#### What's this PR do?
Adds contact_phone to Revenue Program
Add phone number to RP so that we can display it for contributors
yes
no
yes
n/a
@benjaminEwhite , I added a file `.talismanrc` to ignore the issue with `phonenumbers` library when trying to commit. Please leet me know if there is a better way to handle this issue

no
https://news-revenue-hub.atlassian.net/browse/DEV-3870
no
https://news-revenue-hub.atlassian.net/browse/DEV-3871
https://news-revenue-hub.atlassian.net/browse/DEV-3872


apps/organizations/admin.py
```diff
@@ -73,6 +74,16 @@ def clean_tax_id(self):
...
         return self.cleaned_data["tax_id"]
 
+    def clean_contact_phone(self):
+        contact_phone = self.cleaned_data["contact_phone"]
+        try:
+            phone_number = phonenumbers.parse(contact_phone)
+            if not phonenumbers.is_valid_number(phone_number):
+                raise ValidationError(f"Invalid phone number: {contact_phone}")
+        except phonenumbers.phonenumberutil.NumberParseException:
+            raise ValidationError(f"Phone not parsable: {contact_phone}")
```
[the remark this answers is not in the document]
benjaminEwhite: Those lines are not tested, so eve more reason to not duplicate the code here.

apps/organizations/validators.py
```diff
@@ -11,3 +13,12 @@ def validate_statement_descriptor_suffix(value):
 
     if not test_at_least_one_char:
         raise ValidationError("There must be at least one letter in the descriptor.")
+
+
+def validate_contact_phone_number(value):
+    try:
+        phone_number = phonenumbers.parse(value)
+        if not phonenumbers.is_valid_number(phone_number):
+            raise ValidationError(f"Invalid phone number: {value}")
```
benjaminEwhite, line 22: These lines don't get touched in tests, so should add test for this scenario.

.talismanrc
benjaminEwhite, line 1: When I run pre-commit locally after having removed this file, I don't get talisman errors.  I got version of 8.13.34 of that library, whereas this PR is committing 8.13.33, so I'm guessing that if you update phonenumbers, that may solve the problem and then we won't need to add this file.

benjaminEwhite (changes requested):
Left some suggestions and requests for changes.

Also, I would like to see the `TestRevenueProgramViewSet.test_patch_when_have_access` test case updated to include patching phone number (via an parametrized fixture in `patch_test_case`.

apps/organizations/tests/test_validators.py
```diff
@@ -38,3 +41,37 @@ def test_good_characters(good_string):
...
+    "something",
+]
+
+
+@pytest.mark.parametrize("valid_number", test_valid_phone_numbers)
+def test_valid_phone_numbers(valid_number):
+    try:
+        validate_contact_phone_number(valid_number)
+    except ValidationError as er:
+        assert False, f"{valid_number} raised an exception {er}"
```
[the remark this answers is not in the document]
GuiMend: It's not raising any errors. This is to catch, if `ValidationError` occurs, fail test by returning `assert False` with message "{valid_number} raised an exception {er}"

.talismanrc
GuiMend: I tried bumping up the version to 8.13.34 but still got the talisman failure. Can you try committing the deletion of this file to this PR?

.talismanrc
```diff
@@ -0,0 +1,4 @@
+fileignoreconfig:
```
benjaminEwhite, line 1: Still need to delete this file.

apps/organizations/serializers.py
```diff
@@ -227,6 +228,8 @@ class Meta:
             "tax_id",
             "fiscal_status",
             "fiscal_sponsor_name",
+            "contact_phone",
+            "contact_email",
```
benjaminEwhite, line 232: I believe we need to make a small change below.

As i understand it, if you call `.save` manually, you won't get the model-field level validation checks by default.  So that means we need to call `instance.full_clean()` before `instance.save(...` on line 243.

Would also be good to have a test of update that proves that you get a validation error when updating to an invalid phone number.

apps/organizations/tests/test_models.py
```diff
@@ -567,6 +567,36 @@ def test_slug_validated_against_denylist(self, revenue_program):
...
+            "123",
+            "xyz123",
+            "12345678901",
+            "something",
+        ],
+    )
+    def test_contact_phone_validation_invalid_phone(self, revenue_program, invalid_number):
+        revenue_program.contact_phone = invalid_number
+        with pytest.raises(ValidationError):
+            revenue_program.clean_fields()
```
benjaminEwhite, line 582: I think we should do `.full_clean` instead of `.clean_fields` because we're more likely to manually call full_clean in our code base. It calls `.clean_fields` internally, I believe.

Ditto below at line 598.

benjaminEwhite (changes requested):
A few more small requests.

apps/organizations/tests/test_validators.py
```diff
@@ -38,3 +41,47 @@ def test_good_characters(good_string):
...
+)
+def invalid_phone_number(request):
+    return request.param
+
+
+def test_valid_phone_numbers(valid_phone_number):
+    try:
+        validate_contact_phone_number(valid_phone_number)
+    except ValidationError as er:
+        assert False, f"{valid_phone_number} raised an exception {er}"
```
benjaminEwhite, line 82: No need to explicitly add this in the test. If a valid phone number were to trigger a validation error, the test would fail anyways. Can remove try/except here.

.talismanrc
benjaminEwhite: c4f40cfe32e

apps/organizations/serializers.py
GuiMend: From my testing it does check for the validation. When I run the patch in Postman I get the validation errors:
image

And I also added tests for this to make sure the validation error is being returned:
https://github.com/newsrevenuehub/rev-engine/pull/1476/commits/f7b0d6278ada80b994014c22a3e63db0a6866388#diff-2e94684a518366d0afa11fdb95b8daee995519812b209f76498076df07727e46R778-R799

apps/organizations/tests/test_models.py
GuiMend: I used `clean_fields` because it's the pattern I got from other parts of the code. Example:

```python
# Line 561
def test_slug_validated_against_denylist(self, revenue_program):
        denied_word = DenyListWordFactory()
        revenue_program.slug = denied_word.word
        with pytest.raises(ValidationError) as validation_error:
            revenue_program.clean_fields()
```

Should I update all places to use `full_clean`,  leave them all as `clean_fields` or just update the new tests to `full_clean`?

benjaminEwhite: Fine to leave as is. Key thing is that the code gets run in a test.

benjaminEwhite (approved):
Approved (narrowly for backend changes).

nrh-cklimas (approved):

PR #1480, opened 2024-04-08
DEV-4638 | reimplemented to use less memory

benjaminEwhite:

#### What's this PR do?

This PR refactors the Stripe import code so we do not use an excessive amount of memory. It does this by collecting required data and upserting transactions one a time instead of retrieving all required data, storing in memory, and then upserting all transactions at the end.

Some specific changes:

- Adds `apps.common.utils.stripe_call_with_backoff` which can be used to make Stripe API calls that will retry with exponential backoff if Stripe returns a rate limit error. I encountered a Stripe rate limit error one time when running this code in the review app, which prompted this change.
- No longer "unravels" the iterable returned by .`auto_paging_iter` in place in methods that use `.list` methods from Stripe. This means that we don't retrieve all list items up front.
- Removes `StripeTransactionsImporter.get_stripe_event` and related test because this method was not actually used anywhere.

Make stripe transactions data importer usable, and thus unblock new portal.

Yes.

No

Yes

N/A

No.

No

[DEV-4638](https://news-revenue-hub.atlassian.net/browse/DEV-4638)

No

No

[DEV-4638]: https://news-revenue-hub.atlassian.net/browse/DEV-4638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/contributions/stripe_import.py
+++ b/apps/contributions/stripe_import.py
@@ -35,9 +35,6 @@ from apps.pages.models import DonationPage
 
 
 MAX_STRIPE_RESPONSE_LIMIT = 100
-# we only will use half of available because we don't want running this command to cause rate limiting
-# for other stripe operations happening elsewhere in the system
-MAX_STRIPE_API_CALLS_PER_SECOND = settings.STRIPE_API_MAX_READ_CALLS_PER_SECOND / 2
 
 logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
 
@@ -235,6 +232,15 @@ class ContributionImportBaseClass(ABC):
                     case _:
                         pass
 
+    def update_contribution_stats(self, action: str, contribution: Contribution | None) -> None:
+        match action:
+            case "created":
+                self.created_contribution_ids.add(contribution.id)
+            case "updated":
+                self.updated_contribution_ids.add(contribution.id)
+            case _:
+                pass
+
 
 @dataclass
 class PaymentIntentForOneTimeContribution(ContributionImportBaseClass):
@@ -355,13 +361,7 @@ class PaymentIntentForOneTimeContribution(ContributionImportBaseClass):
             raise InvalidStripeTransactionDataError(
                 f"Contribution {contribution.id} has no donation page associated with it"
             )
-        match action:
-            case "created":
-                self.created_contribution_ids.add(contribution.id)
-            case "updated":
-                self.updated_contribution_ids.add(contribution.id)
-            case _:
-                pass
+        self.update_contribution_stats(action, contribution)
         # Even though we only expect one successful charge, it's possible to have multiple refunds
         charge = self.successful_charge
         self.upsert_payments(
@@ -476,13 +476,7 @@ class SubscriptionForRecurringContribution(ContributionImportBaseClass):
             raise InvalidStripeTransactionDataError(
                 f"Contribution {contribution.id} has no donation page associated with it"
             )
-        match action:
-            case "created":
-                self.created_contribution_ids.add(contribution.id)
-            case "updated":
-                self.updated_contribution_ids.add(contribution.id)
-            case _:
-                pass
+        self.update_contribution_stats(action, contribution)
         self.upsert_payments(charges=self.charges, refunds=self.refunds, contribution=contribution)
 
 
--- a/apps/contributions/stripe_import.py
+++ b/apps/contributions/stripe_import.py
@@ -217,6 +217,12 @@ class ContributionImportBaseClass(ABC):
         self, charges: list[stripe.Charge], refunds: list[stripe.Refund], contribution: Contribution
     ) -> None:
         for x, is_refund in list(map(lambda y: (y, False), charges)) + list(map(lambda y: (y, True), refunds)):
+            if not x or not getattr(x, "balance_transaction", None):
+                logger.warning(
+                    "Data associated with contribution %s has no balance transaction associated with it. No payment will be created.",
+                    contribution.id,
+                )
+                continue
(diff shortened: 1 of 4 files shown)


apps/contributions/stripe_import.py
```diff
@@ -89,6 +90,11 @@ class ContributionImportBaseClass(ABC):
 
     def __post_init__(self) -> None:
         self.validate()
+        self.created_contribution_ids = set()
+        self.updated_contribution_ids = set()
+        self.created_payment_ids = set()
+        self.updated_payment_ids = set()
+        self.created_contributor_ids = set()
```
benjaminEwhite, line 97: Started tracking state about what is updated because made easier to keep the logs we had before while making changes to implementation code.

apps/contributions/stripe_import.py
```diff
@@ -100,6 +106,13 @@ def validate(self) -> None:
         self.validate_customer()
         self.validate_email_id()
 
+    def get_effects_string(self) -> str:
```
benjaminEwhite, line 109: This used to exist as inline code that logged effects of running command. I've added some more detail about number of payments and contributors as well.

apps/contributions/stripe_import.py
```diff
@@ -168,6 +181,12 @@ def validate_email_id(self) -> None:
 
     def validate_metadata(self) -> None:
         """Validate the metadata associated with the stripe entity"""
+        if (
+            schema_version := self.stripe_entity.metadata.get("schema_version", None)
+        ) not in STRIPE_PAYMENT_METADATA_SCHEMA_VERSIONS:
+            raise InvalidMetadataError(
+                f"Invalid schema version {schema_version} for {self.entity_name} {self.entity_id}"
```
benjaminEwhite, line 188: This check existed elsewhere in this module, and probably should have lived here the whole time.

apps/contributions/stripe_import.py
```diff
@@ -184,6 +203,44 @@ def conditionally_update_contribution_donation_page(self, contribution: Contribu
                 reversion.set_comment("Donation page was updated from Stripe metadata")
         return contribution
 
+    def get_or_create_contributor(self) -> Contributor:
```
benjaminEwhite, line 206: We were already get-or-creating contributors already, but it was convenient to move into self-contained method for ease of testing and to DRY up code that reports on effects of running import.

apps/contributions/stripe_import.py
```diff
@@ -184,6 +203,44 @@ def conditionally_update_contribution_donation_page(self, contribution: Contribu
...
+        """Get or create a contributor"""
+        logger.debug("Upserting contributor for %s %s", self.entity_name, self.entity_id)
+        contributor, created = Contributor.objects.get_or_create(email=self.email_id)
+        if created:
+            logger.info("Created new contributor %s for %s %s", contributor.id, self.entity_name, self.entity_id)
+            self.created_contributor_ids.add(contributor.id)
+        return contributor
+
+    def upsert_payments(
```
benjaminEwhite, line 215: We were doing all of this work before, but moved into a separate method in base class to dry things up and facilitate updating stats about which entities created/updated.

apps/contributions/tasks.py
```diff
@@ -209,7 +209,7 @@ def process_stripe_webhook_task(self, raw_event_data: dict) -> None:
         logger.info("Could not find contribution. Here's the event data: %s", event, exc_info=True)
 
 
-@shared_task(bind=True, autoretry_for=(RateLimitError,), retry_backoff=True, retry_kwargs={"max_retries": 3})
```
benjaminEwhite, line 212: we no longer want to retry for rate limit error here because the code in `stripe_import.py` itself has exponential back off for rate limits, and if a rate limit error makes its way to this task, that means that after 5 retries in lower level code, we're still getting rate limit errors.

x110dc: But there could be other non-ratelimit exceptions, no?

apps/contributions/stripe_import.py
```diff
@@ -614,51 +646,17 @@ def upsert_one_time_contribution(self, data: dict) -> Tuple[Contribution, str]:
...
-            created_count,
-            updated_count,
-        )
+    def upsert_transaction(self, data: Dict[str, Any]) -> None:
+        handler = (
+            PaymentIntentForOneTimeContribution if "payment_intent" in data else SubscriptionForRecurringContribution
+        )(**data)
+        handler.upsert()
+        logger.info(handler.get_effects_string())
+        self.created_contribution_ids |= handler.created_contribution_ids
```
x110dc, line 655: Probably just me/style but it would be more intuitive to me to use `.union()` as I think of the pipe operator as bitwise.

x110dc (approved):

apps/contributions/stripe_import.py
benjaminEwhite: Yeah, I'm with you on that. Co-pilot was overzealous in offering tab completion here and I went with it. I'm going to update.

apps/contributions/tasks.py
benjaminEwhite: There could be other exceptions, but the only thing we were retrying for here was RateLimitError. If we ever have a need to retry for other exceptions, can add back retry_backoff and retry_kwargs.

apps/contributions/stripe_import.py
benjaminEwhite: Actually, TIL that this is overloaded (`|`) and is in Python docs about sets as common usage: https://docs.python.org/3/tutorial/datastructures.html#sets

PR #1482, opened 2024-04-09
DEV-4644 | fix bug around payment creation, remove unused method

benjaminEwhite:

#### What's this PR do?

Fixes a bug in import stripe transactions code that caused payments to not be created or updated.

Specifically, the bug stemmed from attempting to iterate results of a generator (charges returned by using stripe.Charge.list.auto_paging_iter) more than once. On the second time through the generator was empty, and that second run through was what fed the charges that are used to attempt to create payments.

Additionally, this PR removes a method that was no longer used that I missed removing in DEV-4638.

Makes import transactions data command usable. Unblocks portal.

Yes

No

Yes: current test updated to have mock return a generator instead of a list, to better capture actual implementation

### DIFF
--- a/apps/contributions/tests/test_stripe_import.py
+++ b/apps/contributions/tests/test_stripe_import.py
@@ -831,9 +831,12 @@ class TestStripeTransactionsImporter:
     @pytest.mark.parametrize("invoice", ["invoice_with_subscription", "invoice_without_subscription"])
     def test_assemble_data_for_pi(self, payment_intent, mocker, customer, invoice, request, charge_with_refund):
         invoice = request.getfixturevalue(invoice)
+        charges = (x for x in [charge_with_refund])
+        # want this to be an generator, as that what original method returns and there was a bug caused by treating it like
+        # a list in DEV-4644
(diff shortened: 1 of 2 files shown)


apps/contributions/tests/test_stripe_import.py
```diff
@@ -831,9 +831,12 @@ def test_get_stripe_entity_methods(self, method_name, entity_name, expand_fields
     @pytest.mark.parametrize("invoice", ["invoice_with_subscription", "invoice_without_subscription"])
     def test_assemble_data_for_pi(self, payment_intent, mocker, customer, invoice, request, charge_with_refund):
         invoice = request.getfixturevalue(invoice)
+        charges = (x for x in [charge_with_refund])
+        # want this to be an generator, as that what original method returns and there was a bug caused by treating it like
```
benjaminEwhite, line 835:
```suggestion
        # want this to be a generator, as that what original method returns and there was a bug caused by treating it like
```

x110dc (approved):

PR #1491, opened 2024-04-18
DEV-4659 | reimplement stripe transactions importer to solve for OOM 

benjaminEwhite:

#### What's this PR do?

This PR refactors the stripe transactions import code so that we retrieve (almost all) resources up front by listing them from Stripe, then storing them in a Redis cache.  This should solve for OOM problems and makes the command faster than the previous version.

Here's the new approach.  We retrieve the following data in its entirety (for an account) up front, and cache in redis:

- Subscriptions
- Payment intents
- Customers
- Balance transactions
- Invoices 
- Charges
- Refunds

This gives us all the data we need to upsert transactions, save for PaymentMethod data, which we do on a per contribution basis. This is you cannot list all PaymentMethods with the Stripe API (only all with respect to a customer).

When caching payment intents and subscriptions, we exclude those that do not have a conforming metadata schema version.

We also prune down the cached data to only what we need to process, which reduces the memory footprint considerably, allowing us to use existing provision level for Redis in review apps to be able to run the command.

Additionally:

- Solves for a bug that caused payments for refunded charges created by Stripe transactions import to have negative values for `amount_refunded`.
- Removes the `--skip-one-times-with-payment` flag. This flag made sense on earlier version, but now that we are listing all resources up front and that's what takes most time, this flag made less sense. It also would have had to reimplemented fairly significantly, and I wanted to get this PR done sooner rather than later. We can circle back and add that flag back later if desirable.
- `SubscriptionForRecurringContribution` and `PaymentIntentForOneTimeContribution` are retired, while `StripeTransactionsImporter` expands to include significantly more logic.
- There is more incremental logging which should help us understand when/if the command has hung.
- Adds test that was missing for the import_stripe_transactions_data command (note we were already testing the classes used internally, but missed the command itself).
-  Solves for a missed requirement/bug in earlier versions of this command whereby we were not populating a value for `provider_payment_id` (which is a payment intent ID) for recurring contributions created by the import command. Note that this bug did not affect updated contributions.  See [DEV-4705](https://news-revenue-hub.atlassian.net/browse/DEV-4705) for more detail.

- Solve for OOM errrors
- Make import faster

Yes

No

Yes

N/A

No

No

- [DEV-4659](https://news-revenue-hub.atlassian.net/browse/DEV-4659)
- [DEV-4705](https://news-revenue-hub.atlassian.net/browse/DEV-4705)
- [DEV-4741](https://news-revenue-hub.atlassian.net/browse/DEV-4741)

I don't think so? I'm going to create an ITS ticket out of abundance of caution to confirm that redis/cache works in prod, post deploy. This shouldn't be an issue since the main change is moving config that was in deploy.py to base.py. Since the former inherits from the latter, that should still be okay. There's additional config specific to cache for stripe import, but that shouldn't effect default cache.  That said, we can check that status quo portal works after merging in main to confirm no regressions.

- [ITS-2356](https://news-revenue-hub.atlassian.net/browse/ITS-2356)

- [DEV-4705](https://news-revenue-hub.atlassian.net/browse/DEV-4705)
- [DEV-4741](https://news-revenue-hub.atlassian.net/browse/DEV-4741)

[DEV-4659]: https://news-revenue-hub.atlassian.net/browse/DEV-4659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DEV-4705]: https://news-revenue-hub.atlassian.net/browse/DEV-4705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/stripe_contributions_provider.py
```diff
@@ -299,7 +296,18 @@ def fetch_uninvoiced_subscriptions_for_contributor(self) -> list[stripe.Subscrip
 
     def get_interval_from_subscription(self, subscription: stripe.Subscription) -> ContributionInterval:
         """Gets the ContributionInterval from a stripe.Subscription object."""
-        return SubscriptionForRecurringContribution.get_interval_from_subscription(subscription)
```
benjaminEwhite, line 302: SubscriptionForRecurringContribution went away and there was no need for this logic in its module. I just pasted the logic over into this method.

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+                )
+                continue
+            balance_transaction = self.get_resource_from_cache(
+                self.make_key(entity_id=entity["balance_transaction"], entity_name="BalanceTransaction")
+            )
+            payment, action = upsert_payment_for_transaction(contribution, balance_transaction, is_refund)
+            logger.info("Payment %s for contribution %s was %s", payment.id, contribution.id, action)
+            self.update_payment_stats(action, payment)
+
+    def get_provider_payment_id_for_subscription(self, subscription: dict) -> str | None:
```
benjaminEwhite, line 668: This provides net new data. As mentioned in PR description, we should have been capturing this in earlier versions of the command, but I missed it.  There is a follow up ticket ([DEV-4705](https://news-revenue-hub.atlassian.net/browse/DEV-4705)) to add for any affected recurring contributions created by earlier versions of the stripe import command.

x110dc (changes requested):
I added a few minor comments. At this point my main concern is the CACHE_TTL. If this is working as I think it is then if the cache expires for entities before the process is complete then we'll only be processing part of the entities for that org and it may happen silently. I guess we could also have some indeterminant behavior if some entities  get evicted but related ones don't. Say some charges get evicted before their related PIs or vice versa. So, my inclination would be to (1) make the default TTL longer, (2) issue a warning at the end of the command if the runtime (it looks like you're tracking it) exceeds or nears the TTL. May also want to clear the cache at the beginning of the command so we don't risk running out of space in cache. Also, may be useful to have a simple command to clear the entire cache if something goes wrong and we want to start from a clean slate or we run out of cache.

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        """Get provider payment id for a subscription"""
+        provider_payment_id = None
+        invoice_id = subscription.get("latest_invoice", None)
+        if invoice_id:
```
njharman, line 672: `if invoice_id := ...`

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+            invoice = self.get_resource_from_cache(self.make_key(entity_id=invoice_id, entity_name="Invoice"))
+            provider_payment_id = invoice.get("payment_intent", None)
+        return provider_payment_id
```
njharman, line 675: instead of assignment, could `return invoice.get()` and cut` provider_payment_id = None `and `return provider_payment_id`

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        revenue_program = RevenueProgram.objects.filter(id=rp_id).first()
+        if not revenue_program:
+            logger.warning("No revenue program found for id %s", rp_id)
+            return None
+        slug = parse_slug_from_url(metadata["referer"])
+        return (
+            revenue_program.donationpage_set.filter(slug=slug).first()
+            if slug
+            else revenue_program.default_donation_page
+        )
```
njharman, line 749:
```
if slug :=  parse_slug_from_url(metadata["referer"]):
    return  revenue_program.donationpage_set.filter(slug=slug).first()
return  revenue_program.default_donation_page
```

Isn't that nicer, more comprehensible? If nothing else it's shorter.

Are if/else single liners much more common in JS/webdev? [Which I'm much less exposed to]. Wondering why it seems to be your "default" or "prefereded" construct.

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        """Assemble data and ultimately upsert data for a recurring contribution."""
+        logger.info("Processing transactions for recurring contributions")
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_Subscription_*{self.stripe_account_id}*"):
+            subscription = self.get_resource_from_cache(key)
+            try:
+                contribution, action = self.upsert_contribution(stripe_entity=subscription, is_one_time=False)
+            except (InvalidStripeTransactionDataError, InvalidMetadataError, InvalidIntervalError) as exc:
                 logger.info(
-                    "Skipping payment intent %s because it already has a contribution and at least one payment", pi.id
+                    "Unable to upsert subscription %s because %s %s", subscription["id"], type(exc).__name__, exc
```
njharman, line 805: Maybe is useful (and I see it commonly (at least in switchboard)) to indicate in log that processing is continuing with some like  `; skip` or `; continuing` ?

apps/contributions/stripe_import.py
```diff
@@ -692,17 +881,81 @@ def import_contributions_and_payments(self) -> None:
...
-        self.created_contribution_ids |= handler.created_contribution_ids
-        self.updated_contribution_ids |= handler.updated_contribution_ids
-        self.created_payment_ids |= handler.created_payment_ids
-        self.updated_payment_ids |= handler.updated_payment_ids
-        self.created_contributor_ids |= handler.created_contributor_ids
+    def clear_cache(self) -> None:
+        """Clear the cache"""
+        logger.info("Clearing redis cache of entries related to stripe import for account %s", self.stripe_account_id)
+        cleared = 0
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_*{self.stripe_account_id}", count=1000):
```
njharman, line 888: How did we arrive at the value of 1000?  
How many of these things are we expecting?

Maybe some comment for future devs. Looking at performance/impact on cache system [we are using redis cache for other production stuff right, it's not dedicated to this?]

apps/contributions/stripe_import.py
```diff
@@ -692,17 +881,81 @@ def import_contributions_and_payments(self) -> None:
...
+            self.redis.delete(key)
```
njharman, line 889: Was looking for something like mdelete().  It takes delete(*names).  I'm not sure how python deals with say 100 long argument list.

I'm also probably guilty of pre-mature optimization... above something to keep in mind if performance of this turns out to matter. lots of network round trips to delete.

Should maybe add elapsed time in log to this like you did in import_contributions_and_payments()

apps/contributions/stripe_import.py
```diff
@@ -692,17 +881,81 @@ def import_contributions_and_payments(self) -> None:
...
+    def convert_bytes(size: int) -> str:
+        """Convert bytes to human readable format"""
+        # Hat tip: https://stackoverflow.com/a/12912296
+        UNITS_MAPPING = [
+            (1 << 50, " PB"),
+            (1 << 40, " TB"),
+            (1 << 30, " GB"),
+            (1 << 20, " MB"),
+            (1 << 10, " KB"),
+            (1, (" byte", " bytes")),
```
njharman, line 903: I'll trust you got this right.  [I quit wanting to be C/ASM dev around age 19 cause bit shifting, powers,  thinking about bytes make my brain hurt]

apps/contributions/stripe_import.py
```diff
@@ -692,17 +881,81 @@ def import_contributions_and_payments(self) -> None:
...
+                suffix = multiple
+        return str(amount) + suffix
+
+    def get_redis_memory_usage(self) -> int:
+        """Get redis memory usage for a given stripe account in bytes"""
+        total_memory = 0
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_*{self.stripe_account_id}*"):
+            memory_usage = self.redis.memory_usage(key)
+            if memory_usage:
+                total_memory += memory_usage
```
njharman, line 924: no scan_iter(..., count=1000)?

Do we care about accuracy? Are these nested, see samples=0 https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.memory_usage

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
-    def payment_method(self) -> stripe.PaymentMethod | None:
-        """The payment method associated with the subscription. We prefer default payment method on sub if present,
+    def list_kwargs(self) -> dict:
+        """Generates a query that can supplied for "created" param when listing stripe resources"""
+        list_kwargs = {}
+        if created_query := self.created_query:
+            list_kwargs["created"] = created_query
+        return list_kwargs
+
+    def list_and_cache_payment_intents(self) -> None:
```
njharman, line 325: Each of these list methods has a test coverage branch "miss". That I do not understand.  It's a function call, I don't see branch logic?  Do you understand?  I suggest # pragma: no branch with mention of weirdness. Which would restore 100%ness of this file.


```def list_and_cache_payment_intents(self) -> None: 
        """List and cache payment intents for a given stripe account""" 
        logger.info("Listing and caching payment intents for account %s", self.stripe_account_id) 
        self.cache_stripe_resources(     # **328 ↛ exitline 328 didn't jump to the function exit**
            resources=self.list_stripe_entity((name := "PaymentIntent"), **self.list_kwargs), 
            entity_name=name, 
            exclude_fn=self.should_exclude_from_cache_because_of_metadata, 
            prune_fn=lambda x: {k: v for k, v in x.items() if k in CACHED_PAYMENT_INTENT_FIELDS}, 
        ) 
```

njharman (changes requested):

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
-        """Validate the metadata associated with the stripe entity"""
-        if (
-            schema_version := self.stripe_entity.metadata.get("schema_version", None)
-        ) not in STRIPE_PAYMENT_METADATA_SCHEMA_VERSIONS:
-            raise InvalidMetadataError(
-                f"Invalid schema version {schema_version} for {self.entity_name} {self.entity_id}"
-            )
-        # Callling this will raise a `InvalidMetadataError` if the metadata is invalid
-        cast_metadata_to_stripe_payment_metadata_schema(self.stripe_entity.metadata)
+    _STRIPE_SEARCH_FILTER_METADATA_QUERY: str = '-metadata["referer"]:null AND -metadata["schema_version"]:null'
```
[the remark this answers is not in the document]
benjaminEwhite: 1339dac58b69

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
 
-    def validate(self) -> None:
-        super().validate()
-        self.validate_charges()
-
-    def validate_charges(self) -> None:
-        """Ensure there's only one successful or pending charge.
+    @staticmethod
+    def get_interval_from_plan(plan: dict) -> ContributionInterval:
+        """Map Stripe splan interval to Revengine contribution interval"""
```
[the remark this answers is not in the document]
benjaminEwhite: 7070ae94083

apps/contributions/stripe_import.py
benjaminEwhite: Yeah, I didn't get that "miss" either. 

 a29368fe65c04

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
+    """Convenience class for batch caching stripe resources in Redis"""
 
-    def __post_init__(self) -> None:
-        self.validate()
-        self.created_contribution_ids = set()
-        self.updated_contribution_ids = set()
-        self.created_payment_ids = set()
-        self.updated_payment_ids = set()
-        self.created_contributor_ids = set()
+    pipeline: Pipeline
```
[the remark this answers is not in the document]
benjaminEwhite: So no action required here?

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        """Make a key for a given stripe resource"""
+        return f"{CACHE_KEY_PREFIX}_{entity_name}_{entity_id}_{self.stripe_account_id}"
 
-    @property
-    def created_query(self) -> dict:
-        """Generates a query that can supplied for "created" param when listing stripe resources"""
-        return {k: v for k, v in {"gte": self.from_date, "lte": self.to_date}.items() if v}
+    def cache_charges_by_payment_intent_id(self) -> None:
+        """Cache charges by payment intent id"""
+        pipeline = RedisCachePipeline(
```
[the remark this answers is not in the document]
benjaminEwhite: I went with your suggestion to take advantage of context manager, however, I didn't see a good way to provide the entity name when overriding `__enter__`.  As best I can tell, we would have to do something wonky like manually set `._entity_name` on the initialized RedisCachePipeline in calling context.  

Might be missing something here, so please look at reimplementation via a629d52975 and let me know if you see a way to improve.

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+                logger.debug("Excluding %s %s from cache because excluded by `exclude_fn`", entity_name, resource.id)
+                excluded_count += 1
+                continue
+            pipeline.put_in_cache(
+                entity_id=resource.id,
+                key=self.make_key(resource.id, entity_name),
+                entity=resource.to_dict(),
+                prune_fn=prune_fn,
+            )
+        # need to do this at end in case we have a small number of resources waiting in buffer that didn't trigger a flush
```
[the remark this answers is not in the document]
benjaminEwhite: a629d52975

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
-            revenue_program.donationpage_set.filter(slug=slug).first()
-            if slug
-            else revenue_program.default_donation_page
+        if prune_fn:
+            logger.debug("Pruning %s %s before caching", self.entity_name, entity_id)
+            entity = prune_fn(entity)
+        self.put_in_buffer(key=key, value=entity)
+
+    def put_in_buffer(self, key: str, value: dict) -> None:
+        """ "Put a stripe resource in buffer and if buffer is full, flush it"""
```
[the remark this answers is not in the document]
benjaminEwhite: Went away in a629d52975

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
-        return (
-            revenue_program.donationpage_set.filter(slug=slug).first()
-            if slug
-            else revenue_program.default_donation_page
+        if prune_fn:
+            logger.debug("Pruning %s %s before caching", self.entity_name, entity_id)
+            entity = prune_fn(entity)
+        self.put_in_buffer(key=key, value=entity)
+
+    def put_in_buffer(self, key: str, value: dict) -> None:
```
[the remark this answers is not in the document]
benjaminEwhite: a629d52

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
-        self.validate()
-        self.created_contribution_ids = set()
-        self.updated_contribution_ids = set()
-        self.created_payment_ids = set()
-        self.updated_payment_ids = set()
-        self.created_contributor_ids = set()
+    pipeline: Pipeline
+    entity_name: str
+    batch_size: int = 100
+    buffered: int = 0
```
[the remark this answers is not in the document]
benjaminEwhite: a629d529751

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
 
 
 @dataclass
-class ContributionImportBaseClass(ABC):
-    charges: list[stripe.Charge]
-    refunds: list[stripe.Refund]
-    customer: stripe.Customer | None
+class RedisCachePipeline:
```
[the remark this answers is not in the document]
benjaminEwhite: a629d529751

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
-
-    def validate_charges(self) -> None:
-        """Ensure there's only one successful or pending charge.
+    @staticmethod
+    def get_interval_from_plan(plan: dict) -> ContributionInterval:
+        """Map Stripe splan interval to Revengine contribution interval"""
+        interval = plan["interval"]
+        interval_count = plan["interval_count"]
+
+        if interval == "year" and interval_count == 1:
```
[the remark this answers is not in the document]
benjaminEwhite: Addressed in fe62f44c7ec5

Note that I will need to update test instructions to do smoke tests on recurring contributions for status quo portal with this change.  Also will need to update tests, but going to hold off on test updates until get green light on object code changes.

apps/contributions/stripe_import.py
```diff
@@ -112,205 +142,119 @@ def parse_slug_from_url(url: str) -> str | None:
...
+            return ContributionInterval.YEARLY
+        if interval == "month" and interval_count == 1:
+            return ContributionInterval.MONTHLY
+        raise InvalidIntervalError(f"Invalid interval {interval}")
```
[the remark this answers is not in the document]
benjaminEwhite: Addressed in fe62f44c7ec5

Note that I will need to update test instructions to do smoke tests on recurring contributions for status quo portal with this change.  Also will need to update tests, but going to hold off on test updates until get green light on object code changes.

apps/contributions/stripe_import.py
```diff
@@ -329,96 +273,14 @@ def status(self) -> ContributionStatus:
...
-        # This will raise an `InvalidIntervalError` if the interval is invalid
-        self.get_interval_from_subscription(self.subscription)
-
-    @property
-    def status(self) -> ContributionStatus:
+    @staticmethod
+    def get_status_for_subscription(subscription_status: str) -> ContributionStatus:
         """Map Stripe subscription status to Revengine contribution status."""
-        match self.subscription.status:
+        match subscription_status:
```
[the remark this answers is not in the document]
benjaminEwhite: 223830fb4dd

apps/contributions/stripe_import.py
benjaminEwhite: Also, I dried up list_and_cache_* methods in 9f2fefdd28815 so this issue may be a moot point (haven't refactored/rerun tests yet to determine).

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
-        self.updated_payment_ids = set()
+    def make_key(self, entity_id: str, entity_name: str) -> str:
+        """Make a key for a given stripe resource"""
+        return f"{CACHE_KEY_PREFIX}_{entity_name}_{entity_id}_{self.stripe_account_id}"
 
-    @property
-    def created_query(self) -> dict:
-        """Generates a query that can supplied for "created" param when listing stripe resources"""
-        return {k: v for k, v in {"gte": self.from_date, "lte": self.to_date}.items() if v}
+    def cache_charges_by_payment_intent_id(self) -> None:
```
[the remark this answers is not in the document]
benjaminEwhite: Didn't move to the custom pipeline, but did dry up.

7263c6799d42c

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        self.cache_refunds_by_charge_id()
+
+    def get_resource_from_cache(self, key: str) -> dict | None:
+        """Get a stripe resource from cache, loading JSON"""
+        logger.debug(
+            "Attempting to retrieve value for key %s from redis cache",
+            key,
+        )
+        cached = self.redis.get(key)
+        return json.loads(cached) if cached else None
```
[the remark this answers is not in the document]
benjaminEwhite: ac82a5f210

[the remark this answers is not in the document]
benjaminEwhite: ac82a5f210

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        invoices = self.get_invoices_for_subscription(subscription_id)
+        for invoice in invoices:
+            charge_id = invoice.get("charge", None)
+            charge = (
+                self.get_resource_from_cache(self.make_key(entity_id=charge_id, entity_name="Charge"))
+                if charge_id
+                else None
+            )
+            if charge:
+                results.append(charge)
```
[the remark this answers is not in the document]
benjaminEwhite: 527e071850

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        results = []
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_RefundByChargeId*{charge_id}*"):
+            results.append(self.get_resource_from_cache(key))
+        return results
+
+    def get_refunds_for_subscription(self, subscription_id: str) -> list[dict]:
+        """Get cached refunds, if any for a given subscription id"""
+        results = []
+        charges = self.get_charges_for_subscription(subscription_id)
+        for charge in charges:
```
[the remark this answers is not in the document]
benjaminEwhite: f0e9159273c

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        pm_id = (
+            stripe_entity.get("payment_method", None)
+            if is_one_time
+            else stripe_entity.get("default_payment_method", None)
+        )
+        if not pm_id and customer.get("invoice_settings", None):
+            pm_id = customer["invoice_settings"].get("default_payment_method", None)
+        if pm_id:
+            pm = self.get_payment_method(pm_id)
+            return pm.to_dict() if pm else None
```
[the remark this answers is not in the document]
benjaminEwhite: f190c872b62

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
-            return False
-        return True
+    def update_contribution_stats(self, action: str, contribution: Contribution | None) -> None:
+        match action:
+            case "created":
+                self.created_contribution_ids.add(contribution.id)
+            case "updated":
+                self.updated_contribution_ids.add(contribution.id)
+            case _:
+                pass
```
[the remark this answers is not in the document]
benjaminEwhite: b761d9234e9

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+        charges = []
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_ChargeByPaymentIntentId*{payment_intent_id}*"):
+            charge = self.get_resource_from_cache(key)
+            charges.append(charge)
+        return charges
+
+    def get_successful_charge_for_payment_intent(self, payment_intent_id: str) -> dict | None:
+        """Get single successful charge for a PI. If >1 successful, raises an error"""
+        charges = self.get_charges_for_payment_intent(payment_intent_id)
+        if len([x for x in charges if x["status"] == "succeeded"]) > 1:
```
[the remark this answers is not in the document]
benjaminEwhite: 8c87eb8eb208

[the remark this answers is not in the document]
benjaminEwhite: 8c87eb8eb208

apps/contributions/stripe_import.py
benjaminEwhite: bd978f396c

apps/contributions/stripe_import.py
benjaminEwhite: 1907f245

apps/contributions/stripe_import.py
benjaminEwhite: Was preferred style of a former boss early in my coding career, so that's probably why/

df23bbdaf2

apps/contributions/stripe_import.py
benjaminEwhite: 996ba6356f6

apps/contributions/stripe_import.py
benjaminEwhite: I'm relying on crowd-sourcing via stack overflow here

🤷 

That said, this is pretty low stakes.  If we don't get this exactly right, airplanes won't fall out of the sky, we'll just have this command crash (perhaps?) or an incorrect report on mem footprint.

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+                    key=self.make_key(entity_id, entity_name),
+                    entity=charge,
+                )
+        pipeline.flush()
+
+    def cache_invoices_by_subscription_id(self) -> None:
+        """Cache invoices by subscription id"""
+        logger.info("Caching invoices by subscription id")
+        pipeline = RedisCachePipeline(pipeline=self.redis.pipeline(), entity_name=(entity_name := "InvoiceBySubId"))
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_Invoice_*"):
```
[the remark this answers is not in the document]
benjaminEwhite: That won't work here because this is fuzzy matching, whereas `make_key` makes a full key.

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
+                    key=self.make_key(entity_id, entity_name),
+                    entity=invoice,
+                )
+        pipeline.flush()
+
+    def cache_refunds_by_charge_id(self) -> None:
+        """Cache refunds by charge id"""
+        logger.info("Caching refunds by charge id")
+        pipeline = RedisCachePipeline(pipeline=self.redis.pipeline(), entity_name=(entity_name := "RefundByChargeId"))
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_Refund_*"):
```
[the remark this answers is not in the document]
benjaminEwhite: This went away because DRYed up those methods, but see comment above about how won't work because fuzzy matching.

apps/contributions/stripe_import.py
benjaminEwhite: d3f35cfb

apps/contributions/stripe_import.py
benjaminEwhite: Used redis pipeline ot batch this

d3f35cfb

apps/contributions/stripe_import.py
benjaminEwhite: I'm not able to glean enough from the docs to ascertain what value would make sense to use. That said, most of the values we're storing are only one layer deep. There are some exceptiosn like subscription.metadata (2 levels) and payment_intent.metadata (2 levels) and customer.invoice_settings and subscription.items (~3 levels deep), but so far seems like we're good in review apps with 25 MB for redis (no OOM errors for redis). If that changes, we'll get Sentry errors and can adjust.

In prod, our redis allowance is orders of magnitude bigger (gigabytes) so i don't think there's a big risk here.

revengine/settings/deploy.py
```diff
@@ -49,27 +49,7 @@
 CSRF_COOKIE_SECURE = os.getenv("CSRF_COOKIE_SECURE", "True") == "True"
 SESSION_COOKIE_SECURE = os.getenv("SESSION_COOKIE_SECURE", "True") == "True"
 
-### Performance optimizations
```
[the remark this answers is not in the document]
benjaminEwhite: Review apps and standing envs already use settings.deploy.

settings.dev is only used in CI (or as fallback in wsgi.py if DJANGO_SETTINGS_MODULE not set). The fact that tests pass in CI seems to indicate that's not a problem.

local.example.py (and therefore presumably local.py for most devs) inherits from dev. Running tests locally, everything works. And from what I can tell other parts of app work locally. FWIW, I did not try the status quo portal (which definitely involves caching) locally to see if it's effected, but a.) that code will soon be gone; b.) I don't think devs test that much locally anyways and c.) the risk is pretty low. 

Ultimately, we need cache set like this locally so can run command locally.

apps/contributions/stripe_import.py
```diff
@@ -433,247 +295,574 @@ def status(self) -> ContributionStatus:
...
-    @property
-    def created_query(self) -> dict:
-        """Generates a query that can supplied for "created" param when listing stripe resources"""
-        return {k: v for k, v in {"gte": self.from_date, "lte": self.to_date}.items() if v}
+    def cache_charges_by_payment_intent_id(self) -> None:
+        """Cache charges by payment intent id"""
+        pipeline = RedisCachePipeline(
+            pipeline=self.redis.pipeline(), entity_name=(entity_name := "ChargeByPaymentIntentId")
+        )
+        for key in self.redis.scan_iter(match=f"{CACHE_KEY_PREFIX}_Charge_*"):
```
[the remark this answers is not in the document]
benjaminEwhite: bb8d2184de17

apps/contributions/stripe_import.py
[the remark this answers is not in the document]
benjaminEwhite: bb8d2184de17

apps/contributions/stripe_import.py
njharman: Using github code ai thing, I can't remember name of.  Man, I tab complete a lot of "logic" that I assume is right.

apps/contributions/stripe_import.py
njharman: fair. 

accidental apropos typo? prefereded;  something preferred in the past.

apps/contributions/stripe_import.py
```diff
@@ -111,206 +148,115 @@ def parse_slug_from_url(url: str) -> str | None:
...
-
-        This requirement may change in the future. See [DEV-4562] in JIRA for more detail.
-        """
-        referer = self.referer
-        if (not referer) or f"{(_:=tldextract.extract(referer)).domain}.{_.suffix}" != settings.DOMAIN_APEX:
-            raise InvalidStripeTransactionDataError(
-                f"{self.entity_name} {self.entity_id} has no referer associated with it"
-            )
+    def flush(self) -> None:
+        """Flush the pipeline by cachching its resources in Redis"""
```
x110dc, line 186: typo: "cachching"

x110dc (changes requested):
I noted one minor typo, but my main concern is that I don't see if/where this was addressed: https://github.com/newsrevenuehub/rev-engine/pull/1491#pullrequestreview-2036712711

apps/contributions/stripe_import.py
benjaminEwhite: 3bd813b1652

benjaminEwhite:
@x110dc I just pushed up this commit which I believe addresses almost all of the concerns you raised in the comment about TTL and reaching limits. The only part I didn't act on was this suggestion:

> May also want to clear the cache at the beginning of the command so we don't risk running out of space in cache.

We can easily add that at any point if need arises, but it would also mean we couldn't have multiple async versions of the command running at same time.

Let me know your thoughts.

revengine/settings/base.py
```diff
@@ -237,6 +237,10 @@ class CurrencyDict(TypedDict):
 # Contributions cache set to 30 minutes < CONTRIBUTOR_LONG_TOKEN_LIFETIME
 CONTRIBUTION_CACHE_TTL = timedelta(minutes=30)
 DEFAULT_CACHE = "default"
+STRIPE_TRANSACTIONS_IMPORT_CACHE = "stripe_transactions_import"
+# We set this to two hours. We expec stripe import runs to take less than 2 hours, but this gives
```
x110dc, line 241: Comment should be updated for new value here. 2 vs. 8

x110dc (approved):

x110dc:
> We can easily add that at any point if need arises, but it would also mean we couldn't have multiple async versions of the command running at same time.

Makes sense. We can also run the clear cache command first if needed.

revengine/settings/base.py
benjaminEwhite: b678669b79a

PR #1493, opened 2024-04-25
DEV-4244: Send "All contributors" MC segment, rename one-time segment

nrh-cklimas:

n/a

#### What's this PR do?

- Creates a new Mailchimp segment during integration setup, "All contributors". This has filter criteria defined in the AC of the story, e.g. that a contact has made any purchase.
- Renames the existing segment for one-time contributors from "Contributors" to "One-time contributors".
- Refactors Mailchimp-related functionality for a RP into a separate class, `RevenueProgramMailchimpClient`, which has generic methods for stores, products, and segments.

- Clients often want to email contributors regardless of type.
- Rename is to make the purpose of the segments clearer.
- Refactor is to avoid duplication of code around creating each type of Mailchimp entity. The RP model still has business logic around criteria for segment, etc., but the mechanism of communicating with Mailchimp is now in one, separate place.

Yes.

No.

Yes.

Updated docs.

I didn't rename `RevenueProgram.mailchimp_contributor_segment_id` because it felt like a petty reason to make a migration, and generally left one-time contributor property and method names alone.

It looks like we have a pre-commit check failure on an unrelated Python package.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-4244

No.

No.


apps/organizations/models.py
```diff
@@ -580,6 +580,7 @@ class RevenueProgram(IndexedTimeStampedModel):
     mailchimp_list_id = models.TextField(null=True, blank=True)
     mailchimp_contributor_segment_id = models.CharField(max_length=100, null=True, blank=True)
     mailchimp_recurring_contributor_segment_id = models.CharField(max_length=100, null=True, blank=True)
+    mailchimp_all_contributors_segment_id = models.CharField(max_length=100, null=True, blank=True)
```
benjaminEwhite, line 583: I'm on the fence about this, but here's a thought.

Now that we're at 3 segments we're tracking, this makes me think that it's only a matter of time before there's a fourth, so we might want to consider separating out MailchimpSegment into a separate model/table. 

We might further have a RevenueProgramMailchimpSegment model.

MailchimpSegment would have  a (non unique) name and we'd probably want to start storing the segment ID (which gets returned when created, but which we do not currently store)., along with the parent list ID for the segment.  We'd want a uniqueness constraint on  list_id + segment name.

Then we'd have RevenueProgramMailchimpSegment which would be have just two columns, RP id and MC segment ID (or maybe segment name instead? with a uniqueness constraint on RP id + segment name?).

This would lead to cleaner data modeling, but that said, it would require a fair amount of refactor. 

Again, I'm on the fence about this.  Might be worthwhile asking Norm to weigh in on in this question.

apps/organizations/models.py
```diff
@@ -900,6 +921,32 @@ def make_mailchimp_recurring_segment(self) -> MailchimpSegment:
         except ApiClientError as exc:
             return self.handle_mailchimp_api_client_write_error(self.mailchimp_recurring_segment_name, exc)
 
+    def make_mailchimp_all_contributors_segment(self) -> MailchimpSegment:
```
benjaminEwhite, line 924: This method also could be dried up so there's common abstracted method for each of three segments, and the existing methods for creating call this common method with right args.

apps/organizations/models.py
```diff
@@ -687,6 +688,22 @@ def mailchimp_contributor_segment(self) -> MailchimpSegment | None:
...
+        if not self.mailchimp_all_contributors_segment_id:
+            logger.debug("No mailchimp_all_contributors_segment_id on RP %s, returning None", self.id)
+            return None
+        client = self.get_mailchimp_client()
+        try:
+            response = client.lists.get_segment(self.mailchimp_list_id, self.mailchimp_all_contributors_segment_id)
+            return MailchimpSegment(**response)
+        except ApiClientError as error:
+            return self.handle_mailchimp_api_client_read_error("contributor segment", error)
+
```
benjaminEwhite, line 706: This logic is repeated 3x over now for each segment. Think we should create a method that has the abstracted logic required for retrieving this for each of the three segments. Then we can keep the cached properties, but have them call this new method.

apps/organizations/models.py
benjaminEwhite: Also, looks like this method is not tested. Should add a unit test.

apps/organizations/models.py
```diff
@@ -968,13 +1015,31 @@ def ensure_mailchimp_recurring_segment(self) -> None:
...
+            logger.info("Segment created for rp_id=[%s]", self.id)
+            self.mailchimp_all_contributors_segment_id = segment.id
+            logger.info("Saving mailchimp all contributors segment id for rp_id=[%s]", self.id)
+            with reversion.create_revision():
+                self.save(update_fields={"mailchimp_all_contributors_segment_id", "modified"})
+                reversion.set_comment("ensure_mailchimp_all_contributors_segment updated all contributors segment id")
+        else:
+            logger.info("Segment already exists for rp_id=[%s]", self.id)
+
     def ensure_mailchimp_entities(self) -> None:
```
benjaminEwhite, line 1035: I know you didn't implement original method here (i did!) but since it's being touched, I can't help but note that it's not tested.  Think we should add a unit test for this. Should be an easy one to write because we can mock each of the constituent methods called in body of this method. Think its enough just to mock and call and assert the methods are called.

benjaminEwhite (changes requested):
Made some requests for drying up and adding tests. Plus a comment that I'm genuinely on the fence about as to whether or not we should move mailchimp segment into its own model.

apps/organizations/models.py
```diff
@@ -511,6 +511,170 @@ class MailchimpSegment:
     _links: List[dict]
 
 
+class RevenueProgramMailchimpClient(MailchimpMarketing.Client):
+    """Mailchimp client configured to interact with a revenue program's integration, after it's been initially set up."""
+
+    def __init__(self, rp: "RevenueProgram"):
```
benjaminEwhite, line 517: For "forward referencing" on this type hint, I believe should have full path to the class (so `"apps.organizations.models.RevenueProgram"`).

apps/organizations/models.py
```diff
@@ -907,74 +854,96 @@ def mailchimp_integration_connected(self):
...
-            self.make_mailchimp_one_time_contribution_product()
-        else:
-            logger.info("One-time contribution product already exists for rp_id=[%s]", self.id)
-
-    def ensure_mailchimp_recurring_contribution_product(self) -> None:
-        if not self.mailchimp_recurring_contribution_product:
-            logger.info("RP with ID %s does not have a recurring contributor product. Attempting to create", self.id)
-            self.make_mailchimp_recurring_contribution_product()
+            logger.info("Creating store for RP %s", self.id)
+            RevenueProgramMailchimpClient(self).create_store()
```
benjaminEwhite, line 858: Minor, but can we do `rp=self`?

apps/organizations/models.py
```diff
@@ -907,74 +854,96 @@ def mailchimp_integration_connected(self):
...
-        # TODO: [DEV-3579] Handle edge case where segment has been defined in Mailchimp but the ID is not saved in the RP
-        if not self.mailchimp_contributor_segment:
-            logger.info(
-                "Creating %s segment for rp_id=[%s]",
-                self.mailchimp_contributor_segment_name,
-                self.id,
+            logger.info("Store already exists for RP %s", self.id)
+
+    def ensure_mailchimp_contribution_product(self, type: Literal["one_time", "recurring"]) -> None:
+        if not getattr(self, f"mailchimp_{type}_contribution_product"):
```
benjaminEwhite, line 863: Need to define default otherwise will cause error if no value is there.

```suggestion
        if not getattr(self, f"mailchimp_{type}_contribution_product", None):
```

apps/organizations/models.py
```diff
@@ -907,74 +854,96 @@ def mailchimp_integration_connected(self):
...
+            logger.info("RP %s does not have a %s contribution product. Attempting to create", self.id, type)
+            RevenueProgramMailchimpClient(self).create_mailchimp_product(
```
benjaminEwhite, line 865: try/catch for valueerror/whatever replaces value error?

apps/organizations/models.py
```diff
@@ -907,74 +854,96 @@ def mailchimp_integration_connected(self):
...
-                self.mailchimp_contributor_segment_name,
+                "Creating %s segment for RP %s",
+                type,
                 self.id,
             )
-            segment = self.make_mailchimp_recurring_segment()
-            logger.info("Segment created for rp_id=[%s]", self.id)
-            self.mailchimp_recurring_contributor_segment_id = segment.id
-            logger.info("Saving mailchimp recurring contributor segment id for rp_id=[%s]", self.id)
+            segment = RevenueProgramMailchimpClient(self).create_segment(
```
benjaminEwhite, line 881: ditto try/catch for valueerror/whatever replaces value error?

apps/organizations/models.py
```diff
@@ -1068,7 +1037,7 @@ def mailchimp_email_lists(self) -> list[MailchimpEmailList]:
                 self.id,
                 self.mailchimp_server_prefix,
             )
-            client = self.get_mailchimp_client()
+            client = RevenueProgramMailchimpClient(self)
```
benjaminEwhite, line 1040: ditto minor (`rp=self`).

benjaminEwhite (changes requested):
Made some requests for additional changes, and it looks like most recent set of changes broke a good number of existing tests.

apps/organizations/models.py
nrh-cklimas: Hmm. When I do that, I get:

`apps/organizations/models.py:517:28: F821 undefined name 'apps'`

Wondering if it's something about how we've set up the app.

apps/organizations/models.py
```diff
@@ -907,74 +848,98 @@ def mailchimp_integration_connected(self):
...
-        here is to make `setup_mailchimp_entities_for_rp_mailing_list` more easily testable by providing
-        clean, obvious points to mock out.
-        """
-        # TODO: [DEV-3579] Handle edge case where segment has been defined in Mailchimp but the ID is not saved in the RP
-        if not self.mailchimp_recurring_segment:
+    def ensure_mailchimp_contributor_segment(
+        self, segment_type: Literal["all_contributors", "contributor", "recurring_contributor"], options
+    ) -> None:
+        if not getattr(self, f"mailchimp_{segment_type}_segment", None):
+            print("Creating")
```
benjaminEwhite, line 870: Think this print statement can be removed

apps/organizations/models.py
```diff
@@ -907,74 +848,98 @@ def mailchimp_integration_connected(self):
...
+            setattr(self, f"mailchimp_{segment_type}_segment_id", segment.id)
+            logger.info("Saving Mailchimp %s segment id for RP %s", segment_type, self.id)
             with reversion.create_revision():
-                self.save(update_fields={"mailchimp_recurring_contributor_segment_id", "modified"})
-                reversion.set_comment("ensure_mailchimp_recurring_segment updated recurring contributor segment id")
+                self.save(update_fields={f"mailchimp_{segment_type}_segment_id", "modified"})
+                reversion.set_comment(f"ensure_mailchimp_segment updated {segment_type} segment id")
         else:
-            logger.info("Segment already exists for rp_id=[%s]", self.id)
+            print("already exists")
```
benjaminEwhite, line 886: Ditto removing `print`

apps/organizations/models.py
benjaminEwhite: Huh, yeah, i guess it understands where RevenueProgram comes from but not the fully qualified path 🤷

benjaminEwhite (changes requested):
A couple of small requests for changes, but overall looking good.

apps/organizations/models.py
benjaminEwhite: Looks like that commit didn't update this line.

apps/organizations/models.py
benjaminEwhite: I'm a little hazy on my intentions when originally implementing, but yes, I think that was the idea. That way if task gets retried (that calls this method), it potentially has less to do next time around.

benjaminEwhite (commented):
Left a couple additional comments looking over again.

apps/organizations/models.py
nrh-cklimas: Fixed in 622c49de07a3ea7651f87584bba23d131090dc39

apps/organizations/models.py
nrh-cklimas: Fixed in 622c49de07a3ea7651f87584bba23d131090dc39

apps/organizations/models.py
nrh-cklimas: Argh, apologies. 21a9feaa436273c7902910fb62beb2e7ff868a40

apps/organizations/models.py
```diff
@@ -511,6 +511,164 @@ class MailchimpSegment:
...
+                        {
+                            "id": product_id,
+                            "title": product_name,
+                        }
+                    ],
+                },
+            )
+            return MailchimpProduct(**response)
+        except ApiClientError as error:
+            return self._handle_write_error(product_id, error)
```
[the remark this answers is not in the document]
nrh-cklimas: Updated in d8a2580d2e11d5a7aec6d911b0b73274b476ef19

apps/organizations/models.py
```diff
@@ -511,6 +511,164 @@ class MailchimpSegment:
...
+            return self._handle_read_error("contributor segment", error)
+
+    def get_store(self) -> MailchimpStore | None:
+        """Retrieves the revenue program's Mailchimp ecommerce store, if it exists."""
+        logger.debug("Called for RP %s", self.revenue_program.id)
+        try:
+            response = self.ecommerce.get_store(self.revenue_program.mailchimp_store_id)
+            return MailchimpStore(**response)
+        except ApiClientError as exc:
+            return self._handle_read_error("store", exc)
```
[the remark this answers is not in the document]
nrh-cklimas: Updated in 30c950e66f1e233d8ad6ddf55e379969807957b3

apps/organizations/models.py
```diff
@@ -511,6 +511,164 @@ class MailchimpSegment:
...
+            return MailchimpProduct(**response)
+        except ApiClientError as error:
+            return self._handle_write_error(product_id, error)
+
+    def create_segment(self, segment_name: str, options) -> MailchimpSegment:
+        """Creates a segment of the revenue program's Mailchimp list. This list must be previously created."""
+        logger.info("Called for RP %s", self.revenue_program.id)
+        if not self.revenue_program.mailchimp_list_id:
+            logger.error("No email list ID on RP %s", self.revenue_program.id)
+            raise MailchimpIntegrationError("Mailchimp must be connected and email list ID must be set")
```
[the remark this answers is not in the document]
nrh-cklimas: Added in 4c0ac9d02981db1a57b62d215ca78d1529969e0c

apps/organizations/models.py
```diff
@@ -511,6 +511,164 @@ class MailchimpSegment:
...
+    def get_store(self) -> MailchimpStore | None:
+        """Retrieves the revenue program's Mailchimp ecommerce store, if it exists."""
+        logger.debug("Called for RP %s", self.revenue_program.id)
+        try:
+            response = self.ecommerce.get_store(self.revenue_program.mailchimp_store_id)
+            return MailchimpStore(**response)
+        except ApiClientError as exc:
+            return self._handle_read_error("store", exc)
+
+    def _handle_read_error(
```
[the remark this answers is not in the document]
nrh-cklimas: Added in 3d2d23108631216cd50229e2b331adefd91c8d86

apps/organizations/models.py
```diff
@@ -511,6 +511,164 @@ class MailchimpSegment:
...
+            case 429:
+                logger.warning("Mailchimp rate limit exceeded for RP %s, raising exception", self.id)
+                # We raise this error because we have Celery tasks that interact with Mailchimp API and
+                # in case of rate limit error we will want to retry
+                raise MailchimpRateLimitError("Mailchimp rate limit exceeded")
+            case _:
+                logger.error("Unexpected error from Mailchimp API. The error text is %s", exc.text, exc_info=exc)
+        return None
+
+    def _handle_write_error(self, entity: str, exc: ApiClientError) -> None:
```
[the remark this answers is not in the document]
nrh-cklimas: Added in 3d2d23108631216cd50229e2b331adefd91c8d86

apps/organizations/models.py
nrh-cklimas: Added in 2019725ae09b73942274acaf1092cf4192f4f92d

apps/organizations/models.py
nrh-cklimas: Added in 2019725ae09b73942274acaf1092cf4192f4f92d

apps/organizations/models.py
```diff
@@ -907,74 +858,105 @@ def mailchimp_integration_connected(self):
...
+    def ensure_mailchimp_contribution_product(self, product_type: Literal["one_time", "recurring"]) -> None:
+        if not getattr(self, f"mailchimp_{product_type}_contribution_product", None):
+            logger.info("RP %s does not have a %s contribution product. Attempting to create", self.id, product_type)
+            try:
+                RevenueProgramMailchimpClient(rp=self).create_product(
+                    getattr(self, f"mailchimp_{product_type}_contribution_product_id"),
+                    getattr(self, f"mailchimp_{product_type}_contribution_product_name"),
+                )
+            except MailchimpIntegrationError:
+                logger.exception(
```
benjaminEwhite, line 875: This path through doesn't get tested

apps/organizations/models.py
```diff
@@ -907,74 +858,105 @@ def mailchimp_integration_connected(self):
...
-                reversion.set_comment("ensure_mailchimp_contributor_segment updated contributor segment id")
+    def ensure_mailchimp_contribution_product(self, product_type: Literal["one_time", "recurring"]) -> None:
+        if not getattr(self, f"mailchimp_{product_type}_contribution_product", None):
+            logger.info("RP %s does not have a %s contribution product. Attempting to create", self.id, product_type)
+            try:
+                RevenueProgramMailchimpClient(rp=self).create_product(
+                    getattr(self, f"mailchimp_{product_type}_contribution_product_id"),
+                    getattr(self, f"mailchimp_{product_type}_contribution_product_name"),
+                )
+            except MailchimpIntegrationError:
```
benjaminEwhite, line 874: This path through doesn't get tested

apps/organizations/models.py
```diff
@@ -907,74 +858,105 @@ def mailchimp_integration_connected(self):
...
-            self.mailchimp_recurring_contributor_segment_id = segment.id
-            logger.info("Saving mailchimp recurring contributor segment id for rp_id=[%s]", self.id)
-            with reversion.create_revision():
-                self.save(update_fields={"mailchimp_recurring_contributor_segment_id", "modified"})
-                reversion.set_comment("ensure_mailchimp_recurring_segment updated recurring contributor segment id")
+            try:
+                segment = RevenueProgramMailchimpClient(rp=self).create_segment(
+                    getattr(self, f"mailchimp_{segment_type}_segment_name"), options
+                )
+            except MailchimpIntegrationError:
```
benjaminEwhite, line 894: This path is untested

benjaminEwhite (commented):
Two (small) untested exception paths, but other than that good to go.

benjaminEwhite (approved):
LGTM

apps/organizations/models.py
```diff
@@ -907,74 +858,105 @@ def mailchimp_integration_connected(self):
...
-        """Ensure that a Mailchimp segment exists for recurring contributors for the given RevenueProgram.
-
-        Note that this is intended to be run in the `ensure_mailchimp_recurring_segment` task. The indirection
-        here is to make `setup_mailchimp_entities_for_rp_mailing_list` more easily testable by providing
-        clean, obvious points to mock out.
-        """
-        # TODO: [DEV-3579] Handle edge case where segment has been defined in Mailchimp but the ID is not saved in the RP
-        if not self.mailchimp_recurring_segment:
+    def ensure_mailchimp_contributor_segment(
+        self, segment_type: Literal["all_contributors", "contributor", "recurring_contributor"], options
```
njharman, line 882: Seems worth reducing future confusion to change to mirror property values. And not to have some plurals, some not. Already making migration, adding rename to it doesn't seem big deal.

```
All contributors -> all_contributors
One-time contributors -> contributor
Recurring contributors -> recurring_contributor
```

Suggest `contributor `be changed to `one_time_contributors` and adding s to `recurring_contributors`. Or, if not practicable, a code comment/docstring be added along lines of contributor means One-time contributor

njharman: With perfect future vision (and I'd have to think through it more but knee jerk reaction is) I'd have created JSON field or hash field to store segment_type -> id, name, whatever. The takeaway

If making db fields whose names vary by value, Consider modifying schema to store variable part of name as a value.    x_type, y_type, z_type -> type {x=..., y=..., z=...}.

njharman (changes requested):

apps/organizations/models.py
```diff
@@ -511,6 +511,174 @@ class MailchimpSegment:
...
 
 
+class RevenueProgramMailchimpClient(MailchimpMarketing.Client):
+    """Mailchimp client configured to interact with a revenue program's integration, after it's been initially set up."""
+
+    def __init__(self, rp: "RevenueProgram"):
+        logger.info("Called for RP %s", rp.id)
+        if not rp.mailchimp_integration_connected:
+            logger.warning("Called for RP %s which is not connected to Mailchimp", rp.id)
+            raise MailchimpIntegrationError("Mailchimp integration not connected for this revenue program")
```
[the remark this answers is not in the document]
nrh-cklimas: Makes sense. e8f048038a1ecc17857552dc6c9bd953e937d2ca

apps/organizations/models.py
```diff
@@ -627,100 +778,42 @@ def available_mailchimp_email_lists(self) -> List[MailchimpEmailList]:
...
                 self.id,
             )
             return None
-        try:
-            client = self.get_mailchimp_client()
-            response = client.ecommerce.get_store(self.mailchimp_store_id)
-            return MailchimpStore(**response)
-        except ApiClientError as exc:
-            return self.handle_mailchimp_api_client_read_error("store", exc)
+        return RevenueProgramMailchimpClient(rp=self).get_store()
```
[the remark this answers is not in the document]
nrh-cklimas: I like this a lot, updated in ebc07dbeb58f01574b393287316893c0cd494c66. I called it `mailchimp_client` to parallel other MC-related properties.

apps/organizations/models.py
```diff
@@ -748,157 +841,15 @@ def mailchimp_recurring_contribution_product_name(self):
 
     @property
     def mailchimp_contributor_segment_name(self):
-        return "Contributors"
+        return "One-time contributors"
```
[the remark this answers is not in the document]
nrh-cklimas: I think we are safe because the name is only used for cosmetic purposes. This model looks up segments by ID, e.g. `mailchimp_all_contributors_segment`. Otherwise the name is unused, e.g. not returned by `MailchimpRevenueProgramForSpaConfiguration` or `MailchimpRevenueProgramForSwitchboard`.

It's true this won't help out RPs that have already been connected to Mailchimp, but that's a big can of worms as some orgs have renamed their MC segments on their own, so we can't reliably find an existing segment to update it--at least not easily.

apps/organizations/models.py
```diff
@@ -907,74 +858,105 @@ def mailchimp_integration_connected(self):
 
     def ensure_mailchimp_store(self) -> None:
         if not self.mailchimp_store:
-            logger.info("Creating store for rp_id=[%s]", self.id)
-            self.make_mailchimp_store()
+            logger.info("Creating store for RP %s", self.id)
```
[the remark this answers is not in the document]
nrh-cklimas: The hobgoblin of consistency makes me ask, do you think we should also do this for `ensure_mailchimp_contributor_segment`?

apps/organizations/models.py
```diff
@@ -907,74 +858,105 @@ def mailchimp_integration_connected(self):
...
+                    getattr(self, f"mailchimp_{product_type}_contribution_product_id"),
+                    getattr(self, f"mailchimp_{product_type}_contribution_product_name"),
+                )
+            except MailchimpIntegrationError:
+                logger.exception(
+                    "Couldn't create %s Mailchimp contribution product for RP %s; continuing", product_type, self.id
+                )
         else:
-            logger.info("Segment already exists for rp_id=[%s]", self.id)
+            logger.info("%s contribution product already exists for RP with ID %s", product_type, self.id)
```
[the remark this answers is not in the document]
nrh-cklimas: 👍 Updated in bf9a95a500f52fd2d5ce0eda3809b198b14964c3

apps/organizations/models.py
nrh-cklimas: Right now, switchboard expects the products to follow this naming scheme, or at least expects the API response to do that. [This test](https://github.com/newsrevenuehub/rev-engine/blob/99f04b66f47d1354ec656003f70617b3c535dfb4/apps/organizations/tests/test_serializers.py#L192) shows the fields in question. Not sure the best way to handle this would be. We could rename the field but keep the names in existing format for that serializer? Or change rev-engine and switchboard in tandem?

apps/organizations/models.py
```diff
@@ -748,157 +841,15 @@ def mailchimp_recurring_contribution_product_name(self):
 
     @property
     def mailchimp_contributor_segment_name(self):
-        return "Contributors"
+        return "One-time contributors"
 
     @property
-    def mailchimp_recurring_segment_name(self):
+    def mailchimp_recurring_contributor_segment_name(self):
         return "Recurring contributors"
```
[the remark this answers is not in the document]
nrh-cklimas: Asked Ben about this since he wrote this originally--thought was to parallel other properties like `mailchimp_recurring_contributor_segment`. But I don't have a strongly held opinion about it, one way or the other, so whichever direction makes sense to you.

apps/organizations/models.py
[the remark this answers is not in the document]
njharman: Without looking, yes!

I'm (and it may only be me) am very "pattern recognition" oriented when looking/understanding code.  The differences stick out. And if not immediately apparent why different (inherently or via comment) I am left wondering why? Bug? Lack of my understanding? It distracts and sidelines whatever task I really wanted to accomplish.

A non-small part of my reviewing process is attempting to minimize that occurrence for future maintainers.  So, generally +1 on consistency for things that are consistent,  +1 on being dramatically/obviously different things that are different, -1 muddy middle.

njharman (approved):
Approved with two responses to be read and resolved by Chris

x110dc (dismissed):
LGTM, but it looks like this is a significant refactor of the Mailchimp client. Please mention that in the PR description.

PR #1497, opened 2024-05-01
DEV-4699: Create Contributor Impact endpoint

GuiMend:

main <- [DEV-4699 <- DEV-4565](https://github.com/newsrevenuehub/rev-engine/pull/1501)
#### What's this PR do?
- Creates endpoint to return "contributor impact" (total amount donated)
- filter_by_revenue_programs added to `ContributionQuerySet`
Enable impact tracker in new Portal

yes
no
yes
n/a
no
no
https://news-revenue-hub.atlassian.net/browse/DEV-4699
no
https://news-revenue-hub.atlassian.net/browse/DEV-4565


apps/contributions/views.py
```diff
@@ -619,6 +619,34 @@ def paginate_results(self, queryset, request):
...
+        methods=["get"],
+        url_path="impact",
+        url_name="impact",
+        detail=True,
+    )
+    def contributor_impact(self, request, pk=None):
+        """Endpoint that returns the sum of all contributions for a given contributor"""
+
+        contributor = self._get_contributor_and_check_permissions(request, pk)
+        contribution_qs: QuerySet = contributor.contribution_set
```
GuiMend, line 632: Hey @benjaminEwhite , I just saw that the FE [ticket](https://news-revenue-hub.atlassian.net/browse/DEV-4565) has an AC that the BE didn't contain.

```
It doesn’t include payments that were cancelled, failed, or otherwise did not actually send money to the RP.
```

Do I need to filter out the hidden status contributions like we do in other places or does the Payment `net_amount_paid` already handles this?
```python
qs = PortalContributionFilter().filter_queryset(
            request, self.exclude_hidden_statuses(contributor.contribution_set)
        )
```

apps/contributions/views.py
```diff
@@ -619,6 +619,34 @@ def paginate_results(self, queryset, request):
         serializer = self.get_serializer(page, many=True)
         return paginator.get_paginated_response(serializer.data)
 
+    @action(
+        methods=["get"],
+        url_path="impact",
+        url_name="impact",
+        detail=True,
+    )
+    def contributor_impact(self, request, pk=None):
```
benjaminEwhite, line 628: This doesn't seem to filter by revenue program, which according to AC is what we want. Contributors are not unique to an RP, so need to filter by RP.

apps/contributions/views.py
benjaminEwhite: Yes, I think we need to do something to exclude some statuses. That said, hiding cancelled is problematic. If you have a recurring contribution that, say, has had 6 monthly payments, then you end it, the contribution status in our system will be canceled, and I think we would want the payments that happened under the now canceled subscription to show up in impact.

If a one-time contribution shows up as canceled, I am pretty sure that can only happen if the user clicks "back" on the payment form (in which case we do not want it to count towards impact).  In that case, we shouldn't end up with a payment, because those only get created when a `payment_intent.suceeded` webhook fires (which is caused by completing the payment form in case of non-flagged, or contribution being accepted in Django admin/by task in case of flagged). Note that there are two other events that can lead to payment creation in revengine: `invoice.payment_succeeded` and `charge.refunded`.


Might be worth clarifying this AC in JIRA with Rache/Sophie.

apps/contributions/views.py
benjaminEwhite: Also, I think it would be worthwhile to abstract out the logic for generating impact numbers into a method on `Contributor` model. That method would probably take `revenue_programs` as an argument, and would allow you to filter the impact by 0 or more revenue program ids.

The advantage here is that we could then narrowly unit test the logic around totaling up these numbers. The view layer can then become "thin" in that it just calls the class method on Contributor, and returns the results.

apps/contributions/tests/test_views.py
```diff
@@ -1593,6 +1593,25 @@ def portal_contributor_with_multiple_contributions_over_multiple_rps(
         )
         return contributor
 
+    def test_contributor_impact(self, portal_contributor_with_multiple_contributions, api_client):
```
benjaminEwhite, line 1629: See suggestion below about moving contributor impact calculation into a method on the `Contributor` model. Would like to see that logic narrowly unit tested with a variety of cases.

benjaminEwhite (changes requested):
Several requests for changes.

apps/contributions/tests/test_views.py
```diff
@@ -1489,7 +1489,7 @@ def one_time_contribution(self, revenue_program, portal_contributor, mocker, fak
         )
         PaymentFactory(
             contribution=contribution,
-            amount_refunded=0,
+            amount_refunded=(contribution.amount - 100) / 2,
```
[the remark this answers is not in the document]
GuiMend: I just added some random value so that I could test if the total impact was the sum of payments minus refunded. I'll remove it since, with your suggested changes, this wouldn't be needed anyways.

apps/contributions/tests/test_models.py
```diff
@@ -62,6 +166,52 @@ def test_create_magic_link(self, one_time_contribution):
...
+            total_refunded += payment.amount_refunded
+
+        assert contributor_user.get_impact() == {
+            "total": total_paid - total_refunded,
+            "total_paid": total_paid,
+            "total_refunded": total_refunded,
+        }
+
+    def test_get_impact_with_no_payments(self, contributor_user):
+        assert contributor_user.get_impact() == {"total": 0, "total_paid": 0, "total_refunded": 0}
```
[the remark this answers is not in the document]
benjaminEwhite: Although if alter test_get_impact so that the contributor has some contributions that are canceled that have no payments, I think maybe this test can go away entirely.

apps/contributions/tests/test_views.py
benjaminEwhite: Since contributor impact is being narrowly unit tested else where, I don't think this test needs to prove the values returned are correct. Instead can just show that get expected keys in response and expected status code.

benjaminEwhite (changes requested):
A few small requests (mainly having to do with tests).

benjaminEwhite (changes requested):
Looks good. Just have one unresolved question about a test.

apps/contributions/tests/test_models.py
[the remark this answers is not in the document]
GuiMend: Renamed and added a test with contribution, but no payments

benjaminEwhite (approved):

PR #1503, opened 2024-05-07
DEV-4705 | backfill provider payment id on recurring contributions that were newly create by import command before dev 4659 was shipped

benjaminEwhite:

This PR initially has DEV-4562, which it branches from, as its base. Once DEV-4562 has been merged into main, the base should be changed to main.

#### What's this PR do?

This PR adds a management command that can be used to fix a subset of imported contributions that are a.) recurring and b.) missing a value for `provider_payment_id`.  

This happened because an earlier version of the import stripe transactionss data command did not set a value for `provider_payment_id` on recurring subscriptions.

There was some pre-existing code in the `sync_payment_transaction_time` management command that was useful for `fix_recurring_contribution_missing_provider_payment_id`, so I pulled that code out into a function — `get_stripe_accounts_and_their_connection_status` — that now lives in `apps.common.utils`.

Fixes data integrity issue with some imported recurring contributions that has potential to cause problems with:

- payment creation for first payment on recurrence (this may not happen in practice, because we manually create payments for imported contributions and don't rely on webhook receiver, but still need to be solved for).
- creating payment objects for refunds and updating contribution status for refunds of first charge for recurring contribution

Yes

No

Yes

N/A

No

No

- [DEV-4705](https://news-revenue-hub.atlassian.net/browse/DEV-4705)
- [ITS-2631](https://news-revenue-hub.atlassian.net/browse/ITS-2361)
- [DEV-4746](https://news-revenue-hub.atlassian.net/browse/DEV-4746)

- [ITS-2631](https://news-revenue-hub.atlassian.net/browse/ITS-2361))
- [DEV-4746](https://news-revenue-hub.atlassian.net/browse/DEV-4746)

[DEV-4705]: https://news-revenue-hub.atlassian.net/browse/DEV-4705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/management/commands/fix_recurring_contribution_missing_provider_payment_id.py
```diff
@@ -0,0 +1,94 @@
...
+                self.style.HTTP_INFO("No recurring contributions with missing provider payment ID found, exiting")
+            )
+            return
+        account_ids = set(
+            list(
+                contributions.filter(donation_page__revenue_program__payment_provider__stripe_account_id__isnull=False)
+                .values_list("donation_page__revenue_program__payment_provider__stripe_account_id", flat=True)
+                .distinct()
+            )
+            + list(
```
njharman, line 37: It does not matter here.

[you probably know this]  It is good to get in the habit of using generator expressions/itertools when the iterable is immediately thrown away. Or may short-circuit . 

I'm 63% sure various django ORM things a lazy in that the query doesn't get run until it's needed, but once it is all the results are created (like a list). So, it's probably moot here.

Here we are creating memory for two entire lists, and creating memory for set, then tossing lists away.  A generator is lazy/"generates" items one at a time. time/memory for large lists and lots of time memory if your doing something like any(), 

- any([ x for x in seq]) creates the entire list first, then starts looking at truth. 
- any(x for x in seq), creates one item, checks truthy, creates next.

apps/contributions/management/commands/fix_recurring_contribution_missing_provider_payment_id.py
```diff
@@ -0,0 +1,94 @@
...
+            if subscription.latest_invoice and (pi_id := subscription.latest_invoice.payment_intent):
+                contribution.provider_payment_id = pi_id
+                with reversion.create_revision():
+                    contribution.save(update_fields={"modified", "provider_payment_id"})
+                    reversion.set_comment("Updated by fix_recurring_contribution_missing_provider_payment_id command")
+                updated_ids.append(contribution.id)
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"Updated {len(updated_ids)} contribution(s) out of {fixable_contributions_count} eligible contributions. "
+                f"The following contributions were updated: {', '.join(map(str, updated_ids))}"
```
njharman, line 91: Would we want to "hand fix" contributions that did not match if statement on line #82?  And thus want, also, a list of contributions that "failed" matching that if?

njharman (approved):

apps/contributions/management/commands/fix_recurring_contribution_missing_provider_payment_id.py
benjaminEwhite: Good point. ad8dea08a1

apps/contributions/management/commands/fix_recurring_contribution_missing_provider_payment_id.py
benjaminEwhite: Good point. I'll keep in mind in future. This code will be run once then retired, so think we'll be okay, but I appreciate the concern.

x110dc (approved):

PR #1507, opened 2024-05-09
DEV-4753 | PATCH endpoint for switchboard to change contribution.revenue_program

benjaminEwhite:

#### What's this PR do?

This PR adds a new endpoint for updating contributions at `/api/v1/switchboard-contributions/<contribution_id>/`.

This endpoint is locked down to only allow access to the switchboard API account.

To that end it adds a new permission `IsSwitchboardAccount`, which ensures that the requesting user is authenticated and is the switchboard account (as configured via env var).

For now, we only allow updating a contribution's revenue program (via. `._revenue_program`, not `.donation_page.revenue_program`).  If the contribution currently has `.donation_page.revenue_program` (which most will, in the wild), we set nullify that field along with setting `._revenue_program`. This is to ensure we don't violate our constraint on the contribution model.

Additionally, we validate that the rp that the user is requesting be patched belong to the same organization as the original rp of the contribution.

This will allow us to solve for a client use case where they need to update revenue program on a subset of their contributions (via switchboard).

Coming soon.

No

Yes

N/A

No

N/A

[DEV-4753](https://news-revenue-hub.atlassian.net/browse/DEV-4753)

[ITS-2369](https://news-revenue-hub.atlassian.net/browse/ITS-2369)

Resolve discussion and create any follow up coming out of subtask tracked in [DEV-4771](https://news-revenue-hub.atlassian.net/browse/DEV-4771)

[DEV-4753]: https://news-revenue-hub.atlassian.net/browse/DEV-4753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/api/permissions.py
```diff
@@ -163,3 +163,17 @@ def __init__(self, *args, **kwargs):
...
+    """
+    Permission to lock down access to the switchboard account
+    """
+
+    def has_permission(self, request, view):
+        logger.debug("Checking if user is switchboard account")
+        return (
+            request.user.is_authenticated
+            and request.user.email == settings.SWITCHBOARD_ACCOUNT_EMAIL
+            and settings.SWITCHBOARD_ACCOUNT_EMAIL
```
njharman, line 178: +3.14 
You thought of tricky case, "no email" and "unset" SWITCHBOARD_ACCOUNT_EMAIL.

I don't think I would have...

njharman: The code is covered. I was still saddened to not see a test for IsSwitchboardAccount in api/tests/test_permissions.py

apps/contributions/serializers.py
```diff
@@ -933,3 +933,51 @@ class Meta:
...
+        logger.debug("Updating contribution %s with validated data %s", instance, validated_data)
+        update_fields = {"modified"}
+        _rp_being_set = False
+        if (rp := validated_data.pop("_revenue_program", None)) and instance._revenue_program != rp:
+            instance._revenue_program = rp
+            update_fields.add("_revenue_program")
+            _rp_being_set = True
+        if instance.donation_page and _rp_being_set:
+            instance.donation_page = None
+            update_fields.add("donation_page")
```
njharman, line 966: Instead of _rp_being_set bool, consider indenting this if statement and cutting _rp_being_set

```
if (rp := validated_data.pop("_revenue_program", None)) and instance._revenue_program != rp:
    ...
    if instance.donation_page and _rp_being_set:
       ...
```

apps/contributions/serializers.py
```diff
@@ -933,3 +933,51 @@ class Meta:
...
+        with reversion.create_revision():
```
njharman, line 967: Shouldn't this be indented (so that it only runs if rp_being_set)?

As it is now, it seems, that when if (rp := ...): is false we will be creating revisions consisting only of `modified` being changed.

apps/contributions/serializers.py
```diff
@@ -933,3 +933,51 @@ class Meta:
...
+        Here, if _revenue_program is being set and instance currently has donation_page set,
+        we also set donation_page to None because of our constraint on the model.
+
+        Note that at present we only allow updpating the ._revenue_program field of a contribution. If this changes in future
+        this update override will need to be modified accordingly.
+        """
+        logger.debug("Updating contribution %s with validated data %s", instance, validated_data)
+        update_fields = {"modified"}
+        _rp_being_set = False
+        if (rp := validated_data.pop("_revenue_program", None)) and instance._revenue_program != rp:
```
njharman, line 960: [have to read the comments below first to grok this one]

Another way to handle this, avoiding indentation is to reverse if statement into a "guard clause". I don't feel strongly about it either way. An alternative if you were opposed to indentation suggested below.

```
if not ((rp := validated_data.pop("_revenue_program", None)) and instance._revenue_program != rp):
    return instance
```

Although, due to too much parens and double negation I'd probably rewrite
```
rp = validated_data.pop("_revenue_program", None)
if not rp or rp == instance_revenue_program:
    return instance
```

njharman, line 960: Tests show it works.  Just surprised the validated field is the underscore one. Guess DRF is getting that from the source="_revenue_program".

Self resolving this rubber duck moment.

apps/contributions/serializers.py
```diff
@@ -933,3 +933,51 @@ class Meta:
...
+            update_fields.add("donation_page")
+        with reversion.create_revision():
+            instance.save(update_fields=update_fields)
+            reversion.set_comment("Updated by SwitchboardContributionSerializer.update")
+        return instance
+
+    def validate_revenue_program(self, value: RevenueProgram) -> None:
+        """Ensure that the revenue program being set is from the same organization as the current revenue program
+
+        Note that the actual model attribute here is `._revenue_program`.
```
njharman, line 975: nm, everything is good. The dual source rp is really a pain in the head/tricky.

[edit I don't think I was understanding correctly]

This is "auto called" validator on revenue_program field which is actually checking _revenu_program because of the `source=` above.

~But now the question is:  what if contribution being changed is one that got rp via donation_page~

[/edit]

If I understand correctly, isn't this a bit misleading?  

Code is using, correctly, the property that resolves to either `._revenue_program` or` .dontation_page.revenue_program`.

apps/contributions/tests/test_views.py
```diff
@@ -2334,3 +2334,89 @@ def test_views_when_contribution_not_found(
...
+        contribution.refresh_from_db()
+        if request_has_revenue_program:
+            assert contribution._revenue_program == rp_2
+
+    def test_update_when_not_switchboard_user(self, api_client, other_user, contribution):
+        api_client.force_authenticate(other_user)
+        response = api_client.patch(reverse("switchboard-contribution-detail", args=(contribution.id,)))
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_update_when_patching_rp_for_different_org(self, api_client, switchboard_user, contribution, other_orgs_rp):
```
njharman, line 2412: Maybe contribution fixture already does [if not suggest we add it (so all tests that use fixture cover this)  [fixtures can return > 1 value causing multiple tests similar to parametrize, right?] ].

Do we test both "paths" to rp?  _revenue_program and donation_page.revenue program?

njharman (changes requested):

apps/api/permissions.py
benjaminEwhite: Added in 635811401f

apps/contributions/serializers.py
benjaminEwhite: 07952b080fbf

apps/contributions/serializers.py
benjaminEwhite: 07952b080fbf

apps/contributions/serializers.py
benjaminEwhite: 07952b080fbf

apps/contributions/tests/test_views.py
benjaminEwhite: 431f9ba27

x110dc (approved):
lgtm

PR #1513, opened 2024-05-14
DEV-4726: Remove custom style override on input focus

GuiMend:

n/a
#### What's this PR do?
Removes custom color override from input focus
Style colors should not affect active states or focus rings around text field
yes
no
n/a
n/a
no
n/a
https://news-revenue-hub.atlassian.net/browse/DEV-4726
no
no

### DIFF
--- a/spa/src/styles/themes.ts
+++ b/spa/src/styles/themes.ts
@@ -258,6 +258,15 @@ export function mapCustomStylesToMuiOverrides(styles: Partial<DefaultTheme> = {}
     palette: {
       primary: { main: styles?.colors?.cstm_CTAs ?? revEngineTheme.colors.primary },
       secondary: { main: styles?.colors?.cstm_CTAs ?? revEngineTheme.colors.primary }
+    },
+    overrides: {
+      MuiOutlinedInput: {
+        root: {
+          '&&.Mui-focused fieldset': {
+            borderColor: 'initial'
+          }
+        }
+      }
     }
   });
 }
--- a/spa/src/styles/AdminGlobalStyles.js
+++ b/spa/src/styles/AdminGlobalStyles.js
@@ -8,7 +8,6 @@ export default createGlobalStyle`
   *::after,
   *::before {
     box-sizing: inherit;
-    outline-color: ${(props) => props.theme.colors.primary};
   }
 
   body {
--- a/spa/src/styles/DonationPageGlobalStyles.js
+++ b/spa/src/styles/DonationPageGlobalStyles.js
@@ -8,7 +8,6 @@ export default createGlobalStyle`
   *::after,
   *::before {
     box-sizing: inherit;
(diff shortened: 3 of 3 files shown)


spa/src/styles/themes.ts
```diff
@@ -258,6 +258,15 @@ export function mapCustomStylesToMuiOverrides(styles: Partial<DefaultTheme> = {}
     palette: {
       primary: { main: styles?.colors?.cstm_CTAs ?? revEngineTheme.colors.primary },
       secondary: { main: styles?.colors?.cstm_CTAs ?? revEngineTheme.colors.primary }
+    },
+    overrides: {
```
nrh-cklimas, line 262: I'd strongly prefer we do this in `components/base/TextField/TextField.tsx` or `components/base/Select/Select.tsx` if at all possible, so that it is clearer what the scope of this style is.

nrh-cklimas (changes requested):

spa/src/components/base/Select/Select.tsx
```diff
@@ -46,9 +50,14 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>((props, ref) => {
...
       variant="outlined"
-      // As a select no InputProps are needed, so we override TextField's default to avoid console errors
-      InputProps={{}}
-      SelectProps={{ classes: { selectMenu: 'NreSelectMenu' } }}
+      // Override focused border color so that no custom style is applied
+      InputProps={{
+        classes: {
+          notchedOutline: 'NreSelectNotchedOutline',
+          focused: 'NreSelectFocused'
+        } as any
```
nrh-cklimas, line 58: Instead of using `any`, let's do this at the top:

```tsx
export type SelectProps = Omit<OutlinedTextFieldProps, 'variant'> & {
  options: Array<{ label: ReactNode; value: string | number; selectedLabel?: string | null }>;
};
```
where `OutlinedTextFieldProps` comes from `@material-ui/core`. I've had to do this with other components--issue is that we know we're using the outlined variant, but TypeScript doesn't.

spa/src/components/base/Select/Select.tsx
```diff
@@ -46,9 +50,14 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>((props, ref) => {
...
+      }}
+      SelectProps={{ classes: { selectMenu: 'NreSelectMenu', outlined: 'NreSelectRoot' } }}
```
nrh-cklimas, line 60: Do you need `NreSelectRoot`? I don't see it referenced in styles.

spa/src/components/base/Select/Select.tsx
GuiMend: Nice one. Thanks!

spa/src/components/base/Select/Select.tsx
GuiMend: Nope, forgot to remove it

PR #1516, opened 2024-05-16
DEV-4797 | fix bug in stripe transactions importer update payment stats that causes import to crash, and a couple other things

benjaminEwhite:

#### What's this PR do?

This PR:

- Solves for bug in [DEV-4797](https://news-revenue-hub.atlassian.net/browse/DEV-4797). There was logging code that incorrectly assumed that a reference argument would always have a `.id` value, but this is not a good assumption. It will handle that gracefully moving forward.
- Adds a flag to management command, `--suppress-stripe-info-logs`. When not passed to command, we'll get info level Stripe logs which will hopefully help us determine if/how we're being rate limited by Stripe. Including the flag will cause only error level stripe messages to be logged.
- Fixes a bug in log_backoff where I had made incorrect assumptions about the entries that will be on `details` arg (a dict).  My theory is that a problem in this handler was (perhaps?) causing backoff logging to silently fail, and that's why we didn't see evidence of rate limiting.

Unblocks data import, which unblocks new portal.

Yes

No

Yes

N/A

No

No

-[DEV-4797](https://news-revenue-hub.atlassian.net/browse/DEV-4797)

No

N/A

[DEV-4797]: https://news-revenue-hub.atlassian.net/browse/DEV-4797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-4797]: https://news-revenue-hub.atlassian.net/browse/DEV-4797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nrh-cklimas (approved):

x110dc (approved):

apps/contributions/stripe_import.py
```diff
@@ -90,31 +91,60 @@
...
+    - exception
+    - tries
+    - wait
+
+
+    """
+    try:
+        details = OnBackoffDetails(**details)
+    except ValidationError:
+        logger.exception("Error parsing backoff details: %s", details)
```
njharman, line 123: Not sure what we're using to analyze logs. If something like grep/search/filter it might be useful to start this log with same string as others "Backing off ". Or, other common string.

Although, I believe our log-format includes file/func which would serve same purpose... or is that switchboard that does that?

apps/contributions/management/commands/import_stripe_transactions_data.py
```diff
@@ -10,9 +9,10 @@
 from apps.organizations.models import PaymentProvider
 
 
+# Temporarily commenting out while we try to get more detail about rate limites from Stripe.
 # otherwise we get thousands and thousands of info logs from stripe and hard to find our own logs
-stripe_logger = logging.getLogger("stripe")
-stripe_logger.setLevel(logging.ERROR)
+# stripe_logger = logging.getLogger("stripe")
+# stripe_logger.setLevel(logging.ERROR)
```
njharman, line 15: Alternative option to consider (instead of commenting out); is putting code behind a feature flag. One driven by Environment var. Which would enable easier toggling (set envvar and restart vs code commit/deploy).  

I don't actually know our deployment system/process enough to gauge if that would be an advantage.

njharman (approved):

apps/contributions/management/commands/import_stripe_transactions_data.py
x110dc: I like the idea of a runtime way to opt in/out of these logs.

apps/contributions/stripe_import.py
benjaminEwhite: We should get file/func here as well.

apps/contributions/management/commands/import_stripe_transactions_data.py
benjaminEwhite: Instead of using feature flag, I just added a command option. By default we'll get info logs, and if run with `--suppress-stripe-info-logs` we'll suppress.

 b602237e2097

PR #1517, opened 2024-05-17
DEV-4761: pre Django 4 updates

nrh-cklimas:

n/a

#### What's this PR do?

This pulls the majority of work out of https://github.com/newsrevenuehub/rev-engine/pull/1453. The goal is to make all changes that aren't dependent on Django 4 here.

- Replace usage of csrf._get_new_csrf_token() with csrf._get_new_csrf_string() in tests.
- Fix Heisenbug behavior in ContributionFactory.
- Repace usage of the pytz library with zoneinfo.
- Replace path specs in apps/api/urls.py with plain path instead of re_path.
- Replace usage of short_description in admin.py files with @admin.display annotation.
- Add rule number to noqa comment in revengine.settings.deploy module.
- Revise imports in revengine/urls.py to use best practices.
- Adds robust unit tests for DenyListWord case insensitivity behavior. These pass with status quo implementation, and when we apply new collation with Django 4.2 upgrade, we'll have assurance vs. regression. Also note, we initially wanted to include that db_collation creation and migration in this PR, but it turns out there are bugs around postgres + Django + case insensitivity that are solved for in 4.2 (see https://stackoverflow.com/a/76795125)

Additionally, this PR makes a change to `heroku-postdeploy.sh`. Instead of dropping the review app db and then restoring from staging, we instead exclude `django_migrations` (and its id sequence) when dumping staging db ahead of restoring.

The reason we were dropping the db was to solve for a bug whereby review app builds would fail if on initial build there was a new migration. In that case, the restore step would fail because when it tried to restore `django_migrations` from staging, there would be migrations from PR that would not appear in staging.  

The solution of dropping db then restoring from staging dump worked fine until this PR. This PR adds a collation that is stored at db level and is not in public schema.  So when the public tables were dropped, the collation stuck around (because of the initial migration that happens in the release step of build).  Then when migrations ran in postdeploy there was an integrity error because of attempt to re-migrate the already created collation.

Derisk the Django upgrade itself.

Yes.

No.

A few test updates.

n/a

No.

No.

- https://news-revenue-hub.atlassian.net/browse/DEV-4761
- https://news-revenue-hub.atlassian.net/browse/DEV-4110

No.

Actual upgrade-- probably will be done in DEV-4110.


scripts/heroku-postdeploy.sh
```diff
@@ -15,10 +15,14 @@ set -o xtrace
 # but is already in the target db, it will remain in place, without being reflected as
 # created in `django_migrations`. Subsequent runs of the migrations will fail when
 # Django tries to create the already existing new model/table.
+#
+# DROP OWNED is to clean up DB collations.
 psql -d ${DATABASE_URL} -c "DROP SCHEMA public CASCADE;
+DROP OWNED by current_user CASCADE;
```
nrh-cklimas, line 21: This resolved the deploy issue we had been seeing where it complained about a collation with the same name already existing. I don't believe there are any downsides to adding this but I can't say I'm an expert.

benjaminEwhite: @nrh-cklimas this seems like an okay workaround, but I'm not crazy about modifying this script without understanding why this bug is happening in the first place. 

Think I'd like to defer to @x110dc on best way forward here.  Daniel for additional context, without this line added to post deploy, the build for review app fails because when migrations run, a collation called `case_insensitive` is already found, and get error when trying to create new collation with same name.

I poked around in staging and discovered that we had a collation with that name (probably from testing changes in staging). I deleted the collation and migration records from staging, which I would have thought would have prevented the integrity error, but turns out we still got it even after deleting case_insensitive from staging.

benjaminEwhite (changes requested):
code changes look good to me. I’m basically :thumbsup: save for this oddity around the integrity error around the collation. Ideally, I’d like us to understand root cause instead of changing post deploy script as workaround. That said, I’d also like to unblock this work, so on the fence. I’d like to get Daniel’s feedback on that conversation thread.

 

Other than that, only thing to note is I think it would be good to have a bit more detail in the “What’s this PR do?” section.  Would be good to list the main changes (inclusive of the changes to factory code setting status to paid, which IIRC wasn’t strictly necessary for upgrade, but kills some intermittent test failures that were creating noise when trying to understand if there were regressions with upgrade).

benjaminEwhite (approved):

benjaminEwhite (approved):

apps/config/tests/test_models.py
```diff
@@ -1,7 +1,19 @@
...
+    @pytest.fixture
+    def word(self):
+        return DenyListWord(word=self.text)
+
+    def test_basics(self, word):
+        assert str(word) == self.text
+
+    def test_case_sensitive(self, word):
+        assert word.word != (new_text := self.text.capitalize())
+        DenyListWord(word=new_text).save()
```
njharman, line 19: Man, I find this test massively confusing. I don't know what it is suppose to be testing. It doesn't appear to be testing anything other than if a == b (test_basics), then I change b, a != b. 😕  

First, do I understand collation correctly?  It only has to do with sorting, right https://www.postgresql.org/docs/current/collation.html

1. Why do we need collation at all?  I don't think we are concerned with denylist sort order. We lower() in validator (only thing we need to do). 
def validate_slug_against_denylist(value):
```
    denylist_words = DenyListWord.objects.all().values_list(Lower("word"), flat=True)
    if value.lower() in denylist_words:
```
2. In this test; Why create new fixture and not use DenyListWordFactory (which should be updated to have words of mixed capitalization)
3. Why even have this test?

njharman (changes requested):

apps/config/tests/test_models.py
benjaminEwhite: yeah, this was lazy/trivial testing. added substantive tests in 9dc780bd3a8b2

njharman (approved):

x110dc (approved):

PR #1519, opened 2024-05-21
Dev 4791 create management command to backfill provider payment method details and move that logic out of stripe transactions data import management command

benjaminEwhite:

#### What's this PR do?

- Modifies the Stripe import transactions data code such that it now defaults to not retrieving and upserting provider_payment_method_details (though it's possible to make it do that by including a flag when calling command `--retrieve-payment-method`).  We are doing this to reduce the number of API calls made by running the command.
- Removes override of `Contribution.save()` method which was used to fetch and save provider_payment_method_details if value of `provider_payment_method_id` was changing (and truthy). This was necessary so that we could reduce API calls (if we didn't make this change, even if we changed import transactions code, custom save override would have caused API call to Stripe).
- Removes a bunch of places in test code where we patched `Contribution.fetch_stripe_payment_method` to avoid API calls being triggered in tests.
- Adds a new management command `sync_missing_provider_payment_method_details` which will find Contributions that have a null value for `provider_payment_method_details` and a truthy value for `provider_payment_method_id` and tries to retrieve the PM then save back its value to the contribution.
- Updates the webhook handlers `handle_payment_intent_succeeded`, `handle_subscription_updated`, and `handle_invoice_payment_succeeded` such that they retrieve payment method and include `provider_payment_method_details` if there is a pm id in event data (this is to compensate for the removal of the custom save override behavior).

- Our hope is this will solve for long run times for Stripe transactions import for large accounts.
- Removes tech debt around custom save

yes

No

Yes

N/A

No

No

[DEV-4791](https://news-revenue-hub.atlassian.net/browse/DEV-4791)

No

No

[DEV-4791]: https://news-revenue-hub.atlassian.net/browse/DEV-4791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/tests/test_webhooks.py
```diff
@@ -137,3 +137,23 @@ def test__handle_contribution_update_when_no_contribution(self, mocker, payment_
...
+    @pytest.mark.parametrize("has_pm_id", [True, False])
+    def test__add_pm_id_and_payment_method_details(self, mocker, has_pm_id, payment_intent_succeeded_one_time_event):
+        processor = StripeWebhookProcessor(event=StripeEventData(**payment_intent_succeeded_one_time_event))
+        mocker.patch("apps.contributions.models.Contribution.fetch_stripe_payment_method", return_value=mocker.Mock())
+        mocker.patch.object(
+            processor, "contribution", new_callable=mocker.PropertyMock, return_value=ContributionFactory()
+        )
+        processor._add_pm_id_and_payment_method_details(pm_id="pm_id" if has_pm_id else None, update_data=dict())
+
+    def test_route_request_when_event_type_is_payment_method_attached(
```
benjaminEwhite, line 150: Added this test even though the object code is untouched because got us to 100% coverage on apps/contributions/webhooks.py

apps/contributions/management/commands/import_stripe_transactions_data.py
```diff
@@ -71,6 +77,7 @@ def handle(self, *args, **options):
                     stripe_account_id=account,
                     from_date=int(options["gte"].timestamp()) if options["gte"] else None,
                     to_date=int(options["lte"].timestamp()) if options["lte"] else None,
+                    retrieve_payment_method=options["retrieve_payment_method"],
```
njharman, line 80: Out of scope. [But I can't stop myself from mentioning]

Is it intentional that the options are slightly different between async and sync?

 `from_date=int(options["gte"].timestamp()) if options["gte"] else None,`
 vs
` from_date=options["gte"],`

If not, a DRYer way to accomplish
```
if  options["async_mode"]:
   func = task_import_contributions_and_payments_for_stripe_account.delay
else:
   func = StripeTransactionsImporter
for account in account_ids:
   func(...your kwargs here...)
```
If parameters don't match exactly, may have to "Facade" (or change func def cause params should match). But, that is better than having to maintain two identical parameter sections.

apps/contributions/management/commands/sync_missing_provider_payment_method_details.py
```diff
@@ -0,0 +1,118 @@
...
+import reversion
+import stripe
+
+from apps.common.utils import get_stripe_accounts_and_their_connection_status
+from apps.contributions.models import Contribution
+from apps.contributions.stripe_import import STRIPE_API_BACKOFF_ARGS
+
+
+# otherwise we get spammed by stripe info logs when running this command
+stripe_logger = logging.getLogger("stripe")
```
njharman, line 17: Consider not adding to global namespace.

`logging.getLogger("stripe").setLevel(logging.ERROR)`

apps/contributions/management/commands/sync_missing_provider_payment_method_details.py
```diff
@@ -0,0 +1,118 @@
...
+            contributions.values_list("stripe_account", flat=True).distinct()
+        )
+        unretrievable_accounts = [k for k, v in accounts.items() if not v]
+        connected_accounts = [k for k, v in accounts.items() if v]
+        ineligible_because_of_account = contributions.filter(stripe_account__in=unretrievable_accounts)
+        fixable = contributions.filter(stripe_account__in=connected_accounts)
+        if ineligible_because_of_account.exists():
+            self.stdout.write(
+                self.style.HTTP_INFO(
+                    f"Found {(_inel_account:=ineligible_because_of_account.count())} contribution{'' if _inel_account == 1 else 's'} with "
```
njharman, line 84: "_inel_account" hurts ;).  below is "fixiable_count", consider "ineligible_count" or simply "count".

Also, curious why fixable_count is not prepended with underscore and this one is? I have no opinion other than a preference for consistency.

[I think of := as very temporary vars, to be used in same line or within block of if x := "foo":.
Which is why I'm always pushing back / cringing when I see you := something and use it much later, much further down.
I'm not sure why I think this way...]

apps/contributions/webhooks.py
```diff
@@ -269,15 +275,16 @@ def handle_invoice_payment_succeeded(self):
                 self.obj_data["payment_intent"],
                 stripe_account=self.event.account,
             )
-            self._handle_contribution_update(
-                {
+            udpate_data = self._add_pm_id_and_payment_method_details(
```
njharman, line 278: typo?  udpate vs update?

njharman (approved):

apps/contributions/management/commands/sync_missing_provider_payment_method_details.py
benjaminEwhite: 1ed0c849654 made this change in all places that had this code.

apps/contributions/webhooks.py
benjaminEwhite: bea2622

apps/contributions/management/commands/sync_missing_provider_payment_method_details.py
benjaminEwhite: 7b44c2b3

apps/contributions/management/commands/import_stripe_transactions_data.py
benjaminEwhite: Sync can take a datetime as arg, but async needs it serialized/stringified, hence why different options.

apps/contributions/management/commands/fix_recurring_contribution_missing_provider_payment_id.py
```diff
@@ -11,8 +11,7 @@
 
 
 # otherwise we get spammed by stripe info logs when running this command
-stripe_logger = logging.getLogger("stripe")
-stripe_logger.setLevel(logging.ERROR)
+logging.getLogger("stripe").setLevel(logging.ERROR)
```
benjaminEwhite, line 14: Norm made this suggestion in `sync_missing_provider_payment_method_details` and I acted on it there and figured might as well update other places where we do this.a

apps/contributions/management/commands/sync_payment_transaction_time.py
```diff
@@ -14,8 +14,7 @@
 
 
 # otherwise we get spammed by stripe info logs when running this command
-stripe_logger = logging.getLogger("stripe")
-stripe_logger.setLevel(logging.ERROR)
+logging.getLogger("stripe").setLevel(logging.ERROR)
```
benjaminEwhite, line 17: Norm made this suggestion in `sync_missing_provider_payment_method_details` and I acted on it there and figured might as well update other places where we do this.a

apps/contributions/management/commands/sync_missing_provider_payment_method_details.py
```diff
@@ -0,0 +1,117 @@
...
+                    f"because account is disconnected or some other problem retrieving account: "
+                    f"{', '.join(str(x) for x in ineligible_because_of_account.values_list('id', flat=True))}"
+                )
+            )
+        self.stdout.write(
+            self.style.HTTP_INFO(
+                f"Found {(fixable_count:=fixable.count())} eligible contribution{'' if fixable_count == 1 else 's'} to sync"
+            )
+        )
+        if fixable_count:
```
x110dc, line 94: Maybe it's just me but the I think the walrus operator made this harder to follow. I was looking for where this was set the same or higher indention level. I like walrus, but in this case I think I'd prefer to see it set before the `write()` and then used there and this conditional.

x110dc (approved):
Had one minor style suggestion. Can take or leave it. Otherwise LGTM.

apps/contributions/management/commands/sync_missing_provider_payment_method_details.py
benjaminEwhite: 85cead6fc

PR #1524, opened 2024-05-28
DEV-4818 | Solve for backfill slowness

benjaminEwhite:

#### What's this PR do?

- Adds more granular logging in Stripe import
- Should solve for double-logging issues that stemmed from net adding a handler rather than changing formatter of default
- Removes `StripeTransactionsImporter.get_refunds_for_subscription` and just uses `.get_refunds_for_charge` in place. This method called `.get_charges_for_subscription` internally. This was needlessly increasing amount of retrieval work (and in a place that is relatively inefficient where we have to scan a key pattern)
- Sets up sentry to profile (https://docs.sentry.io/platforms/python/profiling/#enable-profiling). By default we don't get it, but if you want it to run with the stripe import command, include the `--sentry-profiler` flag when running.
- Alters management command so it checks for account connection status for each Stripe account before attempting to process. If there's a permission error retrieving the account, it reports on this but does not error out.

- Better logging
- Hopefully faster execution

Coming soon

No

N/A (only existing one removed as a method was removed).

N/A

No

N/A

[DEV-4818](https://news-revenue-hub.atlassian.net/browse/DEV-4818)

No

No

[DEV-4818]: https://news-revenue-hub.atlassian.net/browse/DEV-4818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/stripe_import.py
```diff
@@ -281,6 +290,11 @@ def __post_init__(self) -> None:
         self.created_payment_ids = set()
         self.updated_payment_ids = set()
 
+        hub = Hub.current
```
x110dc, line 293: I'm a little confused about the Sentry profile setup here. It doesn't seem to follow the instructions here https://docs.sentry.io/platforms/python/profiling/#enable-profiling . I don't see the the DSN anywhere so I don't know how this could work.

x110dc (changes requested):

apps/contributions/management/commands/import_stripe_transactions_data.py
```diff
@@ -66,18 +68,33 @@ def configure_stripe_log_level(self, suppress_stripe_info_logs: bool) -> None:
...
+        else:
+            return True
+
     def handle(self, *args, **options):
         command_name = os.path.basename(__file__).split(".")[0]
         self.stdout.write(self.style.HTTP_INFO(f"Running {command_name}"))
         self.configure_stripe_log_level(options["suppress_stripe_info_logs"])
-        account_ids = self.get_stripe_account_ids(options["for_orgs"], options["for_stripe_accounts"])
+        _account_ids = self.get_stripe_account_ids(options["for_orgs"], options["for_stripe_accounts"])
+        account_ids = [account for account in _account_ids if self.stripe_account_connected(account)]
```
x110dc, line 85: Since most accounts are connected I'd think it would make sense to optimistically assume they'll connect and just catch and handle the exceptions when they don't instead of this long-running up-front work to connect to every single account?

x110dc (changes requested):

apps/contributions/management/commands/import_stripe_transactions_data.py
benjaminEwhite: The alternative was something like adding try/except stripe.error.PermissionError around `.import_contributions_and_payments`, which I don't love, but perhaps for aesthetic reasons more than anything substantive.

benjaminEwhite: Ended up moving to a try/except that wraps the synchronous calling context (if there's a problem in async, that won't stop execution of each account/org). 999e9571adf

x110dc (approved):

PR #1529, opened 2024-05-31
DEV-4777 Don't use naive datetimes

njharman:
#### What's this PR do?

- Enables Ruff DTZ naive datetime rules
- Enlightens naive datetimes.

Note: three `make_aware `and ne `convert_to_timezone_formatted `tests helperare testing naive to non-naive conversion and are noqa'd

PR comments highlight these issues to be resolved during review
- Several tests depend on `user.accepted_terms_of_service` being naive. I'm not sure this is valid. ~But did not want to "fix" before receiving review feedback.~ Have updated to be not naive. per DC. Also the test that was failing; was not specifically testing for naivety. Rather it just made assumptions about string format and compared formatted strings. Changed to compare datetimes.  
- There are some set cookie that i updated to use aware timezones for expiry. Note sure if this is correct HTTP standard?

Time is hard. This helps avoid problems. Like, don't you expect datetime.datetime.utcnow() to have UTC timezone set. I did.

No. Not sure what should be tested. Probably user acceptance of terms.

No.

Slight change to some, no new ones.

N/A

- pytz is used in three test files. I thought we got rid of that. Maybe scope of this ticket should expand to replacing pytz mostly pytz.UTC with one pytz.utc.localize().

N/A

[DEV-4777]

No.

No.

[DEV-4777]: https://news-revenue-hub.atlassian.net/browse/DEV-4777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/api/tests/test_views.py
```diff
@@ -92,7 +91,9 @@ def test_construct_rp_domain_with_hostmap_but_no_map(self, settings):
 
 @pytest.fixture()
 def org_user_with_pw(org_user_free_plan):
-    org_user_free_plan.accepted_terms_of_service = datetime.datetime.utcnow()
+    org_user_free_plan.accepted_terms_of_service = (
+        datetime.datetime.utcnow()  # noqa: DTZ003 test depends on naive, breaks test
```
njharman, line 95: I question if tests should expect naive datetime. 

Isn't this "user" eventually a DB model with timezone? Or, is this simulation "web" data, with no actual TZ.  users test_serializers compares to pytz.utc.localize() on this

conftest.py
```diff
@@ -137,7 +137,9 @@ def org_user_free_plan(default_feature_flags) -> User:
 
 @pytest.fixture()
 def user_with_verified_email_and_tos_accepted():
-    return UserFactory(accepted_terms_of_service=datetime.datetime.utcnow(), email_verified=True)
+    return UserFactory(
+        accepted_terms_of_service=datetime.datetime.utcnow(), email_verified=True  # noqa: DTZ003 test fails
```
njharman, line 141: test fails if not naive

apps/api/tests/test_views.py
```diff
@@ -122,7 +123,7 @@ def user_no_role_assignment_with_pw(user_no_role_assignment):
 @pytest.fixture()
 def rp_user_with_pw(rp_user):
     rp_user.set_password(KNOWN_PASSWORD)
-    rp_user.accepted_terms_of_service = datetime.datetime.utcnow()
+    rp_user.accepted_terms_of_service = datetime.datetime.utcnow()  # noqa: DTZ003 test depends on naive, breaks test
```
njharman, line 126: test fails if not naive

apps/api/views.py
```diff
@@ -119,7 +119,7 @@ def post(self, request, *args, **kwargs):
         return set_token_cookie(
             response,
             jwt_serializer.validated_data["access"],
-            datetime.now() + settings.SIMPLE_JWT["ACCESS_TOKEN_LIFETIME"],
+            datetime.datetime.now(datetime.timezone.utc) + settings.SIMPLE_JWT["ACCESS_TOKEN_LIFETIME"],
```
njharman, line 122: This sets cookie lifetime. I assume it eventually gets formatted into whatever HTTP standard requires for cookies. And that making it tz aware won't break the internet.

apps/api/tests/test_views.py
x110dc: I think we should make all of these TZ aware.

apps/api/tests/test_views.py
x110dc: Make it TZ aware.

apps/api/views.py
x110dc: I would think so but I'm not sure.

conftest.py
x110dc: And we can rewrite the test to use aware?

x110dc (changes requested):

apps/api/views.py
benjaminEwhite: Should be straightforward to look at actual jwt in review app in authorization cookie and decode it at jwt.io and confirm that the iat/exps values make sense.

Screenshot 2024-06-11 at 5 27 00 PM

benjaminEwhite (approved):
LGTM.

x110dc (approved):
lgtm

PR #1538, opened 2024-06-06
DEV-3658 | Quarantine queue bug fixes and improvements

benjaminEwhite:


apps/contributions/tasks.py
```diff
@@ -46,28 +47,44 @@ def ping_healthchecks(check_name, healthcheck_url):
...
+        len(eligible_flagged_contributions),
     )
-    logger.info("Found %s flagged contributions past auto-accept date", len(eligible_flagged_contributions))
 
     for contribution in eligible_flagged_contributions:
-        payment_intent = contribution.get_payment_manager_instance()
+        manager = contribution.get_payment_manager_instance()
         try:
-            payment_intent.complete_payment(reject=False)
+            manager.complete_payment(reject=False)
```
benjaminEwhite, line 80: The original var name here was incoherent. `.get_payment_manager_instance` does not return a PI.

apps/contributions/tests/test_webhooks_integration.py
```diff
@@ -66,23 +66,30 @@ def invoice_upcoming():
         return json.load(f)
 
 
-@pytest.fixture()
-def _invoice_payment_succeeded():
```
benjaminEwhite, line 70: This wasn't being used anywhere.

apps/contributions/payment_managers.py
```diff
@@ -55,16 +56,6 @@ def __init__(self, data=None, contribution=None):
...
 
-    def ensure_contribution(self):
-        if not self.contribution:
-            raise ValueError("Method requires PaymentManager to be instantiated with contribution instance")
-
-    @staticmethod
-    def get_subclass(contribution):
-        payment_provider_used = contribution.payment_provider_used
-        if payment_provider_used == PaymentProvider.STRIPE_LABEL:
-            return StripePaymentManager
```
benjaminEwhite, line 66: This was an odd bit of indirection — maybe introduced to deal with circular imports???

At any rate, I removed this code and just return StripePaymentManagement in situ.

apps/contributions/payment_managers.py
```diff
@@ -76,156 +67,199 @@ def get_serializer_class(self, data=None, contribution=None):
             return StripeOneTimePaymentSerializer
         return StripeRecurringPaymentSerializer
 
-    def attach_payment_method_to_customer(self, stripe_customer_id, org_stripe_account, payment_method_id=None):
```
benjaminEwhite, line 79: This method was unused, so removed.

apps/contributions/payment_managers.py
```diff
@@ -76,156 +67,199 @@ def get_serializer_class(self, data=None, contribution=None):
...
-                stripe_customer_id,
-                org_stripe_account,
-                payment_method_id,
-            )
-            raise PaymentProviderError("Something went wrong with Stripe") from None
-
     def complete_payment(self, reject=False):
         if self.contribution.interval == ContributionInterval.ONE_TIME:
             self.complete_one_time_payment(reject)
-        elif self.contribution.interval:
```
benjaminEwhite, line 99: contributions are guaranteed to have an interval. they cannot be created without one.

apps/contributions/models.py
```diff
@@ -175,6 +179,17 @@ def filtered_by_role_assignment(self, role_assignment: RoleAssignment) -> models
             case _:
                 return self.none()
 
+    def unmarked_abandoned_carts(self) -> models.QuerySet:
+        """Return contributions that have been abandoned.
+
+        We define abandoned as contributions that have been flagged or are in processing state for more than 24 hours.
```
njharman, line 185: Says 24 hours in docstring, but it's 8, right?  Consider referencing CONTRIBUTION_ABANDONED_THRESHOLD in docstring to avoid "value sync" issues in future.

apps/contributions/models.py
```diff
@@ -367,6 +383,17 @@ def formatted_donor_selected_amount(self) -> str:
     def expanded_bad_actor_score(self):
         return None if self.bad_actor_score is None else self.BAD_ACTOR_SCORES[self.bad_actor_score][1]
 
+    @property
+    def is_unmarked_abandoned_cart(self) -> bool:
+        return self.status in (ContributionStatus.FLAGGED, ContributionStatus.PROCESSING) and (
+            self.created < datetime.datetime.now(tz=timezone.utc) - CONTRIBUTION_ABANDONED_THRESHOLD
+        )
```
njharman, line 390: Is this intentionally different than  `unmarked_abandoned_carts()` which also requires   `provider_payment_method_id__isnull=True`?

If not, consider deduplicating this logic. Or, adding comments # if you change this also change other one

apps/contributions/models.py
```diff
@@ -625,6 +646,30 @@ def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.
...
         )
 
+    @cached_property
+    def stripe_subscriptions_for_customer(self) -> list[stripe.Subscription]:
+        """Return all subscriptions for the customer associated with this contribution."""
+        if not self.provider_customer_id:
+            return []
+        return stripe.Subscription.list(
+            customer=self.provider_customer_id, stripe_account=self.stripe_account_id
+        ).auto_paging_iter()
```
njharman, line 656: It's unclear to me if [auto_paging_iter ](https://docs.stripe.com/api/pagination/auto?lang=python) is an actual list().  It sounds like it is an iterable (generator)

If so, it might confuse or cause issue expecting real list based on Type Annotation of return value.

apps/contributions/payment_managers.py
```diff
@@ -76,156 +67,242 @@ def get_serializer_class(self, data=None, contribution=None):
...
+            "StripePaymentManager.complete_one_time_payment canceling Stripe PI %s for contribution %s",
+            payment_intent.id,
+            self.contribution.pk,
+        )
+        payment_intent.cancel(
+            self.contribution.provider_payment_id,
+            stripe_account=self.contribution.revenue_program.payment_provider.stripe_account_id,
+            cancellation_reason="fraudulent",
+        )
+        # we only want to update the status if the payment was successfully canceled
```
njharman, line 98: Is this relying on payment_intent.cancel() to raise if not successful?

Comment implies there should be something like an if cancelled check. If not/not needed, I'm more confused than aided by the comment. $.02

apps/contributions/payment_managers.py
```diff
@@ -76,156 +67,242 @@ def get_serializer_class(self, data=None, contribution=None):
...
-        if not (pi := self.contribution.stripe_payment_intent):
+        try:
+            match (
+                reject,
+                bool(self.contribution.provider_payment_id),
+                bool(pi := self.contribution.stripe_payment_intent),
+            ):
+                # if no provider payment id or unretrievable payment intent,
+                #  we won't be able to cancel in case of reject or capture in case of accept, so raise error.
+                case (_, False, _) | (_, _, False):
```
njharman, line 118: It took a bit of thinking to confirm all combinations are covered. Making this case the default `case _:` would make that more explicit.  But, it's also nice to have the two conditions that cause error represented.

Seems a bit twisted to fit into match statement. 

```
If DOA1 or DOA2:
   raise
try:
  if reject:
     reject()
  else:
     accept()
except:
  pass
```

apps/contributions/payment_managers.py
```diff
@@ -76,156 +67,242 @@ def get_serializer_class(self, data=None, contribution=None):
...
+                    True,
+                ):  # pragma: no cover  This last case makes the set of cases exhaustive, but pytest cov doesn't see it.
+                    update_data = self._handle_one_time_rejection(pi, update_data)
+
+        except stripe.error.StripeError as exc:
+            message = exc.error.message if exc.error else "Could not complete payment"
+            logger.exception("`StripePaymentManager.complete_one_time_payment` raised StripeError")
+            raise PaymentProviderError(message) from exc
+
+        finally:
```
njharman, line 141: I believe finally block will execute for 1st case (log error raise PaymentProviderError). updating with {}. Same if there is exception in other clauses.  Intended?

https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions

apps/contributions/tasks.py
```diff
@@ -46,28 +47,49 @@ def ping_healthchecks(check_name, healthcheck_url):
         time.sleep(1)
 
 
+@shared_task
+def mark_abandoned_carts_as_abandoned():
+    logger.info("Starting task, 'mark_abandoned_carts_as_abandoned'")
+    abandoned = Contribution.objects.unmarked_abandoned_carts()
+    logger.info("Found %s abandoned carts", abandoned.count())
+    updated = 0
+    if abandoned.exists():
```
njharman, line 56: Isn't if redundant?  If there are no abandoned, the for loop body will never execute.

njharman (changes requested):

apps/contributions/models.py
benjaminEwhite: ad3f6229

apps/contributions/models.py
benjaminEwhite: bcad044c6

apps/contributions/models.py
benjaminEwhite: 0727f5f0bc2

apps/contributions/tasks.py
benjaminEwhite: 102fca1adf7

apps/contributions/payment_managers.py
benjaminEwhite: 4a23438357

apps/contributions/payment_managers.py
benjaminEwhite: yeah, that's fine. `_handle_contribution_update` knows what to do when passed empty update data dict.

revengine/settings/base.py
benjaminEwhite, line 1: These changes are only here temporarily because merged in changes from https://github.com/newsrevenuehub/rev-engine/pull/1547 to unbreak in short term

apps/contributions/payment_managers.py
benjaminEwhite: ea107603342d

apps/contributions/models.py
```diff
@@ -367,6 +386,19 @@ def formatted_donor_selected_amount(self) -> str:
     def expanded_bad_actor_score(self):
         return None if self.bad_actor_score is None else self.BAD_ACTOR_SCORES[self.bad_actor_score][1]
 
+    @property
+    def is_unmarked_abandoned_cart(self) -> bool:
+        return (
+            not self.provider_payment_method_id
+            and self.status in (ContributionStatus.FLAGGED, ContributionStatus.PROCESSING)
+            and (self.created < datetime.datetime.now(tz=timezone.utc) - CONTRIBUTION_ABANDONED_THRESHOLD,)
```
njharman, line 394: Spurious comma `CONTRIBUTION_ABANDONED_THRESHOLD,)`

apps/contributions/payment_managers.py
```diff
@@ -76,156 +67,234 @@ def get_serializer_class(self, data=None, contribution=None):
...
+            else:
+                update_data = self._handle_one_time_acceptance(pi, update_data)
+        except stripe.error.StripeError as exc:
+            message = exc.error.message if exc.error else "Could not complete payment"
+            logger.exception("`StripePaymentManager.complete_one_time_payment` raised StripeError")
+            raise PaymentProviderError(message) from exc
         else:
+            self._handle_contribution_update(
+                update_data=update_data, caller="StripePaymentManager.complete_one_time_payment"
+            )
```
njharman, line 129: I'm maybe confused. The commit below (which cuts comment, also renamed  `self._handle_contribution_update()` to `self.git()`  But I don't see that here. That means it was a change you've undone in current state of branch, right?

If not, I'm not understanding the name .git() over ._handle_contribution_update()

https://github.com/newsrevenuehub/rev-engine/commit/4a2343835721c9ad3c369a4a59d16a8ecc7237ad

njharman (approved):

apps/contributions/models.py
benjaminEwhite: ecf9273572

apps/contributions/payment_managers.py
benjaminEwhite: That was a bunk commit (was in editor starting a git command thinking i was in terminal, i gather). Original method name restored.

revengine/settings/base.py
```diff
@@ -11,37 +11,34 @@
 
 import json
```
x110dc, line 12: It looks like everything in these settings files is from a bad merge. Can you fix?

x110dc (changes requested):

apps/contributions/tasks.py
```diff
@@ -46,28 +47,48 @@ def ping_healthchecks(check_name, healthcheck_url):
...
+    logger.info("Found %s abandoned carts", abandoned.count())
+    updated = 0
+    for contribution in abandoned:
+        logger.info("Marking contribution %s as abandoned", contribution.id)
+        with reversion.create_revision():
+            contribution.status = ContributionStatus.ABANDONED
+            contribution.save(update_fields={"status", "modified"})
+            reversion.set_comment("`mark_abandoned_carts_as_abandoned` task marked as abandoned")
+            updated += 1
+    logger.info("Marked %s contributions carts as abandoned", updated)
```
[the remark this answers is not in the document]
benjaminEwhite: de92ba22a

Also created an ITS ticket to set the related env var in prod.

revengine/settings/base.py
benjaminEwhite: These were merged in from a different branch to solve for regression. Moving conversation about this to JIRA.

PR #1539, opened 2024-06-07
DEV-4110: Upgrade Django to 4

nrh-cklimas:

#### What's this PR do?

This is a redo of https://github.com/newsrevenuehub/rev-engine/pull/1453 now that we pulled a lot of its content into a separate PR we've already merged, https://github.com/newsrevenuehub/rev-engine/pull/1517.

The most significant change here is using a DB collation + `CharField` in the `DenyListWord` model instead of `CICharField`, which has been deprecated in Django 4.

Our current Django version is no longer supported; this puts us on the most recent LTS release.

Yes.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-4110

No.

No.


apps/users/models.py
```diff
@@ -69,13 +69,13 @@ def role_type(self) -> RoleTypeReturnValues:
         ra = self.get_role_assignment()
         return (ra.role_type, ra.get_role_type_display()) if ra else None
 
-    def validate_unique(self, exclude) -> None:
+    def validate_unique(self, exclude: set) -> None:
```
nrh-cklimas, line 72: This was a little odd because it wasn't in the original PR, but I got an error in unit tests without this change. [Docs say this is a set, not a list](https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.validate_unique).

apps/config/models.py
```diff
@@ -1,13 +1,16 @@
...
 class DenyListWord(models.Model):
-    word = CICharField(max_length=255, unique=True)  # CI == Case Insensitive
+    word = models.CharField(max_length=255, unique=True)
 
     class Meta:
         verbose_name = "Deny-list word"
         verbose_name_plural = "Deny-list words"
+        constraints = [
+            models.UniqueConstraint(Lower("word"), name="word_unique_case_insensitive"),
+        ]
```
benjaminEwhite, line 13: Why are we going with this over using collation strategy we were pursuing in past?

Might be worth addressing this in PR description.  Downside of this approach is that it's likely less performant than using a db-level collation, but that's probably not a big deal here since `DenyListWord` is infrequently accessed.

benjaminEwhite (commented):
One comment around pivot away from collation strategy and PR description.

Also wanted to point out that we could at least consider moving the case insensitive validation bits into a separate PR so that we have the upgrade changes here in isolation, but that might be overkill.

benjaminEwhite (changes requested):
Final request:

Think it's worthwhile adding unit tests of `validate_slug_against_denylist`. It is not tested in isolation in status quo.

apps/config/models.py
nrh-cklimas: Updated in b78c5ea6876252891341b1f4612f019d91998bce

nrh-cklimas:
> Think it's worthwhile adding unit tests of `validate_slug_against_denylist`. It is not tested in isolation in status quo.

ac1d96d880f49b2c8d03eb77ece0f0b2baa02297

apps/config/migrations/0003_DEV-4110_add_case_insensitive_collation.py
benjaminEwhite, line 1: Probably worth squashing the two migrations into a single one if possible (fwiw, [this is the standard established in 17 here](https://news-revenue-hub.atlassian.net/wiki/spaces/TECH/pages/44466177/Pull+Request+Policies)).

benjaminEwhite (changes requested):
@nrh-cklimas 

- The PR description needs to be updated to reflect the current decisions around collation
-  I think in an earlier version there was testing around the confiword model that proved that you got case inensitive search. I think that would be good to bring back, because the business logic here spans both not allowing creation of duplicate (from perspective of case insensitivity) and furnishing case insensitive results.

benjaminEwhite (approved):

njharman (approved):

apps/common/tests/test_utils.py
```diff
@@ -73,18 +73,21 @@ def test_custom_length_enforced():
 def setup_request(user, request):
     request.user = user
 
-    # Annotate a request object with a session
-    middleware = SessionMiddleware()
+    # Annotate a request object with a session.
+    # Note that we need to pass a callable to the middleware in Django 4.2. Here, we just have a dummy
+    # that returns the request object.
+    middleware = SessionMiddleware(get_response=(dummy := lambda request: request))
```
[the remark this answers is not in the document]
nrh-cklimas: 6b8569e67cf1b6b477bff8cab3c13e3db45e321e

revengine/settings/deploy.py
```diff
@@ -30,14 +30,22 @@
...
-GS_DEFAULT_ACL = None
-# Media files are stored in a 'media' directory
-GS_MEDIA_LOCATION = "media"
+STORAGES = {
+    # Store user-uploaded files in Google Cloud.
+    "default": {
+        "BACKEND": "storages.backends.gcloud.GoogleCloudStorage",
+        # https://django-storages.readthedocs.io/en/latest/backends/gcloud.html#settings
+        "OPTIONS": {
+            "bucket_name": os.getenv("GS_BUCKET_NAME"),
```
x110dc, line 39: It looks like there was a default here (`rev-engine-media`) that was removed here. Are we certain this is defined in the environments where it's required?

revengine/settings/deploy.py
```diff
@@ -30,14 +30,22 @@
...
 
 ### Google Cloud Storage ###
-GS_BUCKET_NAME = os.getenv("GS_BUCKET_NAME", "rev-engine-media")
-DEFAULT_FILE_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
-GS_PROJECT_ID = os.getenv("GS_PROJECT_ID", "revenue-engine")
-# https://django-storages.readthedocs.io/en/latest/backends/gcloud.html#settings
-GS_QUERYSTRING_AUTH = False
-GS_DEFAULT_ACL = None
-# Media files are stored in a 'media' directory
-GS_MEDIA_LOCATION = "media"
```
x110dc, line 40: I see why the other params are no longer necessary because they're specified inline, but will the removal of `GS_MEDIA_LOCATION` affect anything?

x110dc (changes requested):
A couple of questions about settings.

revengine/settings/deploy.py
nrh-cklimas: This is currently set in all envs, but I don't see much upside to removing the default. Updated in c79bef069849dd9cd6bdeeec9783a0b7f51ee129

revengine/settings/deploy.py
nrh-cklimas: I think this was an oversight on my part when I was translating settings to the new config structure. Added in in c79bef069849dd9cd6bdeeec9783a0b7f51ee129

x110dc (approved):
lgtm

PR #1530, opened 2024-05-31
DEV-4779 consistent boolean environment vars

njharman:
#### What's this PR do?
Resolve Ruff issue PLW1508 which has to do with os.getenv default values being something other than string.

Refactor to use [environs](https://pypi.org/project/environs/)

Notes:
- not using env.url because it returns the parsed url object, which broke tests. And also it fails if no URL.
- Added many "" defaults. As env.foo() without default marks a "required" envvar. Which is change in behavior from os.getenv.  Alternatively could have added a bunch of required ENVVARS. Which seems much more disruptive and risky

~Creates function to booleanize envvar strings.~

#### Why are we doing this? How does it help us?
To replace the hodgepodge of inconsistent conversion to bool.
Missing or malformed envvars produce clear error message (exception).

#### Are there detailed, specific, step-by-step testing instructions in the associated ticket?
A test script to run.

#### Did this PR make changes to the user interface that may require changes to the user-facing docs?
No.

#### Have automated unit tests been added? If not, why?
Yes.

#### How should this change be communicated to end users?
N/A

#### Are there any smells or added technical debt to note?
No.

#### Has this been documented? If so, where?
N/A

#### What are the relevant tickets? Add a link to any relevant ones.
[DEV-4779]

#### Do any changes need to be made before deployment to production (adding environment variables, for example)?
This "changes" how environment variables are read. 

1. Production settings `prod.py`? should be altered to use env.str()
2. Each environments envvars should be audited for compatibility
3. Run test script https://news-revenue-hub.atlassian.net/browse/DEV-4779

**Colon separated list envvars.** Now converted into a list by env.list(), previously by ":".split()

- DASHBOARD_SUBDOMAINS    
- ALLOWED_HOSTS  

**Boolean envvars.** 
The following values will parse to False and True respectively:

- falsy = {'0', 'F', 'FALSE', 'False', 'N', 'NO', 'No', 'OFF', 'Off', 'f', 'false', 'n', 'no', 'off', 0}
- truthy = {'1', 'ON', 'On', 'T', 'TRUE', 'True', 'Y', 'YES', 'Yes', 'on', 't', 'true', 'y', 'yes', 1}

These are boolean envvars with default value from setings/base.py
- CSP_REPORTING_ENABLE, False
- DATABASE_SSL, False
- DJANGO_DEBUG, False
- ENABLE_API_BROWSER, False
- ENABLE_GOOGLE_CLOUD_SECRET_MANAGER, False
- ENABLE_PUBSUB, False
[...]

PR #1549, opened 2024-06-19
DEV-4904 string settings, default emptystring

njharman:
#### What's this PR do?
Combines DEV-4779 with DEV-4904.  DEV-4779 was rolled back. DEV-4904 includes (hopefully) fix for issue that caused rollback. Which is environs treating defaultless env() as "required" and raising if envvar is not present. The replaced stdlib os.getenvcode did not raise.

1. Refactor settings to use 3rd party [enviorns ](https://pypi.org/project/environs/)env()
2. Add defaults to all env() calls. Alternatively could have added a bunch of required ENVVARS. Which seems much more disruptive and risky.
3. Replace None default to empty string for all env.str with None default.
4. I Verified for each constant changed from None to ““; that no code expected it to be None. A CONSTANT is None for example.
5. Add empty string default to: CSP_REPORT_URL, DATABASE_URL, DJANGO_SECRET_KEY
6. Update organizations tasks.py and tests to also use empty string instead of None for relevant constants.
7. Cut REQUEST_ID_HEADER as unused. https://news-revenue-hub.atlassian.net/browse/DEV-4904?focusedCommentId=116951

Not using env.url because it returns the parsed url object, which broke tests. And also it fails if no URL provided (doesn't allow empty string).

- To replace the hodgepodge of inconsistent conversion to bool.
- Missing or malformed envvars produce clear error message (exception).
- Consistency (some settings had "" default, others None).
- Environment variables are strings. Defaults should be strings.

_Python treating empty string as false makes it easy to not make every string a Union(str|None). Being strongly typed as str, means we don't have to check `if not None` every time we try some str method or function that fails with None, len(), strip(), etc._

Hard to test. There is pre-flight.py smoketest script attached to ticket.

No.

No. Settings files are not tested (and testing them would mostly just be testing environs lib and Python.

N/A

Adding default for "required" envvars is safer but possibly riskier in that it hides issues. Such as 
missing DJANGO_SECRET_KEY causing empty string to be used.  
 
N/A

https://news-revenue-hub.atlassian.net/browse/DEV-4904
https://news-revenue-hub.atlassian.net/browse/DEV-4779 (merged with this one)

- Production settings prod.py? should be altered to use environs / env()
- Each environments envvars should be audited for compatibility
- Run the pre-flight script attached to ticket as a smoke test.

**Colon separated list envvars** Now converted into a list by env.list(), previously by ":".split()

- DASHBOARD_SUBDOMAINS
- ALLOWED_HOSTS

**Boolean envvars**
The following values will parse to False and True respectively:

- falsy = {'0', 'F', 'FALSE', 'False', 'N', 'NO', 'No', 'OFF', 'Off', 'f', 'false', 'n', 'no', 'off', 0}
- truthy = {'1', 'ON', 'On', 'T', 'TRUE', 'True', 'Y', 'YES', 'Yes', 'on', 't', 'true', 'y', 'yes', 1}

These are boolean envvars with default value from setings/base.py

- CSP_REPORTING_ENABLE, False
- DATABASE_SSL, False
- DJANGO_DEBUG, False
- ENABLE_API_BROWSER, False
- ENABLE_GOOGLE_CLOUD_SECRET_MANAGER, False
- ENABLE_PUBSUB, False
- ENFORCE_CSP, True
- GS_CREDENTIALS_RAISE_ERROR_IF_UNSET, True
- SECURE_SSL_REDIRECT, True
- SENTRY_ENABLE_BACKEND, False
- SENTRY_ENABLE_FRONTEND, False
- SENTRY_ENABLE_PII, True
- SPA_ENV_CAPTURE_PAGE_SCREENSHOT, False
- STRIPE_LIVE_MODE, False
- UPDATE_LAST_LOGIN, True
- USE_DEBUG_INTERVALS, False

and from settings/deploy.py

- CSRF_COOKIE_SECURE, True
- SESSION_COOKIE_SECURE, True
- USE_DEBUG_INTERVALS, False

Not exactly but related https://news-revenue-hub.atlassian.net/browse/DEV-4924


x110dc (approved):

revengine/settings/base.py
```diff
@@ -612,58 +611,57 @@ class CurrencyDict(TypedDict):
...
-
 logger = logging.getLogger(f"{DEFAULT_LOGGER}.{__name__}")
 
 try:
-    HOST_MAP = json.loads(os.getenv("HOST_MAP", "{}"))
-except json.JSONDecodeError:
+    # Host map for client custom hostnames. This should be a JSON-encoded dictionary
+    # of { "customhostname.org": "rp-slug" } values.
+    HOST_MAP = env.json("HOST_MAP", "{}")
+except (json.JSONDecodeError, environs.EnvError, environs.EnvValidationError):
```
benjaminEwhite, line 629: Which conditions would case each of the three distinct exception types?

revengine/settings/deploy.py
```diff
@@ -10,29 +9,29 @@
 
 #### Critical settings
 
-SECRET_KEY = os.getenv("DJANGO_SECRET_KEY")
+SECRET_KEY = env.str("DJANGO_SECRET_KEY", "")
 
 ### Environment-specific settings
-ALLOWED_HOSTS = os.getenv("ALLOWED_HOSTS", "localhost").split(":")
+ALLOWED_HOSTS = env.list("ALLOWED_HOSTS", "localhost", delimeter=":")
```
benjaminEwhite, line 15: Typo here

```suggestion
ALLOWED_HOSTS = env.list("ALLOWED_HOSTS", "localhost", delimiter=":")
```

I understand why we don't unit test settings files, but this is an example of where it would be helpful to do so because catch something like this, which would potentially lead to prod being broken post deploy.  To be clear, not suggesting we pivot to testing settings.

revengine/settings/deploy.py
```diff
@@ -116,4 +115,4 @@
     ignore_logger("django.security.DisallowedHost")
 
 
-USE_DEBUG_INTERVALS = os.getenv("USE_DEBUG_INTERVALS", False)
```
benjaminEwhite, line 119: Isn't this redundant? We have same line in base.py, which this imports from.

revengine/settings/deploy.py
```diff
@@ -10,29 +9,29 @@
...
 
 ANYMAIL = {
-    "MAILGUN_API_KEY": os.getenv("MAILGUN_API_KEY", ""),
+    "MAILGUN_API_KEY": env.str("MAILGUN_API_KEY", ""),
     "MAILGUN_SENDER_DOMAIN": MAILGUN_SENDER_DOMAIN,
 }
 
 EMAIL_SUBJECT_PREFIX = f"[revengine {ENVIRONMENT.title()}] "
-DEFAULT_FROM_EMAIL = f"noreply@{os.getenv('DOMAIN', 'example.com')}"
+DEFAULT_FROM_EMAIL = f"noreply@{env.str('DOMAIN', 'example.com')}"
```
benjaminEwhite, line 27: This is redundant. Have same line in base.py, which is imported here.

Also, thinking about prod in particular, wouldn't we want to not have a default here and raise an error if DOMAIN not defined, because I don't think we'd want to have emails being sent with from email that has a fake domain.

revengine/settings/base.py
```diff
@@ -572,18 +571,18 @@ class CurrencyDict(TypedDict):
 REQUESTS_TIMEOUT_DEFAULT = 31
 
 ## Email and ESP Settings
-DEFAULT_FROM_EMAIL = f"noreply@{os.getenv('DOMAIN', 'example.com')}"
+DEFAULT_FROM_EMAIL = f"noreply@{env.str('DOMAIN', 'example.com')}"
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
```
benjaminEwhite, line 575: looks like this this var gets redundantly redefined at line 583.

revengine/settings/deploy.py
```diff
@@ -10,29 +9,29 @@
...
 EMAIL_BACKEND = "anymail.backends.mailgun.EmailBackend"
 MAILGUN_SENDER_DOMAIN = "fundjournalism.org"
 
 ANYMAIL = {
-    "MAILGUN_API_KEY": os.getenv("MAILGUN_API_KEY", ""),
+    "MAILGUN_API_KEY": env.str("MAILGUN_API_KEY", ""),
     "MAILGUN_SENDER_DOMAIN": MAILGUN_SENDER_DOMAIN,
 }
 
 EMAIL_SUBJECT_PREFIX = f"[revengine {ENVIRONMENT.title()}] "
```
benjaminEwhite, line 26: This is redundant. Already defined this way in base.py.

benjaminEwhite (changes requested):

revengine/settings/base.py
njharman: Quoting Nick https://github.com/newsrevenuehub/rev-engine/pull/1530/files/6ab2c23d466b1c1d8ae913dd88554141cf179c22#r1634915449
> I tried this out in the REPL and invalid JSON seems to now raise environs.EnvValidationError, not JSONDecodeError.

I'm overcompensating for causing prod exception / deployment reversion. 

environs.EnvValidationError is completely redundant. And appears to not be "public" from environs/__init__.py

```
__all__ = ["EnvError", "Env"]
...
 class EnvValidationError(EnvError):
```

Compromise? 81ce2c746

revengine/settings/base.py
njharman: Yes, this is already resolved in https://news-revenue-hub.atlassian.net/browse/DEV-4924 (which is up for review but we should finish this, 4904, first.

37ea0b678

revengine/settings/deploy.py
njharman: This is the second bug due to not having unittests for settings. https://github.com/newsrevenuehub/rev-engine/pull/1530/files#r1636745783

The pre-flight.py looks good cause default is only one-element list. If prod had longer list, and deployer was eagle-eyed, might be caught.

There's probably a reason, but really annoyed they accept all kwargs (which is why my silly import test did not fail)
```
def _preprocess_list(value, *, delimiter= ",", **kwargs):
   kwargs not used
```

I was about to write tests (for 2nd time). But, there's really so little to test, that isn't just verifying environs works. This is all I came up with 4f1bb8b05

Adding/maintaining tests for each env.list() and similar; seems not ROI worthy. These are issues where we have to rely on reviews and QA testing to catch. IMHO.

benjaminEwhite (approved):

nrh-cklimas:
Closed this by accident--reopening.

PR #1552, opened 2024-06-24
DEV-4706: Add payment history to recurring contribution resend receipt email

GuiMend:

n/a
#### What's this PR do?
- Adds Billing History to email template 
- Send Billing History in email when user requests receipt 
- Improve user experience when requesting receipt on past contributions
- Yes
- no
- yes
- Billing history now available in receipts emails
- no
- no
- https://news-revenue-hub.atlassian.net/browse/DEV-4706
- no


apps/contributions/models.py
```diff
@@ -541,19 +545,54 @@ def cancel(self):
             self.save(update_fields={"status", "modified"})
             reversion.set_comment(f"`Contribution.cancel` saved changes to contribution with ID {self.id}")
 
-    def handle_thank_you_email(self):
+    def handle_thank_you_email(self, billing_history: list[Contribution.Billing] | None = None):
```
benjaminEwhite, line 548: Seems strange to define this class on the Contribution class. I'd move it to module scope.

apps/contributions/models.py
```diff
@@ -541,19 +545,54 @@ def cancel(self):
...
+        payment_status: str
+
+    def get_billing_history(self) -> list[Billing] | None:
+        """Get the billing history of a recurring contribution."""
+        if self.interval == ContributionInterval.ONE_TIME:
+            logger.info(
+                "`Contribution.get_billing_history` called on ID: %s with one-time interval. No billing history available",
+                self.id,
+            )
+            return None
```
benjaminEwhite, line 575: There should still be a billing history in case of one time contribution, no? It would have one payment object for successful payment and could have additional for refunds.

apps/contributions/models.py
```diff
@@ -541,19 +545,54 @@ def cancel(self):
...
+
+        payments: list[Payment] = self.payment_set.all()
```
benjaminEwhite, line 577: This can just be referenced in place at line 588, no need to assign to var.

apps/contributions/views.py
```diff
@@ -674,7 +674,8 @@ def send_contribution_receipt(self, request, pk=None, contribution_id=None) -> R
             contribution = self.exclude_hidden_statuses(contributor.contribution_set).get(pk=contribution_id)
         except Contribution.DoesNotExist:
             return Response({"detail": "Contribution not found"}, status=status.HTTP_404_NOT_FOUND)
-        contribution.handle_thank_you_email()
+        billing_history = contribution.get_billing_history()
+        contribution.handle_thank_you_email(billing_history)
```
benjaminEwhite, line 678: Given this context, might make more sense to have `handle_thank_you_email` make call to `.get_billing_history` internally, rather than passing as param.

apps/emails/views.py
```diff
@@ -49,6 +49,18 @@ def preview_contribution_email_template(request, template_name: str):
...
+                "payment_date": datetime.now(),
+                "payment_amount": 500,
+                "payment_status": "paid",
+            },
+            {
+                "payment_date": datetime.now(),
+                "payment_amount": 500,
+                "payment_status": "refunded",
+            },
+        ],
```
benjaminEwhite: Should use `BillingHistory` typed dict here.

conftest.py
```diff
@@ -293,6 +293,44 @@ def annual_contribution(live_donation_page):
...
+        created=then,
+    )
+    for x in (then, then + datetime.timedelta(days=30)):
+        PaymentFactory(
+            created=x,
+            contribution=contribution,
+            amount_refunded=contribution.amount,
+            gross_amount_paid=contribution.amount,
+            net_amount_paid=0,
+        )
```
benjaminEwhite, line 311: Why create two refunded payments?  Should be one that is normal and one that's refund, no?

benjaminEwhite (changes requested):
Have some requests for changes. Also, noting that there were several failing tests at time of review.

apps/emails/views.py
```diff
@@ -49,6 +49,18 @@ def preview_contribution_email_template(request, template_name: str):
...
+            {
+                "payment_date": datetime.datetime.now(datetime.timezone.utc),
+                "payment_amount": 500,
+                "payment_status": "paid",
+            },
+            {
+                "payment_date": datetime.datetime.now(datetime.timezone.utc),
+                "payment_amount": 500,
+                "payment_status": "refunded",
+            },
```
benjaminEwhite, line 62: Should use `BillingHistory` typed dict here.

apps/contributions/views.py
GuiMend: Ok. But I still need a flag to indicate if the billing history should show or not.

conftest.py
GuiMend: Sure, I'm not familiar with how the data should actually be. I was just creating a mock with refunded payment

apps/contributions/tests/test_models.py
```diff
@@ -550,16 +578,22 @@ def handle_thank_you_email_contribution(self, request):
...
+    ):
         """Show that when org configured to have NRE send thank you emails, send_templated_email gets called with expected args."""
         settings.CELERY_TASK_ALWAYS_EAGER = True
         (org := contribution.revenue_program.organization).send_receipt_email_via_nre = send_receipt_email_via_nre
         org.save()
         send_thank_you_email_spy = mocker.spy(send_thank_you_email, "delay")
 
         mocker.patch("apps.contributions.models.Contributor.create_magic_link", return_value="fake_magic_link")
-        contribution.handle_thank_you_email()
+        contribution.handle_thank_you_email(show_billing_history)
```
benjaminEwhite, line 592: Better to be explicit here since tests serve as documentation:

```suggestion
        contribution.handle_thank_you_email(send_billing_history=show_billing_history)
```

apps/contributions/models.py
```diff
@@ -48,6 +49,12 @@ class ContributionStatusError(Exception):
     pass
 
 
+class BillingHistory(TypedDict):
```
benjaminEwhite, line 52: Can we call this `BillingHistoryItem` or something similar since it refers to individual items that together comprise an entire billing history?

apps/contributions/models.py
```diff
@@ -541,19 +551,43 @@ def cancel(self):
...
+        billing_history = [
+            {
+                "payment_date": convert_to_timezone_formatted(payment.transaction_time, "America/New_York"),
+                "payment_amount": (
+                    self.format_amount(payment.amount_refunded)
+                    if payment.amount_refunded
+                    else self.format_amount(payment.gross_amount_paid)
+                ),
+                "payment_status": "Paid" if payment.amount_refunded == 0 else "Refunded",
+            }
```
benjaminEwhite, line 581: Shouldn't this use the `BillingHistory` typed dict?

apps/contributions/models.py
```diff
@@ -541,19 +551,43 @@ def cancel(self):
...
+            for payment in payments
+            if isinstance(payment, Payment)
```
benjaminEwhite, line 583: This should be guaranteed by virtue of calling `.payment_set.all()`. That will be empty or else it will have instances of payments.

```suggestion
```

apps/contributions/tests/test_models.py
```diff
@@ -331,6 +332,33 @@ class TestContributionModel:
...
+        expected_value = [
+            {
+                "payment_date": convert_to_timezone_formatted(payment.transaction_time, "America/New_York"),
+                "payment_amount": (
+                    contribution_billing_history.format_amount(payment.amount_refunded)
+                    if payment.amount_refunded
+                    else contribution_billing_history.format_amount(payment.gross_amount_paid)
+                ),
+                "payment_status": "Paid" if payment.amount_refunded == 0 else "Refunded",
+            }
```
benjaminEwhite, line 353: Use the typed dict here too.

apps/contributions/models.py
```diff
@@ -541,19 +551,43 @@ def cancel(self):
...
 
-    def handle_thank_you_email(self):
+    def handle_thank_you_email(self, send_billing_history: bool = False):
         """Send a thank you email to contribution's contributor if org is configured to have NRE send thank you email."""
         logger.info("`Contribution.handle_thank_you_email` called on contribution with ID %s", self.id)
         if (org := self.revenue_program.organization).send_receipt_email_via_nre:
             logger.info("Contribution.handle_thank_you_email: the parent org (%s) sends emails with NRE", org.id)
             data = make_send_thank_you_email_data(self)
+            if send_billing_history:
+                data["billing_history"] = self.get_billing_history()
```
benjaminEwhite, line 561: I think i'd prefer to always send a billing history in the template, and introduce a `send_billing_history` flag for the templates.  This would guarantee referential integrity of  'billing_history' in the template.

apps/contributions/tests/test_models.py
```diff
@@ -331,6 +332,33 @@ class TestContributionModel:
...
+            for payment in contribution_billing_history.payment_set.all()
+        ]
+
+        if contribution_billing_history.interval.value is ContributionInterval.ONE_TIME.value:
+            assert billing_history == []
```
benjaminEwhite, line 358: As mentioned in earlier review, I think that one time contributions should still have a billing history. They will typically have a single payment,  and occasionally may have refunds.

apps/contributions/tests/test_models.py
```diff
@@ -550,16 +578,22 @@ def handle_thank_you_email_contribution(self, request):
...
         (org := contribution.revenue_program.organization).send_receipt_email_via_nre = send_receipt_email_via_nre
         org.save()
         send_thank_you_email_spy = mocker.spy(send_thank_you_email, "delay")
 
         mocker.patch("apps.contributions.models.Contributor.create_magic_link", return_value="fake_magic_link")
-        contribution.handle_thank_you_email()
+        contribution.handle_thank_you_email(show_billing_history)
         expected_data = make_send_thank_you_email_data(contribution)
+        if show_billing_history:
+            expected_data["billing_history"] = contribution.get_billing_history()
```
benjaminEwhite, line 595: Given earlier comment, think this test should not have conditionality here around whether show billing history is true.

apps/contributions/views.py
```diff
@@ -674,7 +674,7 @@ def send_contribution_receipt(self, request, pk=None, contribution_id=None) -> R
             contribution = self.exclude_hidden_statuses(contributor.contribution_set).get(pk=contribution_id)
         except Contribution.DoesNotExist:
             return Response({"detail": "Contribution not found"}, status=status.HTTP_404_NOT_FOUND)
-        contribution.handle_thank_you_email()
+        contribution.handle_thank_you_email(contribution.interval != ContributionInterval.ONE_TIME.value)
```
benjaminEwhite, line 677:
```suggestion
        contribution.handle_thank_you_email(send_billing_history=contribution.interval != ContributionInterval.ONE_TIME.value)
```

benjaminEwhite, line 677: I still think we should be sending back billing history in case of one-time. See earlier comments.

apps/emails/templates/nrh-default-contribution-confirmation-email.html
```diff
@@ -84,3 +112,49 @@ <h1 class="thank-you-title inner-center">Thank You for Your Contribution!</h1>
 {% endblock header %}
 
 {% block timestamp_label %}Received{% endblock timestamp_label %}
+
+{% block billing_history %}
+  {% if billing_history %}
```
benjaminEwhite, line 117: If going with my suggestion around having flag for `show_billing_history` and always sending billing history data even if we don't show in case of one-time, this would probably become (assuming we don't want to show a headers only table in case of show_billing_history being true and empty billing history being passed)

```
  {% if billing_history and show_billing_history %}
```

apps/contributions/tests/test_models.py
benjaminEwhite, line 358: Still think we want to get back a billing history here and handle view-layer display conditionality elsewhere.

benjaminEwhite (changes requested):
Had some comments and requests.

apps/contributions/models.py
```diff
@@ -304,6 +311,9 @@ def formatted_amount(self) -> str:
         currency = self.get_currency_dict()
         return f"{currency['symbol']}{f'{self.amount / 100:.2f}'} {currency['code']}"
 
+    def format_amount(self, amount: int, symbol="$", code="USD") -> str:
```
benjaminEwhite, line 314: Didn't notice this in previous reviews because the `formatted_amount` method didn't appear in diff here and only saw in VSCode.

I think it would make sense to change `format_amount` to a `@staticmethod` here.  

```
@staticmethod
def format_amount(amount: int, symbol="$", code="USD") -> str:
```


Then refactor `formatted_amount` so it uses `Contribution.format_amount`. 

```
    @property
    def formatted_amount(self) -> str:
        currency = self.get_currency_dict()
        return self.format_amount(self.amount, currency["symbol"], currency["code"])
```

This allows us to DRY this code up.

benjaminEwhite (changes requested):
@GuiMend had one final (99% confident on that 😄 ) round of suggestions around the type hinting issue, DRYing up format amount, and using an Enum class to capture the statuses that BillingHistoryItem can have.

apps/emails/views.py
```diff
@@ -49,6 +50,18 @@ def preview_contribution_email_template(request, template_name: str):
         "rp_name": rp.name,
         "style": rp_style,
         "default_contribution_page_url": rp.default_donation_page.page_url if rp.default_donation_page else None,
+        "billing_history": [
```
benjaminEwhite, line 53: Should add `show_billing_history: True` here so can see billing history in template preview.

apps/emails/tasks.py
```diff
@@ -1,6 +1,10 @@
+# Needed for postponing evaluation of annotations (because of "if TYPE_CHECKING")
```
benjaminEwhite, line 1: Is this actually needed? I was able to get type hints to work as expected with no complaints from VSCode wihtout doing import from future. But if you're finding it's required @GuiMend that's fine (and is what was recommended in the reddit post I shared).

benjaminEwhite (changes requested):
@GuiMend my apologies and thanks for your patience but another round (hopefully last) here.

- Think we need to back out of the enum change, but we can discuss reason why in Slack huddle.
- Added request to include `"show_billing_history": True` in preview email template so we can see the billing history there (FWIW, that's what led me to discover the problem with the enum for status).
- Question about if we actually need to import annotations from __future__

apps/contributions/models.py
```diff
@@ -52,6 +54,17 @@ class ContributionStatusError(Exception):
...
+    PAID = "Paid"
+    REFUNDED = "Refunded"
+
+
+class BillingHistoryItem(TypedDict):
+    payment_date: datetime.datetime
+    payment_amount: int
+    payment_status: BillingHistoryItemStatus
+
+
```
benjaminEwhite, line 67: Apologies, but after using email template preview locally, looks like this was bad advice. Problem is that payment_status won't be a string value, but an enum object if we take this approach, and we'd need to do item.PAID.value (for instance).

Instead we could consider doing 

```
# imports at top of module:

from typing import  Literal

...

PaymentStatus = Literal["Paid", "Refunded"]

class BillingHistoryItem(TypedDict):
    payment_date: datetime.datetime
    payment_amount: int
    payment_status: BillingHistoryItemStatus
```

That gets us typing, but doesn't solve for the issue of being consistent about statuses (i.e., the place i pointed out earlier that changed the capitalization).

Think might make most sense to just remove enum, not use literal here, and just make sure we're consistent with spelling.

GuiMend: I like the Literal approach Ben. It doesn't enforce the correct string, but it shows an autocomplete that facilitates the correct usage.
image

apps/emails/views.py
benjaminEwhite: @GuiMend I see this acknowledged, but no change. Maybe you missed pushing up a commit?

benjaminEwhite (changes requested):
@GuiMend looks like you might have forgotten to push up a commit? It's for the one-remaining open conversation (around show_payment_history).

apps/emails/views.py
GuiMend: It's in the above line, so your comment didn't get the `Outdated` flag
image

benjaminEwhite (approved):

PR #1553, opened 2024-06-26
DEV-4924 Duplicate Settings

njharman:
#### What's this PR do?
Delete duplicate settings
Format settings into sections a bit more (double #, double line between)
Moves some items around to "correct section"
Reuses HOST_MAP instead of getting it anew from environment.

duplicates are bad
so many setting, organization helps find/maybe not introduce duplicates in future.
HOST_MAP is only used in this file, and it wasn't being used!

No. What to test?

No.

N/A. 

N/A

No.

No.

DEV-4924

No.

### DIFF
--- a/revengine/settings/base.py
+++ b/revengine/settings/base.py
@@ -49,7 +49,8 @@ SECURE_CONTENT_TYPE_NOSNIFF = True
 X_FRAME_OPTIONS = "SAMEORIGIN"
 SECURE_REFERRER_POLICY = "strict-origin-when-cross-origin"
 
-# Internationalization
+
+## Internationalization
 # https://docs.djangoproject.com/en/3.2/topics/i18n/
 LANGUAGE_CODE = "en-us"
 TIME_ZONE = "UTC"
@@ -73,7 +74,8 @@ class CurrencyDict(TypedDict):
 
 CURRENCIES: CurrencyDict = {"USD": "$", "CAD": "$"}
 
-# Static files (CSS, JavaScript, Images)
+
+## Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/3.2/howto/static-files/
 STATIC_ROOT = BASE_DIR / "public" / "static"
 STATIC_URL = "/static/"
@@ -189,7 +191,7 @@ TEMPLATES = [
 ]
 
 
-# Database
+## Database
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
 # https://docs.djangoproject.com/en/3.2/ref/settings/#databases
@@ -216,7 +218,7 @@ if os.getenv("DATABASE_URL"):
     DATABASES["default"].update(db_from_env)
 
 
-# Password validation
+## Password validation
 # https://docs.djangoproject.com/en/3.2/ref/settings/#auth-password-validators
 AUTH_PASSWORD_VALIDATORS = [
     {
@@ -251,17 +253,13 @@ STRIPE_TRANSACTIONS_IMPORT_CACHE_TTL = int(
 
 REDIS_URL = os.getenv("REDIS_TLS_URL", os.getenv("REDIS_URL", "redis://redis:6379"))
 
-
 CACHE_HOST = REDIS_URL
 CONNECTION_POOL_KWARGS = {}
 if CACHE_HOST.startswith("rediss"):
     import ssl
 
     # See: https://github.com/mirumee/saleor/issues/6926
-    CONNECTION_POOL_KWARGS = {
-        "ssl_cert_reqs": ssl.CERT_NONE,
-    }
-
+    CONNECTION_POOL_KWARGS["ssl_cert_reqs"] = ssl.CERT_NONE
 
 base_cache_config = {
     "BACKEND": "django_redis.cache.RedisCache",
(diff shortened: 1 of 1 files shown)


revengine/settings/base.py
```diff
@@ -452,6 +454,13 @@ class CurrencyDict(TypedDict):
 REVERSION_COMPARE_IGNORE_NOT_REGISTERED = True
 
 
+### django-healthcheck Settings
```
njharman, line 457: When merging main, this was duplicated with another setting (cut below). I chose to move new one here,  to group django settings together.

revengine/settings/base.py
```diff
@@ -479,6 +488,15 @@ class CurrencyDict(TypedDict):
     "charge.succeeded",
 ]
 
+RETRIEVED_STRIPE_ENTITY_CACHE_TTL = 60 * 3  # Three minutes
```
njharman, line 491: Moved from end of file to this Stripe settings "section"

revengine/settings/base.py
```diff
@@ -581,13 +595,7 @@ class CurrencyDict(TypedDict):
...
     "EMAIL_DEFAULT_TRANSACTIONAL_SENDER", "News Revenue Engine <no-reply@fundjournalism.org>"
 )
-
-# Transactional Email
-EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
-
-EMAIL_DEFAULT_TRANSACTIONAL_SENDER = os.getenv(
-    "EMAIL_DEFAULT_TRANSACTIONAL_SENDER", "News Revenue Engine <no-reply@fundjournalism.org>"
-)
+SWITCHBOARD_ACCOUNT_EMAIL = os.getenv("SWITCHBOARD_ACCOUNT_EMAIL", None)
```
njharman, line 598: Moved to email settings section from end of file.

revengine/settings/base.py
```diff
@@ -609,32 +618,28 @@ class CurrencyDict(TypedDict):
...
 MAILCHIMP_CLIENT_SECRET = os.getenv("MAILCHIMP_CLIENT_SECRET", None)
 
 # see https://mailchimp.com/developer/release-notes/message-search-rate-limit-now-enforced/#:~:text=We're%20now%20enforcing%20the,of%20the%20original%2020%20requests.
 MAILCHIMP_RATE_LIMIT_RETRY_WAIT_SECONDS = 60
 
-# Host map for client custom hostnames. This should be a JSON-encoded dictionary
-# of { "customhostname.org": "rp-slug" } values.
+RP_MAILCHIMP_LIST_CONFIGURATION_COMPLETE_TOPIC = os.getenv("RP_MAILCHIMP_LIST_CONFIGURATION_COMPLETE_TOPIC")
 
-logger = logging.getLogger(f"{DEFAULT_LOGGER}.{__name__}")
```
njharman, line 629: Moved to inside except block as this is only place it is used.

benjaminEwhite: I'm on the fence about this. If in future we decide to log somewhere else, we'd ideally define at module scope and reuse. but if import is buried in this except, only way we'd know to dry up is if we remember to look specifically at this exception path code.

Don't see much downside to having logger defined in module scope -- especially since this is the implicit pattern in most modules in this repo.

revengine/settings/base.py
```diff
@@ -655,17 +660,9 @@ class CurrencyDict(TypedDict):
     "SENTRY_ENABLE_FRONTEND": SENTRY_ENABLE_FRONTEND,
     "SENTRY_DSN_FRONTEND": SENTRY_DSN_FRONTEND,
     "STRIPE_CLIENT_ID": os.getenv("SPA_ENV_STRIPE_CLIENT_ID"),
-    "HOST_MAP": os.getenv("HOST_MAP", "{}"),
+    "HOST_MAP": HOST_MAP,
```
benjaminEwhite, line 663: In status quo, this value is set to string, however with this change it will be set to JSON. I don't know if this breaks custom domain mapping or not, but the end result is that HOST_MAP in header of SPA index.html ends up with different value than it does in status quo here.

Here's what it looks like in test:


Screenshot 2024-07-15 at 10 34 13 AM


And here's what it looks like in review app for this PR:

Screenshot 2024-07-15 at 10 38 37 AM

revengine/settings/base.py
njharman: moved to under logging config with comment 42e5c0c70

I don't think there should be any logging in settings. Same settings where logging is configured. esp django style settings where dev/test/prod settings will import base and possibly modify logging settings (too late as logging already been done.

revengine/settings/base.py
njharman: Good catch. Regressed with comment 42e5c0c70

revengine/settings/base.py
```diff
@@ -189,7 +191,7 @@ class CurrencyDict(TypedDict):
 ]
 
 
-# Database
+## Database
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
```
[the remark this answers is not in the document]
njharman: Searched code for djangoproject.com urls and updated to 4.2 9342754fa

x110dc (approved):
lgtm

PR #1562, opened 2024-07-03
DEV-4910_dual_revenue_program_queries

njharman:
#### What's this PR do?
Isolate the dual RP source logic to/in Contribution model. Also rename filtered -> filter for consistency.

I exceeded "timeboxing" trying to get fix_incident_2445.py to work with queryset method. So, skipped it/left unchanged.

This was difficult to isolate because of difference in Python and SQL worlds.
In Python Foreign Relations are "magical" in that they're objects. In SQL Foreign Relations are only id. Making a queryset that hid the dual rp logic was no problem. But downstream users expected the rp to be object, not the id query builder returned. The partial solution was to make multiple tailored methods, `.with_foo()`

In addition, the existing @properties on contribution conflicted [too much risk/effort to change all that]. The partial solution was to use `.alias()`, whose values only exists in query builder.

DRY. It's potentially tricky to remember to dual access everywhere it's needed. Also, simplying complex SQL to method call reduces chance for errors. 

No. Not sure how to explain what needs to test

No.

No. Other than renames tests were purposefully untouched to show existing (tested) behavior was not borkened

N/A

Not, real happy with this implementation.  Spent a lot of frustrating time trying different solutions. Final code probably rushed.  

The filter_queryset_for_contributor returning list and not queryset unlike all other filter_* will probably trip people up when they try to chain queryset methods.  [Caching is why we can't have nice things. I bet someone has written a caching layer that returns queryset compatible things.]

Also really want to reorder methods in Contribution model to like ones are together. but that creates diff visual noise.

N/A

[DEV-4910]

No.

No.


apps/contributions/models.py
```diff
@@ -124,6 +125,32 @@ class ContributionQuerySet(models.QuerySet):
         ContributionStatus.REJECTED,
     ]
 
+    def with_revenue_program_id(self):
+        """Alias revenue_program_id as "revenue_program".
+
+        Alias (which is only in SQL) as Contribuition had existing property "revenue_program" which handles dual revenue_program source.
```
benjaminEwhite, line 131:
```suggestion
        Alias (which is only in SQL) as Contribution had existing property "revenue_program" which handles dual revenue_program source.
```

benjaminEwhite (approved):
Found one typo in code comment, but otherwise looks good (and no need for me to re-review on that coment).with

apps/contributions/models.py
njharman: 71727514c

apps/contributions/models.py
```diff
@@ -124,6 +125,32 @@ class ContributionQuerySet(models.QuerySet):
...
+    def with_organization_id(self):
+        """Alias organization_id as "organization_id"."""
+        return self.alias(
+            organization_id=Coalesce("_revenue_program__organization", "donation_page__revenue_program__organization")
+        )
+
+    def with_stripe_account(self):
+        """Annotate stripe_account_id as "stripe_account".
+
+        stripe_account even though it is the id and not object instead of *_id because Contribuition had existing
```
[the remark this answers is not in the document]
njharman: 71727514c

x110dc (approved):

PR #1582, opened 2024-07-25
DEV-4814: Add PATCH endpoint for contribution to change amount

GuiMend:

n/a
#### What's this PR do?
- Enables changes to contribution amount with PATCH
- Modifies Stripe's Subscription amount to reflect new value

- Enable user to change donation amount
- n/a. I don't think it's testable without the FE.
- no
- yes
- n/a
- no
- no
- https://news-revenue-hub.atlassian.net/browse/DEV-4814
- no
https://news-revenue-hub.atlassian.net/browse/DEV-5005


README.md
```diff
@@ -160,17 +160,29 @@ Then, in Django-admin, create an Organization for that connected stripe account
...
+The above command will execute and return something like: `> Ready! You are using Stripe API Version [2020-08-27]. Your webhook signing secret is whsec_9xx (^C to quit)`.
+With this secret, all you have to do is add it to your `.envrc`:
+
+```
+echo "export STRIPE_WEBHOOK_SECRET_CONTRIBUTIONS=whsec_9xx" >> .envrc
+```
+
+Obs 1: Make sure you have revengine and celery running. You can do that by having 2 terminals running. One with `make run-redis` and the other with `make start-celery`
+
+Obs 2: You will probably need to add `"https://*.stripe.com/"` to `CSRF_TRUSTED_ORIGINS` list
```
benjaminEwhite, line 185: I searched for this in repo, and only place it comes up is README.md. So i guess would need to add `CSRF_TRUSTED_ORGINS` to local settings.

apps/contributions/models.py
```diff
@@ -264,6 +264,12 @@ class Contribution(IndexedTimeStampedModel):
 
     objects = ContributionManager.from_queryset(ContributionQuerySet)()
 
+    ACTIVE_SUBSCRIPTION_STATUSES = (
+        ContributionStatus.PAID,
+        ContributionStatus.FAILED,
+        ContributionStatus.REFUNDED,
+    )
+
```
benjaminEwhite, line 272: It's possible for the NRE contribution status to be out of sync with the Stripe subscription status.  I think we probably want to be looking at the Stripe sub statu to be safe.

apps/contributions/models.py
```diff
@@ -1164,6 +1170,78 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
+    def update_amount_for_subscription(self, amount: int) -> None:
+        """Update Stripe's Subscription Item amount to the new value.
+
+        If it's a recurring subscription, update the amount of the next payment
+        without proration (paying difference of existing month).
+        """
+        if amount <= 0:
+            raise ValueError("Amount value must be greater than 0")
+        if self.status not in self.ACTIVE_SUBSCRIPTION_STATUSES:
+            raise ValueError("Cannot update amount for inactive subscription")
```
benjaminEwhite, line 1182: In line with earlier comment, I think we should be checking the underlying Stripe subscription's status here.

apps/contributions/serializers.py
```diff
@@ -900,18 +900,23 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
     card_owner_name = serializers.CharField(read_only=True, allow_blank=True)
     payments = PortalContributionPaymentSerializer(many=True, read_only=True, source="payment_set")
     provider_payment_method_id = serializers.CharField(write_only=True, required=False)
+    amount = serializers.IntegerField(required=False)
```
benjaminEwhite, line 903: Probably this is the place to do the validation around minimum amount.

apps/contributions/serializers.py
```diff
@@ -900,18 +900,23 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
...
+    def update(self, instance: Contribution, validated_data) -> Contribution:
         if validated_data:
             if provider_payment_method_id := validated_data.get("provider_payment_method_id", None):
                 instance.update_payment_method_for_subscription(
                     provider_payment_method_id=provider_payment_method_id,
                 )
+            if amount := validated_data.get("amount", None):
+                instance.update_amount_for_subscription(
+                    amount=amount,
+                )
```
benjaminEwhite, line 919: This block is not tested.

apps/contributions/models.py
```diff
@@ -1164,6 +1170,78 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
+            items = stripe.SubscriptionItem.list(subscription=sub_id, stripe_account=self.stripe_account_id)
+        except StripeError:
+            logger.exception(
+                "Encountered a Stripe error while trying to fetch subscription items on sub_id: %s",
+                sub_id,
+            )
+            raise
+
+        if len(items["data"]) != 1:
+            raise ValueError("Subscription should have only one item")
```
benjaminEwhite, line 1202: This path is not tested.

benjaminEwhite (changes requested):
Left some comments and requests.

Additionally, I think we should have a test that shows amount update behavior via API layer, not only via model layer. In particular, we should show that if you send an amount of `1` in patch request, that ends up as `100` in the contribution.amount.

README.md
GuiMend: Yes, this is added to our `local.py`. I updated `local.example.py` with this

apps/contributions/models.py
```diff
@@ -1164,6 +1170,78 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
+
+        try:
+            logger.info(
+                "Updating Stripe Subscription's %s (item %s), amount to %s",
+                sub_id,
+                item["id"],
+                amount,
+            )
+
+            stripe.Subscription.modify(
```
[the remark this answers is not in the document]
nrh-cklimas: @x110dc should sign off on this, but my opinion is that the backend should use the amount as it's provided by the SPA, same as what we do with initial contributions. This will be more an issue when we do the SPA side of this, as we'd (probably) want the new amount to reflect the existing fee setting. Changing pay fees/don't pay fees is also a separate story.

apps/contributions/models.py
```diff
@@ -1164,6 +1172,80 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
+                sub_id,
+            )
+            raise
+
+        if len(items["data"]) != 1:
+            raise ValueError("Subscription should have only one item")
+
+        item = items["data"][0]
+        metadata = self.contribution_metadata
+        metadata["donor_selected_amount"] = amount
```
njharman, line 1208: Is assigning to `metadata `an attempt to not modify self.contribution_metadata. But assignment creates a "reference", not a copy. [it also appears no as changes to `contribution_metadata `are saved below]. Is also not saving "space" as metadata is not used often, once I believe.

Unnecessary, confusing, and adds another entry in namespace to keep track of/cause errors/etc.

`self.contribution_metadata["donor_selected_amount"] = amount`

apps/contributions/models.py
```diff
@@ -1164,6 +1172,80 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
+                        },
+                    }
+                ],
+                metadata=metadata,
+                stripe_account=self.stripe_account_id,
+                proration_behavior="none",
+            )
+        except StripeError:
+            logger.exception(
+                "Encountered a Stripe error while trying to update payment method for subscription on contribution %s",
```
njharman, line 1238: Would it be useful to log more data, amount, other parameters to `stripe.Subscription.modify()`?

I believe Sentry will capture all the locals(). So, maybe redundant?

apps/contributions/models.py
```diff
@@ -1164,6 +1172,80 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
 
             raise
 
+    def update_subscription_amount(self, amount: int) -> None:
+        """Update Stripe's Subscription Item amount to the new value.
+
+        If it's a recurring subscription, update the amount of the next payment
+        without proration (paying difference of existing month).
+        """
+        if amount < 100:
```
njharman, line 1181: REVENGINE_MIN_AMOUNT inplace of literal `100`?

apps/contributions/models.py
njharman: It is now.

apps/contributions/models.py
```diff
@@ -1164,6 +1172,80 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
 
             raise
 
+    def update_subscription_amount(self, amount: int) -> None:
+        """Update Stripe's Subscription Item amount to the new value.
+
+        If it's a recurring subscription, update the amount of the next payment
+        without proration (paying difference of existing month).
```
njharman, line 1179: Does Stripe just handle making it the "next payment" for all three; billing_date is before today, today and after today?

apps/contributions/models.py
njharman, line 1181: The serializer validates amount is between REVENGINE_MIN_AMOUNT and STRIPE_MAX_AMOUNT. Should we not do same here?

njharman (changes requested):

apps/contributions/models.py
```diff
@@ -264,6 +264,14 @@ class Contribution(IndexedTimeStampedModel):
...
     objects = ContributionManager.from_queryset(ContributionQuerySet)()
 
+    INACTIVE_SUBSCRIPTION_STATUSES = (
+        "paused",
+        "incomplete_expired",
+        "past_due",
+        "canceled",
+        "unpaid",
+    )
+
```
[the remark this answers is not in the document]
GuiMend: Actually @njharman , I'd like to confirm these statuses as I don't know what indicates an "active" subscription. Would you know? I read the documentation and assumed these were the statuses that seemed inactive / amount updates didn't make sense. (But it's my opinion)

These are the available options:
image

apps/contributions/models.py
GuiMend: I did this because in line `1232` I assigned the metadata in the Stripe subscription payload:
```python
...
metadata=metadata,
stripe_account=self.stripe_account_id,
proration_behavior="none",
```

Would you prefer that I did this instead?
```python

# and later
metadata= self.contribution_metadata
```

apps/contributions/models.py
GuiMend: I'm not sure @njharman , but in other places that we log stripe.Subscription.modify errors, it's in the same format as this one.

apps/contributions/models.py
GuiMend: Yes. Stripe will make the next payment with the updated amount.

apps/contributions/models.py
njharman: Yes. It's minor.

Each time is minor, but (in reviews and when coding) I'm always looking to reduce name space "pollution", vertical space and the "vertical distance" something has to be remembered.

The more things in namespace (assignments to local variables, from x import y, etc)

- The more chances there are for bugs; typos, name collisions 
- The more items that need to be kept in ones head when understanding and refactoring code (this is also main reason for less vertical distance, such as code blocks that are ~ one screenful and assigning things close to were they are used.

apps/contributions/models.py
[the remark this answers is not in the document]
njharman: Oh, opinions are dangerous ;) 

I also only have an opinion. Really should have been defined in AC. Should ask whoever wrote the AC to add clarification (to AC) what inactive means. Probably @ them what you posted above in a comment on this ticket. [ticket comments are more visible and permanent than PR ones]

If you don't know who to ask, ping DC.

apps/contributions/models.py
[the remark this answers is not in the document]
njharman: @x110dc eh?

njharman (changes requested):

apps/contributions/models.py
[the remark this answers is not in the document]
GuiMend: Added a comment there @njharman. Thanks

apps/contributions/models.py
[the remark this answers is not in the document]
x110dc: Yes, I agree that it should be consistent with the way it happens with the initial contribution.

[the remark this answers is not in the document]
GuiMend: This means that BE stays the way it is now, and we will address this in the FE.

apps/contributions/models.py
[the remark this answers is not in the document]
GuiMend: Daniel confirmed in JIRA that the only active status is the one with value `active`, all the others are considered `inactive`

njharman (changes requested):
build failing

njharman (approved):

nrh-cklimas (approved):

PR #1586, opened 2024-07-29
DEV-5059: Add first_payment_date method to Contribution model

nrh-cklimas:

n/a

#### What's this PR do?

- Adds a `first_payment_date` field to the Contribution model. It is required to have a value.
- Populates this field with `created` during the initial migration.
- Overrides the `save` method to set this field to `created` when an object is initially created.

Allows us to show correct dates for backfilled contributions in the contributor portal.

Yes.

No.

Yes.

n/a

I am unsure about testing around the save() override. It would be nice to assert that `super().save()` is getting called correctly by it, but not sure how to go about it.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5059

No.

- https://news-revenue-hub.atlassian.net/browse/DEV-5064
- https://news-revenue-hub.atlassian.net/browse/DEV-5065


apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
```diff
@@ -0,0 +1,34 @@
...
+
+from django.db import migrations, models
+
+
+def set_first_payment_date_default(apps, schema_editor):
+    # cribbed from
+    # https://stackoverflow.com/questions/29787853/django-migrations-add-field-with-default-as-function-of-model
+    for contribution in apps.get_model("contributions", "contribution").objects.all().iterator():
+        contribution.first_payment_date = contribution.created
+        contribution.save()
```
njharman, line 11: I'm triggered by loop over all rows with DB save.

Recommend some method of bulk update; https://docs.djangoproject.com/en/5.0/ref/models/querysets/#bulk-update

apps/contributions/tests/test_models.py
```diff
@@ -1802,6 +1802,10 @@ def test_exclude_paymentless_canceled(self):
         assert contribution.payment_set.count() == 0
         assert Contribution.objects.exclude_paymentless_canceled().count() == 0
 
+    def test_first_billing_date_defaults_to_created(self):
+        contribution = ContributionFactory()
+        assert contribution.first_payment_date == contribution.created
```
njharman, line 1807: This is testing ContributionFactory more than Contribution. If ContributionFactory now (or is changed in future) to set first_payment_date it's prone to false positive.  Contribution it self may not set default.

Also, should not first_payment_date be added to ContributionFactory?

apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
```diff
@@ -0,0 +1,34 @@
...
+    operations = [
+        migrations.AddField(
+            model_name="contribution",
+            name="first_payment_date",
+            field=models.DateTimeField(null=True),
+            preserve_default=False,
+        ),
+        # Doesn't need to do anything to reverse it because we'd remove the field.
+        migrations.RunPython(set_first_payment_date_default, migrations.RunPython.noop),
+        migrations.AlterField(
```
njharman, line 29: Consider constraint that first_payment_date must be >= created?

apps/contributions/models.py
```diff
@@ -679,6 +680,12 @@ def send_recurring_contribution_change_email(
             render_to_string(f"{template_name}.html", data),
         )
 
+    def save(self, *args, **kwargs):
+        """Force first_payment_date to have the same value as created if not already set."""
+        if not self.first_payment_date:
+            self.first_payment_date = self.created
+        super().save(*args, **kwargs)
```
njharman, line 687: I'm surprised this works. I didn't think self.created got set until after save(). It also won't catch if first_payment_date passed in kwargs.

I would like to see the following model tests on existing Contribution, and new Contribution() [not ContributionFactory, see below]

1. save()
2. save(first_payment_date=None)

btw, I believe agood way to cover kwargs case is [setdefault](https://docs.python.org/2/library/stdtypes.html#dict.setdefault).
```
    kwargs.setdefault("first_payment_date", self.created)
    super().save(*args, **kwargs)
```

njharman (changes requested):

apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
nrh-cklimas: Makes sense-- 2b07a7daf3dbc82239e52064f37f625177eb2178

apps/contributions/tests/test_models.py
nrh-cklimas: Is the fix here to create a `Contribution()` directly and assert on that?

apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
nrh-cklimas: Added in c812529b717a136594b097d83ea9b1174f1c212a, but double-check my work--I did this all by hand and haven't done that before.

apps/contributions/models.py
nrh-cklimas: I don't love this approach either and am open to doing it in a better way if you have a suggestion. I was looking at [Django docs that suggested preferring this override to using a signal](https://docs.djangoproject.com/en/5.0/ref/signals/#module-django.db.models.signals) but it felt like there has to be a better way here. If not, can do the suggestions you had here.

njharman: Believe doing it in save() is fine. Mostly interested in tests to verify it works way we think it does (and to proof against future changes breaking it).

apps/contributions/tests/test_models.py
njharman: yes.

apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def set_first_payment_date_default(apps, schema_editor):
+    # cribbed from
+    # https://stackoverflow.com/questions/29787853/django-migrations-add-field-with-default-as-function-of-model
+    Contribution = apps.get_model("contributions", "contribution")
+    all_contributions = Contribution.objects.all()
+    for contribution in all_contributions:
+        contribution.first_payment_date = contribution.created
+    Contribution.objects.bulk_update(all_contributions, ["first_payment_date"])
```
njharman, line 13: depending how many rows in production, batch_size is probably appropriate. batch_size = 1,000 or 2,000

njharman (changes requested):

apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
nrh-cklimas: @njharman realized in the stacked PR we can't do a constraint on the field because `created` will be after the first payment occurred for migrated contributions. e.g. We are creating contributions today that correspond to contributions that occurred in legacy last year. I backed this out in 70ee13f31bc78757badb7a04411cd4227b7d527b

apps/contributions/tests/test_models.py
nrh-cklimas: I'm trying this approach but the assert fails if I don't save the instance, which makes sense given the implementation approach. However, if I do try to save it, it complains about a lot of validation errors... and to fix that, it feels like I would have to essentially re-implement ContributionFactory.

What if we do it like this?

```python
    def test_first_billing_date_defaults_to_created(self):
        contribution = ContributionFactory(first_payment_date=None)
        assert contribution.first_payment_date == contribution.created
```

apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
nrh-cklimas: :+1: 8b1634236c6724d2c32c8fb655327f4b138cfd88

apps/contributions/models.py
nrh-cklimas: If I try the code snippet you suggested, I get this error in tests:
```
TypeError: Model.save() got an unexpected keyword argument 'first_payment_date'
```

I think this is because [save() doesn't let you set model fields](https://docs.djangoproject.com/en/5.0/ref/models/instances/#saving-objects) in the call? I might have this totally wrong.

apps/contributions/migrations/0012_DEV_5059_contribution_first_payment_date.py
njharman: >  created will be after the first payment occurred for migrated contributions. 

Well, that seems like should set created on migration. But investigating what that'll break is way out of scope so 

+!

apps/contributions/models.py
njharman: No. I'm wrong... I think I've made same wrong assumption before (the *args, **kwargs) meaning it optionally takes fields.

Sorry for wasting your time.  my #2 test case above is N/A. #1 is discussed elsewhere.

apps/contributions/tests/test_models.py
njharman: The worry I have is that by using ContributionFactory, we are testing **test code** and not **production code**.  Factory would be and is fine when we just need a Contribution to test something else. But this test, is testing that Contributions are created correctly, not that ContributionFactory returns correct Contribution.

As a contrived example; could delete you're changes to def save()  and add to ContributionFactory [I forget the exact faker syntax, but it's something like...]` first_payment_date = lazyAttribute(lambda o: o.created)`. You're obviously not gonna do that now. But future edits of `def save()` / other code might unintentionally break it in similar way where test won't fail, but production code would.

You have to decide if you think that risk large enough to justify test effort. As reviewer, I'd not disagree if you decided it was. 

How are Contributions instantiated in production? I gave up trying to figure out by looking at code.  At some point something must be calling Contribution(). But there might be some higher level function analogous to ContributionFactory(). If all production calls to Contribution() are via that "factory"; could use that production factory for test (the assumption being that would handle all the validation errors).

njharman (approved):

apps/contributions/models.py
```diff
@@ -247,6 +247,7 @@ class Contribution(IndexedTimeStampedModel):
     provider_payment_method_id = models.CharField(max_length=255, blank=True, null=True)
     provider_payment_method_details = models.JSONField(null=True)
 
+    first_payment_date = models.DateTimeField(null=False, default=timezone.now)
```
joelfkehub, line 250: My training/experience tells me to never, ever store datetime info in anything but UTC.  Is that not "de rigeur" anymore?  I guess if we know the timezone that applies, we can still do comparisons and whatnot, but the potential of having to do careful conversions within a query to make comparisons and/or filtrations frightens me.  Am I overblowing the issue?

njharman: [just happened to notice this comment]

UTC never goes out of style ;)

Good catch.  django.timezone.now() https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.timezone.now appears to return naive or UTC depending on setting of [USE_TZ](https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-USE_TZ)

I'm reasonably sure Postgress will interpret naive as UTC. But I don't like that expected behavior can flip based on setting. Assuming my interpretation is correct. Maybe use some now() function that always returns aware UTC datetime.

apps/contributions/tests/test_views.py
```diff
@@ -1649,7 +1653,10 @@ def test_contributions_list_filter_by_rp(
         self, api_client, portal_contributor_with_multiple_contributions_over_multiple_rps
     ):
         contributor = portal_contributor_with_multiple_contributions_over_multiple_rps
-        rps = contributor.contribution_set.values_list("donation_page__revenue_program", flat=True).distinct()
+        # "Set" initial list to get unique revenue programs, than transform back to "list" for indexing
+        rps = list(
+            set(contributor.contribution_set.values_list("donation_page__revenue_program", flat=True).distinct())
```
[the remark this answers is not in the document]
joelfkehub: Also, if we make sure distinct is working properly at the database, we should not need to convert the results to `set`.

joelfkehub (changes requested):
Two comments/questions.  See conversations!

apps/contributions/tests/test_views.py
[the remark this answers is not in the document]
joelfkehub: I just noticed this is in the context of a test, so not as important, and unlikely to produce positive results given the likely scale of the test data.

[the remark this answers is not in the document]
nrh-cklimas: My gut says that this would be a premature optimization because this test is using a Pytest fixture (`portal_contributor_with_multiple_contributions_over_multiple_rps`) that has only three contributions ([definition](https://github.com/newsrevenuehub/rev-engine/blob/f1682fec9b487c6ad2b5d5decfd4b58d8bd4b8e5/apps/contributions/tests/test_views.py#L1573), [fixture that it's based on](https://github.com/newsrevenuehub/rev-engine/blob/f1682fec9b487c6ad2b5d5decfd4b58d8bd4b8e5/apps/contributions/tests/test_views.py#L1534)). But if you feel strongly about it, can look at writing this in a way that makes this fact clearer.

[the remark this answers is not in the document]
joelfkehub: If tests are documentation, and even sometimes a guide as to how one might "do things" elsewhere in the code, then I vote we change line 1658 above to this:
`contributor.contribution_set.order_by().values_list("donation_page__revenue_program", flat=True).distinct())`

That would satisfy me.

apps/contributions/models.py
joelfkehub: I'll do some quick reading of the relevant documentation and of DEV-2329 and then go create that ticket.

We can resolve this convo based on Chris's suggestion to address this in a different changeset.

apps/contributions/tests/test_views.py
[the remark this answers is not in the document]
joelfkehub: Meh.  Norm is right.  Juice, squeeze, not worth it.

joelfkehub (approved):
As discussed, we'll open another tech debt ticket to address the timezone-switch-based-on-settings issue (which is both minor and out of scope).

And, I've "stood down" on my critique of Contribution query structure within the tests given that we're filtering over a very small number of items in that test.   Address it for bonus points, if you like.

So, Approved!

x110dc (approved):
lgtm

PR #1592, opened 2024-08-07
DEV-4704: Send failed email if payment intent failed

GuiMend:

n/a
#### What's this PR do?
- Sends "Failed Contribution" email if stripe triggers event "payment_intent.failed"
- Update "generate_magic_link" to include redirect query param when passed in as prop
- Create email template for "Failed Contribution"
- Handle "redirect" in FE after validating token
- Warn user of a failed payment and send a magic link to redirect them directly to where they can update payment.
- not yet, not sure how to trigger failed payment
- yes, email template
- yes
- Failed payment email is sent to donors with a link to fix the issue
- no
- no
- https://news-revenue-hub.atlassian.net/browse/DEV-4704
- no
- no

### DIFF
--- a/apps/api/views.py
+++ b/apps/api/views.py
@@ -149,8 +149,9 @@ class RequestContributorTokenEmailView(APIView):
     throttle_classes = [ContributorRateThrottle]
 
     @staticmethod
-    def get_magic_link(domain, token, email):
-        return f"https://{domain}/{settings.CONTRIBUTOR_VERIFY_URL}?token={token}&email={quote_plus(email)}"
+    def get_magic_link(domain, token, email, next_url=None):
+        redirect = f"&redirect={quote_plus(next_url)}" if next_url else ""
+        return f"https://{domain}/{settings.CONTRIBUTOR_VERIFY_URL}?token={token}&email={quote_plus(email)}{redirect}"
 
     def post(self, request, *args, **kwargs):
         logger.info("[RequestContributorTokenEmailView][post] Request received for magic link %s", request.data)
--- a/apps/contributions/webhooks.py
+++ b/apps/contributions/webhooks.py
@@ -1,16 +1,19 @@
 import datetime
 import logging
 import operator
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from functools import cached_property, reduce
 
 from django.conf import settings
 from django.db import transaction
 from django.db.models import Q
+from django.template.loader import render_to_string
+from django.utils.safestring import mark_safe
 from django.utils.timezone import make_aware
 
 import reversion
 import stripe
+from stripe.error import StripeError
 
 from apps.contributions.models import (
     Contribution,
@@ -19,6 +22,7 @@ from apps.contributions.models import (
     Payment,
 )
 from apps.contributions.types import StripeEventData
+from apps.emails.tasks import generate_magic_link, send_templated_email
 
 
 logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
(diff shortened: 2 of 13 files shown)


njharman (approved):

apps/contributions/webhooks.py
```diff
@@ -231,6 +235,62 @@ def handle_payment_intent_canceled(self):
...
+                customer = stripe.Customer.retrieve(
+                    self.contribution.provider_customer_id,
+                    stripe_account=self.contribution.stripe_account_id,
+                )
+            except StripeError:
+                logger.exception(
+                    "Something went wrong retrieving Stripe customer for contribution with ID %s",
+                    self.contribution.id,
+                )
+                return
```
njharman, line 258: Uncovered

njharman (changes requested):

apps/api/views.py
```diff
@@ -149,8 +149,9 @@ class RequestContributorTokenEmailView(APIView):
     throttle_classes = [ContributorRateThrottle]
 
     @staticmethod
-    def get_magic_link(domain, token, email):
-        return f"https://{domain}/{settings.CONTRIBUTOR_VERIFY_URL}?token={token}&email={quote_plus(email)}"
+    def get_magic_link(domain, token, email, next_url=None):
+        redirect = f"&redirect={quote_plus(next_url)}" if next_url else ""
```
nrh-cklimas, line 153: Something for backend folks to weigh in on: do we want stronger protection on this? Like raise an exception if it looks like a full URL and not an absolute one without domain? The PR relies on react-router not allowing redirects offsite, so there's the theoretical possibility that a security hole in react-router would cause us problems.

apps/api/views.py
```diff
@@ -149,8 +149,9 @@ class RequestContributorTokenEmailView(APIView):
     throttle_classes = [ContributorRateThrottle]
 
     @staticmethod
-    def get_magic_link(domain, token, email):
-        return f"https://{domain}/{settings.CONTRIBUTOR_VERIFY_URL}?token={token}&email={quote_plus(email)}"
+    def get_magic_link(domain, token, email, next_url=None):
```
nrh-cklimas, line 152: I'd prefer we use either `redirect` or `next_url` for this, but not both in different places.

nrh-cklimas (changes requested):

apps/api/views.py
GuiMend: I used 2 different variables so not to confuse the following line:
```python
 def get_magic_link(domain, token, email, next_url=None):
        redirect = f"&redirect={quote_plus(next_url)}" if next_url else ""
```

spa/src/components/portal/TokenVerification/TokenVerification.test.tsx
```diff
@@ -31,6 +31,12 @@ describe('TokenVerification', () => {
     expect(screen.getByTestId('mock-redirect').dataset.to).toBe('/portal/my-contributions/');
   });
 
+  it('redirects to the custom URL if redirect is a query param', () => {
```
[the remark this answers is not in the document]
nrh-cklimas: I think maybe a misunderstanding here--it looks like you moved it below, to `When the user isn't authenticated`.

apps/api/views.py
nrh-cklimas: This was a code style comment, so if backend review was OK with it, this can proceed.

spa/src/components/portal/TokenVerification/TokenVerification.test.tsx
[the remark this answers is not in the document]
GuiMend: You are right; sorry about the misunderstanding. Fixed it.

nrh-cklimas (approved):

apps/api/views.py
njharman: tldr; +1 on code as it is now.

My understanding is `next_url` is the variable/name we care about. `redirect` is just a temporary variable used to so the optional &redirect if/else clause isn't jammed into the f-string (which would be less readable). Temp vars names are not important. esp in this very short function. Could be `x` or whatever correct term for "&name=value" fragment of url is. **NOT suggesting this change**

njharman: @GuiMend ^^^

apps/api/views.py
njharman: I asked about AC4 (security) and was told handled in front end. Since front-end review brings it up, we should secure it here ;). Also, it's often more robust to solve things "lower", "deeper", closest to source. So that, if something else (other part of FE, management command, etc) uses next_url, it is protected and that something else doesn't have to remember to check for full URL

I suggest 

1. Write parameterized tests that feed get_magic_link a bunch of next_urls; empty "", various full url bogus ones ftp:// https://, malformed relative urls [if there is such a thing], full url to contributor portal (I'm assuming we want to allow this?), . 
2. Update get_magic_link until tests pass;  this might be helpful https://stackoverflow.com/questions/8357098/how-can-i-check-if-a-url-is-absolute-using-python
3. Bad URLs should be ignored (next_url set to empty string) and log.warning made with details.

njharman (changes requested):

apps/api/views.py
GuiMend: [26eb352](https://github.com/newsrevenuehub/rev-engine/pull/1592/commits/26eb352df5140c4af577c43fe42d7ccaf571852b)

njharman (approved):

PR #1593, opened 2024-08-12
DEV-5000: Add default value for contributor name in email templates

GuiMend:

n/a
#### What's this PR do?
- Adds default value for contributor name in email templates
- Refactors `make_send_thank_you_email_data` to `generate_email_data` and use this function to generate all data necessary for email related to the contribution
- Use `generate_email_data` in all the places that send contribution email 
If the contributor name is undefined, we display "Dear contributor," instead of "Dear None,"
yes
no
yes
The contributor name on the email now has a default value
no
no
https://news-revenue-hub.atlassian.net/browse/DEV-5000
no
no


nrh-cklimas (commented):
This seems OK to me but I think this is a good opportunity to expand test coverage. I see two possibilities here:

1. We move the "if defined, contributor_name, if not, 'contributor' literal" logic into a method or property and add tests around that.
2. We add tests on the output of the other emails.

I like option 1 better because it's probably simpler to do :).

apps/emails/tasks.py
```diff
@@ -127,13 +128,13 @@ def make_send_thank_you_email_data(
 
     return SendContributionEmailData(
         contribution_amount=contribution.formatted_amount,
-        timestamp=convert_to_timezone_formatted(contribution.created, "America/New_York"),
+        timestamp=custom_timestamp or convert_to_timezone_formatted(contribution.created, "America/New_York"),
```
nrh-cklimas, line 131: Could you talk through why this date formatting is done differently than when custom_timestamp is set in models.py:655? My naive take on this would be to make custom_timestamp a datetime value and only format it here so that it's always consistent.

apps/emails/tasks.py
```diff
@@ -101,7 +102,7 @@ class SendMagicLinkEmailData(TypedDict):
 
 
 def make_send_thank_you_email_data(
```
nrh-cklimas, line 104: Consider renaming this since this is being used for other types of emails as well?

apps/emails/tasks.py
GuiMend: I don't know the reason for the different formatting @nrh-cklimas, I left that part unchanged and only added the prop `custom_timestamp` to override the default `contribution.created` time so that this function is more generic

apps/emails/tasks.py
GuiMend: [b5b6ac2](https://github.com/newsrevenuehub/rev-engine/pull/1593/commits/b5b6ac2660982d4929530a44713e11dd499ae731)

apps/emails/tasks.py
nrh-cklimas: I looked into the history and the reason for this distinction is that this method is used for upcoming charge reminders, and we specify only a date there for when the next charge will occur. This feels complex to untangle because the timestamp is serving multiple purposes (e.g. it mostly refers to when a contribution occurs, which makes sense to include a time). I'm OK with proceeding here but it feels like next time we touch this code, we try to re-organize it better.

nrh-cklimas (approved):

apps/emails/tasks.py
njharman: I suggest this be summarized into code comment

njharman: Also, using TZ "America/New_York" deserves and explanatory comment. [Unless I'm missing something obvious]

apps/emails/tasks.py
```diff
@@ -145,6 +144,11 @@ def make_send_thank_you_email_data(
         show_upgrade_prompt=False,
         billing_history=contribution.get_billing_history(),
         show_billing_history=show_billing_history,
+        default_contribution_page_url=(
+            contribution.revenue_program.default_donation_page.page_url
+            if contribution.revenue_program.default_donation_page
+            else None
+        ),
```
njharman, line 151: I don't know if it matters, but this is slightly different behavior than replaced code. This will always have key "default_contribution_page_url". Whereas previous code will only have key when default_page.page_url is truthy.

```
 if default_page := self.revenue_program.default_donation_page:
            data["default_contribution_page_url"] = default_page.page_url
```

apps/emails/tasks.py
```diff
@@ -120,20 +119,20 @@ def make_send_thank_you_email_data(
...
-        timestamp=convert_to_timezone_formatted(contribution.created, "America/New_York"),
+        timestamp=custom_timestamp or convert_to_timezone_formatted(contribution.created, "America/New_York"),
         contribution_interval_display_value=(
             contribution.interval if contribution.interval != ContributionInterval.ONE_TIME else ""
         ),
         contribution_interval=contribution.interval,
         contributor_email=contribution.contributor.email,
-        contributor_name=customer.name,
+        contributor_name=customer.name or "contributor",
         copyright_year=contribution.created.year,
```
njharman, line 136: Was this change from original intentional? If so, it probably should be documented in AC along with why.

`"copyright_year": datetime.datetime.now(datetime.timezone.utc).year,`

apps/emails/tasks.py
```diff
@@ -100,18 +101,16 @@ class SendMagicLinkEmailData(TypedDict):
...
     logger.info("make_send_than_you_email_data: called with contribution id %s", contribution.id)
 
     # vs circular import
     from apps.contributions.models import Contributor
 
     if not contribution.provider_customer_id:
-        logger.error(
-            "make_send_thank_you_email_data: No Stripe customer id for contribution with id %s", contribution.id
-        )
+        logger.error("[generate_email_data]: No Stripe customer id for contribution with id %s", contribution.id)
```
njharman, line 113: Unless this is a new standard. I don't think` [function_name]` should be included. Or, rather I do think it should be included via logging/sentry configuration. I get our two projects confused. But I belive at least one of them has logging configured to automatically include "where was this log statement from" information.

But I also don't think this needs to be changed back if it delays high priority ticket.

njharman (changes requested):

apps/emails/tasks.py
GuiMend: This is a pattern I saw in several places in rev-engine, so I think we can keep it this way for now @njharman .

apps/emails/tasks.py
GuiMend: This `"copyright_year": datetime.datetime.now(datetime.timezone.utc).year,` is only located in `preview_contribution_email_template` (so it's only visible for devs when previewing emails).

The code `copyright_year=contribution.created.year,` has been like this for at least 17 months. I don't believe it's a mistake

apps/emails/tasks.py
GuiMend: Thanks for pointing that out, but it isn't a problem @njharman. 
It is currently used only in `recurring-contribution-canceled` email, and we render it conditional `{% if default_contribution_page_url %}` (so having the key exist as `None` won't affect the usage).

apps/emails/tasks.py
GuiMend: I'm not sure about the reason for America/New_York timezone, but I added a comment with @nrh-cklimas' findings

apps/emails/tasks.py
njharman: Your explaination re: only visible for devs makes resolves issue. But...

I understood that you replaced this code  

https://github.com/newsrevenuehub/rev-engine/pull/1593/files#diff-80bc4fbf647c521d99e0b1e9e38a568c928ed7477e600e7bd110a8e9bc870ae1L671

line 671 has

`"copyright_year": datetime.datetime.now(datetime.timezone.utc).year,`

with calling the function in emails/tasks.py The actual line here might be old. Replacing the original code with this call on line 652 changes functionality.

njharman (approved):

apps/emails/tasks.py
GuiMend: Oh, right, that's a good point. I'd like to ask @benjaminEwhite for this.

Ben, do you think we can always use `copyright_year` as the current year instead of `contribution.created`?
Or do we need both cases? If we do, I'll update the function to also accept a `copyright_year` override.

apps/emails/tasks.py
```diff
@@ -120,20 +119,23 @@ def make_send_thank_you_email_data(
...
+        # because we have places where the format is different.
+        # Ex: upcoming charge reminders format is "MM/DD/YYYY"
+        timestamp=custom_timestamp or convert_to_timezone_formatted(contribution.created, "America/New_York"),
         contribution_interval_display_value=(
             contribution.interval if contribution.interval != ContributionInterval.ONE_TIME else ""
         ),
         contribution_interval=contribution.interval,
         contributor_email=contribution.contributor.email,
-        contributor_name=customer.name,
+        contributor_name=customer.name or "contributor",
```
benjaminEwhite, line 138: "contributor" gets hard coded in two places in this file. better off to define as constant at top of module and refer to by var name.

apps/emails/tasks.py
```diff
@@ -100,18 +101,16 @@ class SendMagicLinkEmailData(TypedDict):
     style: TransactionalEmailStyle
 
 
-def make_send_thank_you_email_data(
-    contribution: Contribution, show_billing_history: bool = False
+def generate_email_data(
```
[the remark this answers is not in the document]
benjaminEwhite: Also, would be nice to have a docstring.

apps/emails/tasks.py
benjaminEwhite: @GuiMend I don't have a strong opinion on what is right source for copyright year. I think that's ultimately a product question (do we always have current year as copyright, or is it the year at time of contribution). Personally, I'd lean to defaulting to current year, but again, that's a product/design question more than anything else.

apps/contributions/views.py
```diff
@@ -651,7 +651,7 @@ def send_contribution_receipt(self, request, pk=None, contribution_id=None) -> R
         logger.info("send receipt with contribution_id %s", contribution_id)
         contributor = self._get_contributor_and_check_permissions(request, pk)
         try:
-            contribution = self.get_contributor_queryset(contributor).get(pk=contribution_id)
+            contribution: Contribution = self.get_contributor_queryset(contributor).get(pk=contribution_id)
```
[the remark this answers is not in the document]
GuiMend: I think there is a slight improvement, as it makes the function "findable" and "clickable".
image

You can also see the documentation of the function:
image

apps/emails/tasks.py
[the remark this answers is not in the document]
GuiMend: We do have it here: 
image

Or do you mean something different?

apps/emails/tasks.py
```diff
@@ -120,20 +119,23 @@ def make_send_thank_you_email_data(
...
-            "make_send_thank_you_email_data: Something went wrong retrieving Stripe customer for contribution with id %s",
+            "[generate_email_data]: Something went wrong retrieving Stripe customer for contribution with id %s",
             contribution.id,
         )
         raise EmailTaskException("Cannot get required data from Stripe") from exc
 
     return SendContributionEmailData(
         contribution_amount=contribution.formatted_amount,
-        timestamp=convert_to_timezone_formatted(contribution.created, "America/New_York"),
+        # We are not formatting the "custom_timestamp" in "convert_to_timezone_formatted"
```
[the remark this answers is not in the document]
GuiMend: This is the context @benjaminEwhite: https://github.com/newsrevenuehub/rev-engine/pull/1593#discussion_r1725028823

Let me know if you still think I need to change anything.

apps/emails/tasks.py
GuiMend: Confirmed with Sophie and Rachel: `It should always be current year, consistent with all our emails`

benjaminEwhite (approved):

x110dc (changes requested):
One question. But I think we also need to revise the "what does this ticket so" section of the PR description. Because it looks like this ticket could've been accomplished with a 3-line change (adding the constant for contributor and then using it in two places). So some refactoring has been done. Please explain that in the PR description.

apps/emails/tasks.py
```diff
@@ -120,21 +137,24 @@ def make_send_thank_you_email_data(
...
+        # because we have places where the format is different.
+        # Ex: upcoming charge reminders format is "MM/DD/YYYY"
+        timestamp=custom_timestamp or convert_to_timezone_formatted(contribution.created, "America/New_York"),
```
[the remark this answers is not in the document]
GuiMend: Yes, [DEV-5170](https://news-revenue-hub.atlassian.net/browse/DEV-5170)

PR #1616, opened 2024-09-04
DEV-4744: RevEngine not sending country code to Bad Actor API (Cf-Ipcountry header)

GuiMend:

n/a
#### What's this PR do?
Sends user's country_code to bad actor
Help identify in bad actor if card issuer and user country are the same
yes
no
n/a
n/a
no
no
https://news-revenue-hub.atlassian.net/browse/DEV-4744
~~Probably need to remove the `logging` commit added for BadActor to assist in QA.~~ no longer needed
no


apps/contributions/serializers.py
```diff
@@ -394,7 +394,9 @@ def get_bad_actor_score(self, data):
             "page": data["page"].id,
             "referer": self.context["request"].META.get("HTTP_REFERER"),
             "ip": get_original_ip_from_request(self.context["request"]),
+            "country_code": self.context["request"].headers.get("Cf-Ipcountry", None),
         }
+        logger.info("BadActorSerializer data: %s", data)
```
GuiMend, line 399: Needed for QA. If removing this log is better, I can remove the commit right after QA passes.

apps/contributions/serializers.py
```diff
@@ -394,7 +394,9 @@ def get_bad_actor_score(self, data):
             "page": data["page"].id,
             "referer": self.context["request"].META.get("HTTP_REFERER"),
             "ip": get_original_ip_from_request(self.context["request"]),
+            "country_code": self.context["request"].headers.get("Cf-Ipcountry", None),
```
benjaminEwhite, line 397: Believe it will default to `None` by default.

benjaminEwhite (changes requested):
One small comment inline. Aside from that, I think it would make sense to have a unit test around the business logic here (show that when header has CF-Ipcountry, the data passed to `make_bad_actor_request` contains the value, and when header does not a have an entry, not included in data).

apps/contributions/serializers.py
benjaminEwhite: Also, the AC say that if no code in request header, then should not be included in request to bad actor.  So need some conditionality around that.

GuiMend: Got it, my bad. I thought that sending `"country_code": None` was enough to pass AC2.

benjaminEwhite (approved):

apps/contributions/serializers.py
x110dc: I think it's fine to leave it as-is.

x110dc (approved):

benjaminEwhite (approved):

PR #1617, opened 2024-09-04
DEV-5146 | Bad actor response or lack thereof should never block a contribution

benjaminEwhite:

### DIFF
--- a/apps/contributions/bad_actor.py
+++ b/apps/contributions/bad_actor.py
@@ -4,20 +4,38 @@ from json.decoder import JSONDecodeError
 from django.conf import settings
 
 import requests
+from pydantic import BaseModel, ValidationError
 
 
 logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
+
+
+class BadActorScore(BaseModel):
+    """Represents nested items of the Bad Actor API response."""
+
+    judgment: int
+    label: str
+    value: str
+
+
+class BadActorOverallScore(BaseModel):
+    """Represents the top-level structure of the Bad Actor API response."""
+
+    overall_judgment: int
+    items: list[BadActorScore]
 
 
 class BadActorAPIError(Exception):
     pass
 
 
-def make_bad_actor_request(validated_data):
-    """Bad Actor API result.
+def get_bad_actor_score(validated_data) -> BadActorOverallScore:
+    """Make a request to the bad actor API and return score data.
+
+    Note that this function tries hard to return the response.json() as a dict or else raise a BadActorAPIError.
 
     param validated_data: json posted to bad actor api.
-    return: response.json() API response json
+    return: BadActorOverallScore derived from API response's JSON
     raises: BadActorAPIError
     """
     if not settings.BAD_ACTOR_API_KEY or not settings.BAD_ACTOR_API_URL:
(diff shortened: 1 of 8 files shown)


apps/contributions/bad_actor.py
```diff
@@ -33,15 +51,23 @@ def make_bad_actor_request(validated_data):
...
+    try:
+        response = requests.post(
+            url=settings.BAD_ACTOR_API_URL,
+            headers=headers,
+            json=validated_data,
+            timeout=settings.REQUESTS_TIMEOUT_DEFAULT,
+        )
+    except Exception as exc:
+        logger.exception("BadActor API request failed communicating with API")
+        raise BadActorAPIError("BadActor API request failed") from exc
```
njharman, line 63: Maybe include response.status_code in logging text (for diagnosis)

apps/contributions/bad_actor.py
```diff
@@ -33,15 +51,23 @@ def make_bad_actor_request(validated_data):
     headers = {
         "Authorization": f"Bearer {settings.BAD_ACTOR_API_KEY}",
     }
-    json_data = validated_data.copy()
```
njharman, line 36: I, probably like you, assumed copy() is spurious?  Only reason I can think is to prevent changes to validated_data (the mutable object) being passed back to caller.  I can't imagine requests.post() alters json parameter.

apps/contributions/bad_actor.py
```diff
@@ -33,15 +51,23 @@ def make_bad_actor_request(validated_data):
...
+    try:
+        json_response = response.json()
+    except JSONDecodeError as exc:
+        logger.exception("BadActor API request failed with malformed JSON")
+        raise BadActorAPIError("BadActor API request failed with malformed JSON") from exc
+    try:
+        return BadActorOverallScore(**json_response)
+    except ValidationError as exc:
+        logger.exception("BadActor API request failed with invalid JSON")
+        raise BadActorAPIError("BadActor API request failed with invalid JSON") from exc
```
njharman, line 73: fyi, can have multiple except blocks.

```
    try:
        return BadActorOverallScore(**response.json())
    except JSONDecodeError as exc:
        logger.exception("BadActor API request failed with malformed JSON")
        raise BadActorAPIError("BadActor API request failed with malformed JSON") from exc
    except ValidationError as exc:
        logger.exception("BadActor API request failed with invalid JSON")
        raise BadActorAPIError("BadActor API request failed with invalid JSON") from exc
```

apps/contributions/serializers.py
```diff
@@ -414,7 +415,7 @@ def should_flag(self, bad_actor_score):
         """Determine if bad actor score should lead to contribution being flagged."""
         return bad_actor_score == settings.BAD_ACTOR_FLAG_SCORE
 
-    def build_contribution(self, contributor, validated_data, bad_actor_response=None):
+    def build_contribution(self, contributor, validated_data, bad_actor_response=BadActorOverallScore | None):
```
njharman, line 418: We never call build_contribution() without bad_actor_repsonse.  It should not be optional/not have a default. 

```
def build_contribution(self, contributor, validated_data, bad_actor_response):
  # I don't know syntax to type annotate it as a BadActorOverallScore
```


` bad_actor_response=BadActorOverallScore | None`

Is that really type annotation syntax? Saying it can be either of those but default is None.  It just looks like operator.  bitwise or dict merge (both of which should raise error) Wondering if pydantic overloads pipe for something?

njharman (changes requested):

apps/contributions/bad_actor.py
benjaminEwhite: I have same theory about why .copy() is here.

I agree that it's unlikely requests.post would mutate that arg.

apps/contributions/bad_actor.py
benjaminEwhite: We won't have response because exception was raised in initializing that value.

apps/contributions/bad_actor.py
benjaminEwhite: In general, I try to isolate a single exception-raising action on to a single line, with a catch block immediately following that line, so going to stick with that pattern even though in this case your suggestion is more succinct.

apps/contributions/serializers.py
benjaminEwhite: That was a typo on my part.  I meant to leave a type hint, not set a default.

Updated in cc883138cd.

apps/contributions/serializers.py
```diff
@@ -402,7 +403,7 @@ def get_bad_actor_score(self, data):
             logger.warning("BadActor serializer raised a ValidationError", exc_info=exc)
             return None
         try:
-            return make_bad_actor_request(data).json()
+            return get_bad_actor_score(data)
         except BadActorAPIError:
```
x110dc, line 407: Are we certain no other exceptions can be thrown? I'd prefer to see a catchall `except Exception` after this.

apps/users/views.py
```diff
@@ -291,7 +291,7 @@ def validate_bad_actor(self, data):
         except BadActorAPIError:
             logger.warning("Something went wrong with BadActorAPI", exc_info=True)
             return
```
x110dc, line 293: again, is there any possibility of another class of exception? I'd like to see an additional `except Exception` clause here.

x110dc (changes requested):
two requests for changes

apps/contributions/serializers.py
benjaminEwhite: Shored this up in ccd62f0253

apps/contributions/models.py
```diff
@@ -619,6 +619,7 @@ def cancel(self):
                 self.interval,
             )
             raise ContributionIntervalError()
+        # at this poin, we know the contribution is recurring
```
x110dc, line 622: typo: poin

x110dc (approved):

apps/contributions/models.py
benjaminEwhite: 1ebdd6c

PR #1621, opened 2024-09-09
DEV-5209: Add org id and action to bad actor

GuiMend:

n/a
#### What's this PR do?
Adds "action" and "org_id" to bad actor request
We need to be able to treat some orgs differently in bad actor and knowing the org would enable that.
yes
no
yes
n/a
no
n/a
https://news-revenue-hub.atlassian.net/browse/DEV-5209
no
no


apps/contributions/serializers.py
```diff
@@ -386,15 +390,19 @@ def validate(self, data):
         self.do_conditional_validation(data)
         return data
 
-    def get_bad_actor_score(self, data):
+    def get_bad_actor_score(self, data, action: str = "contribution"):
```
benjaminEwhite, line 393: "contribution" here is a "magic value" so probably should be defined in all caps const at top of module and can be referenced elsewhere (or alternatively defined somehwere else that makes sense and imported here).

benjaminEwhite, line 393: All else equal, it would be nice to provide more detailed type hints about `data` param.

apps/users/tests/test_views.py
```diff
@@ -670,6 +671,20 @@ def test_create_happy_path(self, mocker, api_client, valid_create_request_data):
...
 
+    def test_make_bad_actor_request_payload(self, mocker, api_client, valid_create_request_data):
+        mock_bad_actor_request = mocker.patch(
+            "apps.users.views.make_bad_actor_request", return_value=MockResponseObject({"overall_judgment": 0})
+        )
+        response = api_client.post(reverse("user-list"), data=valid_create_request_data)
+        assert response.status_code == status.HTTP_201_CREATED
+
+        mock_bad_actor_request.assert_called_once()
+        assert mock_bad_actor_request.call_args[0][0]["action"] == "create-account"
```
benjaminEwhite, line 682: "create-account" should be in a const var in object code that gets imported into tests.

apps/users/views.py
```diff
@@ -286,6 +286,7 @@ def validate_bad_actor(self, data):
                     # Bad actor api requires this field because it was created
                     # with contributors in mind, not org users, so we supply a dummy value
                     "amount": BAD_ACTOR_FAKE_AMOUNT,
+                    "action": "create-account",
```
benjaminEwhite, line 289: define in constant

apps/contributions/serializers.py
```diff
@@ -141,13 +141,17 @@ class CompSubscriptions(TextChoices):
 
 
 class BadActorSerializer(serializers.Serializer):
+    ACTION_CHOICES = ["contribution", "create-account"]
+
```
benjaminEwhite, line 145: Thes should be set to constant vals accessible in (relatively) global scope.

apps/users/tests/test_views.py
```diff
@@ -670,6 +671,20 @@ def test_create_happy_path(self, mocker, api_client, valid_create_request_data):
...
+        assert mock_bad_actor_request.call_args[0][0]["email"] == valid_create_request_data["email"]
+        assert mock_bad_actor_request.call_args[0][0]["amount"] == BAD_ACTOR_FAKE_AMOUNT
+        assert mock_bad_actor_request.call_args[0][0]["first_name"] == ""
+        assert mock_bad_actor_request.call_args[0][0]["last_name"] == ""
```
benjaminEwhite, line 686: Seems fine, but the reason for these particular values is kind of opaque, if i'm trying to understand what the business logic is here. Are first name and last name empty string because that's default value for those fields? If so, shared in situ var to that effect with appropriate name will make that meaning clear

apps/contributions/serializers.py
```diff
@@ -386,15 +390,19 @@ def validate(self, data):
...
+    def get_bad_actor_score(self, data, action: str = "contribution"):
         """Based on validated data, make a request to bad actor API and return its response."""
         data = data | {
             # we use a PrimaryKeyRelated serializer field for page in BaseCreatePaymentSerializer
             # but BadActorSerializer wants to pk, so we reformat here.
             "page": data["page"].id,
             "referer": self.context["request"].META.get("HTTP_REFERER"),
             "ip": get_original_ip_from_request(self.context["request"]),
+            "action": action,
+            "org": data["page"].organization.id,
```
benjaminEwhite, line 402: How confident are we about data['page"].organization.id having referential integrity?  Do we have a strong gaurantee here, and if not, what happens if this long chain isn't found at runtime?

apps/users/tests/test_views.py
```diff
@@ -670,6 +671,20 @@ def test_create_happy_path(self, mocker, api_client, valid_create_request_data):
 
         mock_bad_actor_request.assert_called_once()
 
+    def test_make_bad_actor_request_payload(self, mocker, api_client, valid_create_request_data):
```
benjaminEwhite, line 674: Docstring describing higher level what this test is trying to prove would be helpful here.

benjaminEwhite (changes requested):

apps/contributions/serializers.py
GuiMend: `data['page']` is a `DonationPage` instance. I'd think a `DonationPage` would always have an `RP` and `Org` associated with it. What do you think?

Or should we do something like:
```suggestion
            "org": data["page"].organization.id if data["page"].organization else None,
```

apps/users/tests/test_views.py
GuiMend: I did some testing, and both `first_name` and `last_name` will always return `""` as the serializer doesn't even have these fields available
image

I could change these lines to 
image

```python
"first_name": "",
"last_name": "",
```

Or remove these fields entirely.

apps/contributions/serializers.py
GuiMend: I'm not sure how to type hint here. The `data` type is the same as `validated_data` in the `create` function, and it is `any` over there.
image

GuiMend: [b571797](https://github.com/newsrevenuehub/rev-engine/pull/1621/commits/b571797add900d583b11524af66b19e2cc63a570)

apps/users/tests/test_views.py
GuiMend: [b571797](https://github.com/newsrevenuehub/rev-engine/pull/1621/commits/b571797add900d583b11524af66b19e2cc63a570)

apps/users/views.py
GuiMend: [b571797](https://github.com/newsrevenuehub/rev-engine/pull/1621/commits/b571797add900d583b11524af66b19e2cc63a570)

apps/contributions/serializers.py
GuiMend: [b571797](https://github.com/newsrevenuehub/rev-engine/pull/1621/commits/b571797add900d583b11524af66b19e2cc63a570)

apps/contributions/serializers.py
```diff
@@ -386,15 +388,19 @@ def validate(self, data):
         self.do_conditional_validation(data)
         return data
 
-    def get_bad_actor_score(self, data):
+    def get_bad_actor_score(self, data, action: BadActorAction = BadActorAction.CONTRIBUTION):
```
benjaminEwhite, line 391: Think it's better to have no default here and to force callers to specify. Either that or default to None.

benjaminEwhite (approved):
Had one comment about a default value. Feel free to resolve that on your own @Gui without re-review from me.  I'm approving with understanding you'll resolve that before assigning to next reviewer.

benjaminEwhite (approved):

x110dc (approved):

PR #1628, opened 2024-09-25
DEV-4882: Make provider_subscription_id unique

GuiMend:

n/a
#### What's this PR do?
- Makes `provider_subscription_id` unique
Doing this will eliminate a potential source of problematic data and allow us to make simpler assumptions about these values when thinking about contributions.
n/a
no
n/a
n/a
no
no
https://news-revenue-hub.atlassian.net/browse/DEV-4882
no
https://news-revenue-hub.atlassian.net/browse/DEV-5271


benjaminEwhite (changes requested):
Changes seem fine, but looking at this now, I think we should discuss (in JIRA or Slack) possibility of having data migration for duplicate cases. In practice, we don't have duplicate subscription IDs in prod so this should be safe to run, and we may decide that's a good enough guarantee, but we can at least consider having a de-duping step.

benjaminEwhite (approved):

x110dc (approved):

PR #1633, opened 2024-10-01
DEV-5276: Send amount change email

GuiMend:

n/a
#### What's this PR do?
- Sends email for contributors when they change a contribution amount
- Improve communication 
- ye
- yes
- yes
- Email is sent for contributors when they change a contribution amount
- no
- no
- https://news-revenue-hub.atlassian.net/browse/DEV-5276
- no
- no

### DIFF
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -714,6 +714,11 @@ class Contribution(IndexedTimeStampedModel):
             "New change to your contribution", "recurring-contribution-payment-updated"
         )
 
+    def send_recurring_contribution_amount_updated_email(self) -> None:
+        self.send_recurring_contribution_change_email(
+            "New change to your contribution", "recurring-contribution-amount-updated"
+        )
+
     def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.date = None) -> None:
         self.send_recurring_contribution_change_email(
             f"Reminder: {self.revenue_program.name} scheduled contribution",
@@ -1229,6 +1234,8 @@ class Contribution(IndexedTimeStampedModel):
 
         item = items["data"][0]
         self.contribution_metadata["donor_selected_amount"] = amount
+        # Need to update contribution amount here so that it reflects the new amount in the email
+        self.amount = amount
 
         logger.info(
             "Updating Stripe Subscription's %s (item %s), amount to %s",
@@ -1264,10 +1271,13 @@ class Contribution(IndexedTimeStampedModel):
             raise
 
         with reversion.create_revision():
-            self.save(update_fields={"contribution_metadata", "modified"})
+            self.save(update_fields={"contribution_metadata", "amount", "modified"})
             reversion.set_comment(
                 f"`Contribution.update_subscription_amount` saved changes to contribution with ID {self.id}"
             )
+
+        # Send amount updated email to contributor
+        self.send_recurring_contribution_amount_updated_email()
 
 
 def ensure_stripe_event(event_types: list[str] = None) -> Callable:
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -1234,7 +1234,6 @@ class Contribution(IndexedTimeStampedModel):
 
         item = items["data"][0]
         self.contribution_metadata["donor_selected_amount"] = amount
-        # Need to update contribution amount here so that it reflects the new amount in the email
         self.amount = amount
 
         logger.info(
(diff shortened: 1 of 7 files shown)


apps/contributions/models.py
```diff
@@ -1229,6 +1234,8 @@ def update_subscription_amount(self, amount: int) -> None:
 
         item = items["data"][0]
         self.contribution_metadata["donor_selected_amount"] = amount
+        # Need to update contribution amount here so that it reflects the new amount in the email
+        self.amount = amount
```
nrh-cklimas, line 1238: I think there's a potential problem here if the Stripe update fails. In that case, the db model has the amount that the user wanted to update to, but Stripe doesn't.

I think this can be solved by either:
- Passing the updated amount to `send_recurring_contribution_amount_updated_email` (I think I prefer this but I'm not 100% on it)
- Moving the change to the model field to after the Stripe call and add a unit test for this scenario

apps/emails/templates/nrh-default-contribution-base-template.html
nrh-cklimas, line 1: Since we're changing the base template, it makes sense to me to add regression tests in QA instructions to run through existing emails to make sure there aren't any unintended changes.

nrh-cklimas (changes requested):

apps/contributions/models.py
GuiMend: I don't think this will be a problem @nrh-cklimas.
We only ever `save` the contributor after the try/except (line 1274)

```python
self.save(update_fields={"contribution_metadata", "amount", "modified"})
```

nrh-cklimas: I see what you're saying. If it's ok with you, would like an opinion from @benjaminEwhite on approach here.

nrh-cklimas (approved):

apps/contributions/models.py
```diff
@@ -1264,11 +1271,14 @@ def update_subscription_amount(self, amount: int) -> None:
...
         with reversion.create_revision():
-            self.save(update_fields={"contribution_metadata", "modified"})
+            self.save(update_fields={"contribution_metadata", "amount", "modified"})
             reversion.set_comment(
                 f"`Contribution.update_subscription_amount` saved changes to contribution with ID {self.id}"
             )
 
+        # Send amount updated email to contributor
+        self.send_recurring_contribution_amount_updated_email()
+
```
benjaminEwhite, line 1281: I know we already save contribution at this moment in status quo, but all else equal, I think we would want to rely on Stripe webhooks to trigger a handler that updates revengine contribution amount and triggers reminder email, because that's evidence of transation finality on Stripe side (which is the source of truth here).

As implemented here, we can end up in a situation where Stripe is updated, but revengine is not.

benjaminEwhite (changes requested):

apps/contributions/models.py
nrh-cklimas: @GuiMend could you clarify if on the main branch, we are updating amount in Django correctly when a subscription amount is changed? I thought we were but looking at [the original PR](https://github.com/newsrevenuehub/rev-engine/pull/1633), I'm getting confused.

benjaminEwhite: FWIW, another way we also might be able to use `@transaction.atomic` such that we update we first save the changes to revengine then call stripe to update. If things work out with stripe, the amount change is committed to DB, but if not, it is not (despite being first in terms of code line).  This would require that we allow exceptions to bubble up.

apps/contributions/models.py
```diff
@@ -1229,6 +1234,8 @@ def update_subscription_amount(self, amount: int) -> None:
 
         item = items["data"][0]
         self.contribution_metadata["donor_selected_amount"] = amount
+        # Need to update contribution amount here so that it reflects the new amount in the email
```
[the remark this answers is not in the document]
GuiMend: This comment is unnecessary. I'll remove it. It could have been a GitHub PR comment instead.

I just wanted to point out that we are already saving the "amount" change, but I had to make it earlier so that the email data generation works as expected.

```
# File /contributions/serializers.py > line 974
                instance.save(update_fields={*validated_data.keys(), "modified"})
                reversion.set_comment("Updated by PortalContributionDetailSerializer.update")
```

```suggestion
```

apps/contributions/models.py
GuiMend: Yes @nrh-cklimas , we are. The amount is being saved here:

```python
# File /contributions/serializers.py > line 974
                instance.save(update_fields={*validated_data.keys(), "modified"})
                reversion.set_comment("Updated by PortalContributionDetailSerializer.update")
```

benjaminEwhite (approved):

PR #1635, opened 2024-10-01
DEV-5216 | Make BadActorSerializer more idiomatic

benjaminEwhite:

#### What's this PR do?

This PR refactors `BadActorSerializer`, removing an over-ride of `to_internal_value`. As noted in ticket for this PR, I discovered the following oddity:

```
from apps.contributions.serializers import BadActorSerializer

instance  = BadActorSerializer(data={"first_name": "foo", "last_name": "bar", "email": "foo@bar.com", "amount": "12400", "ip":  "192.0.2.1", "referer": "https://fundjournalism.org" })
instance.is_valid()
# false
instance.errors
Out[4]: {'street': [ErrorDetail(string='This field may not be null.', code='null')], '.....
```

Not totally clear why, but somehow using the over-ride of to_internval_value cased the default values for those fields to not be present, triggering a validation error.

Unclear if any bugs directly resulted from that, but it causes the serializer to behave in unexpected ways.

To solve for this, instead of mapping source values to destination values in `to_internal_value` we just create input data with necessary keys ahead of initializing serializer.

Make BadActorSerializer's behavior more predictable

Yes

No

No. We remove existing code.

N/A

No

N/A

[DEV-5216](https://news-revenue-hub.atlassian.net/browse/DEV-5216)

No

No

[DEV-5216]: https://news-revenue-hub.atlassian.net/browse/DEV-5216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -397,7 +397,7 @@ class BaseCreatePaymentSerializer(serializers.Serializer):
         Note that this method is meant to be ironclad against exceptions. If anything goes wrong, it should return None.
         """
         try:
-            data = {
+            serializer_data = {
                 "amount": data["amount"],
                 "first_name": data["first_name"],
                 "last_name": data["last_name"],
@@ -417,9 +417,9 @@ class BaseCreatePaymentSerializer(serializers.Serializer):
                 "page": data["page"].id,
             }
             if country_code := self.context["request"].headers.get("Cf-Ipcountry", None):
-                data["country_code"] = country_code
-            logger.info("BadActorSerializer data: %s", data)
-            serializer = BadActorSerializer(data=data)
+                serializer_data["country_code"] = country_code
+            logger.info("BadActorSerializer data: %s", serializer_data)
+            serializer = BadActorSerializer(data=serializer_data)
             serializer.is_valid(raise_exception=True)
             return get_bad_actor_score(serializer.validated_data)
         except serializers.ValidationError as exc:
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -154,31 +154,28 @@ class BadActorSerializer(serializers.Serializer):
     org = serializers.CharField(max_length=255, required=False, default="", allow_blank=True)
     street = serializers.CharField(max_length=255, required=False, default="", allow_blank=True)
     complement = serializers.CharField(max_length=255, required=False, default="", allow_blank=True)
-    city = serializers.CharField(max_length=40, required=False, default="", allow_blank=True)
-    state = serializers.CharField(max_length=80, required=False, default="", allow_blank=True)
+    city = serializers.CharField(
+        max_length=40,
+        required=False,
+        default="",
+        allow_blank=True,
+    )
+    state = serializers.CharField(
+        max_length=80,
+        required=False,
+        default="",
+        allow_blank=True,
+    )
     country = serializers.CharField(max_length=80, required=False, default="", allow_blank=True)
     zipcode = serializers.CharField(max_length=20, required=False, default="", allow_blank=True)
     country_code = serializers.CharField(max_length=10, required=False, default="", allow_blank=True)
-
     # Third-party risk assessment
     captcha_token = serializers.CharField(max_length=2550, required=False, allow_blank=True)
-
     # Request metadata
     ip = serializers.IPAddressField()
     referer = serializers.URLField()
-
     # Contribution additional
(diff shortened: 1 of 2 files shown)


apps/contributions/serializers.py
```diff
@@ -154,32 +154,29 @@ class BadActorSerializer(serializers.Serializer):
     org = serializers.CharField(max_length=255, required=False, default="", allow_blank=True)
     street = serializers.CharField(max_length=255, required=False, default="", allow_blank=True)
     complement = serializers.CharField(max_length=255, required=False, default="", allow_blank=True)
-    city = serializers.CharField(max_length=40, required=False, default="", allow_blank=True)
-    state = serializers.CharField(max_length=80, required=False, default="", allow_blank=True)
+    city = serializers.CharField(
```
njharman, line 157: I don't see any changes in these lines. Wonder why formatter chose to break apart, ... Oh, probably the trailing comma.

I find that annoying.

apps/contributions/serializers.py
```diff
@@ -399,23 +396,29 @@ def get_bad_actor_score(self, data, action: BadActorAction) -> BadActorOverallSc
 
         Note that this method is meant to be ironclad against exceptions. If anything goes wrong, it should return None.
         """
+        data = {
```
njharman, line 399: Given docstring 
> Note that this method is meant to be ironclad against exceptions. If anything goes wrong, it should return None.

All of this should be in try block

njharman, line 399: There's a typo in the catch Exception "sore" should be "score".  But real reason for comment is consider adding debugging info to exception log. Such as "data" (before it's modified)

apps/contributions/serializers.py
```diff
@@ -399,23 +396,29 @@ def get_bad_actor_score(self, data, action: BadActorAction) -> BadActorOverallSc
...
+        }
+        if country_code := self.context["request"].headers.get("Cf-Ipcountry", None):
+            data["country_code"] = country_code
+        logger.info("BadActorSerializer data: %s", data)
+        serializer = BadActorSerializer(data=data)
         try:
-            data = data | {
-                # we use a PrimaryKeyRelated serializer field for page in BaseCreatePaymentSerializer
-                # but BadActorSerializer wants to pk, so we reformat here.
-                "page": data["page"].id,
```
njharman, line 406: I don't see "page" replicated above.

Was it intentional removal?

apps/contributions/serializers.py
njharman, line 399: Suggest not reusing name "data" also used as argument.

njharman (changes requested):

apps/contributions/serializers.py
```diff
@@ -399,23 +396,29 @@ def get_bad_actor_score(self, data, action: BadActorAction) -> BadActorOverallSc
...
         """
+        data = {
+            "amount": data["amount"],
+            "first_name": data["first_name"],
+            "last_name": data["last_name"],
+            "email": data["email"],
+            "action": action,
+            "org": data["page"].organization.id,
+            "street": data["mailing_street"],
+            "complement": data["mailing_complement"],
```
njharman, line 407: complement, captcha_token, ip, referer, reason_for_giving don't seem to be tested for in `test_get_bad_actor_score_fields()`

But there probably should be a separate test to prove AC2. Something with all default values and none of those defaults are `None`. At least that is my understanding of issue

> string='This field may not be null.'

njharman (changes requested):

apps/contributions/serializers.py
benjaminEwhite: And that's why we have code review...

apps/contributions/serializers.py
benjaminEwhite: 25c2a0e899

apps/contributions/serializers.py
benjaminEwhite: fd41aec9ff0

apps/contributions/serializers.py
benjaminEwhite: 1abbbc571

benjaminEwhite: Resolved the typo, but not sure which commit.

RE: adding data d59fb117e49b

benjaminEwhite: Resolved the typo, but not sure in which commit.

Re: data and logging d59fb117e49b

nrh-cklimas (approved):

PR #1646, opened 2024-10-23
DEV-5359: Conditionally update donor_selected_amount

nrh-cklimas:

n/a

#### What's this PR do?

- When a contribution amount is updated, this conditionally updates `donor_selected_amount` if the metadata schema supports it.
- Adds a `set_metadata_field` method to Contribution that allows arbitrary changes to data that is validated.
- Fixes tests that are unrelated to the PR, but were failing on main after a recent merge.

Only some contributions' metadata schemas support `donor_selected_amount`.

Yes.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5359

No.

https://news-revenue-hub.atlassian.net/browse/ITS-2610


apps/contributions/models.py
```diff
@@ -1271,7 +1271,13 @@ def update_subscription_amount(self, amount: int) -> None:
             raise ValueError("Subscription should have only one item")
 
         item = items["data"][0]
-        self.contribution_metadata["donor_selected_amount"] = amount
+
+        # Not all contribution metadata schemas support the
+        # donor_selected_amount field, so we should only set it when it's
+        # already present.
+
+        if "donor_selected_amount" in self.contribution_metadata:
```
benjaminEwhite, line 1279: I think we might want to be stricter here by:

1. cast contribution_metadata to an instance of known schema version via `cast_metadata_to_stripe_payment_metadata_schema`
2. update the condition to be current condition + "donor_selected_amount" is in the specific schema

apps/contributions/models.py
```diff
@@ -1271,7 +1271,13 @@ def update_subscription_amount(self, amount: int) -> None:
...
+            self.contribution_metadata["donor_selected_amount"] = amount
```
benjaminEwhite, line 1280: Would also be good to use the metadata schema to validate the value being set for donor selected amount (since the schemas are pydantic models).

apps/contributions/tests/test_models.py
```diff
@@ -1852,6 +1852,36 @@ def test_update_subscription_amount_when_success_update_contribution(
         assert monthly_contribution.contribution_metadata == metadata
         assert monthly_contribution.amount == new_amount
 
+    @pytest.mark.parametrize("contribution_metadata", [{}, {"donor_selected_amount": 123}])
```
benjaminEwhite, line 1855: Getting stricter above would potentially requiring fully mocking out the metadata here (though there may be pre-existing fixtures that would help, i don't recall off hand).

benjaminEwhite (changes requested):

apps/contributions/models.py
nrh-cklimas: So semi-pseudocoding your suggestion, are you saying..?

```python
typed_metadata = cast_metadata_to_stripe_payment_metadata_schema(contribution.contribution_metadata)

if (isinstance(typed_metadata, StripePaymentMetadataSchemaV1_4) or \
    isinstance(typed_metadata, StripePaymentMetadataV1_5)):
```

or:

```python
typed_metadata = cast_metadata_to_stripe_payment_metadata_schema(contribution.contribution_metadata)

if (hasattr(type(typed_metadata), "donor_selected_amount")):
```

apps/contributions/tests/test_models.py
```diff
@@ -1647,6 +1648,34 @@ def test_card_owner_name_when_no_provider_payment_method_details(self, mocker):
         contribution = ContributionFactory(provider_payment_method_details=None)
         assert contribution.card_owner_name == ""
 
+    def test_set_metadata_field_happy_path(self, contribution: Contribution):
+        # This field is present in all schema versions.
+        contribution.set_metadata_field(
+            "agreed_to_pay_fees", not contribution.contribution_metadata["agreed_to_pay_fees"]
```
benjaminEwhite, line 1654: if only testing one field, might as well use "donor_selected_amount" since that's the only field we're going to be setting with this function for now.

benjaminEwhite (changes requested):

apps/contributions/tests/test_models.py
nrh-cklimas: 19bdff7a2ff7f1eb06300eb65d502a1998795ad2

apps/contributions/models.py
```diff
@@ -764,6 +771,26 @@ def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.
...
+        schema = STRIPE_PAYMENT_METADATA_SCHEMA_VERSIONS.get(
+            key if key == "schema_version" else self.contribution_metadata["schema_version"], None
+        )
+        if not schema:
+            raise InvalidMetadataError(f"No schema found for version {self.contribution_metadata['schema_version']}")
+        try:
+            schema(**(self.contribution_metadata | {key: value}))
+        except ValidationError as error:
+            raise InvalidMetadataError(f"Change to {key} results in invalid contribution metadata") from error
+        self.contribution_metadata[key] = value
```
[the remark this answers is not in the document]
nrh-cklimas: - I'm on board with blocking schema version changes.
- Would it make sense to ask callers to log as appropriate? e.g. it wouldn't make sense here, I don't think, unless we do it as a debug log?
- I tried the code suggestion and got test regressions elsewhere:

```
FAILED apps/contributions/tests/test_models.py::TestContributionModel::test_update_subscription_amount_when_success_update_contribution - TypeError: Object of type Url is not JSON serializable
FAILED apps/contributions/tests/test_models.py::TestContributionModel::test_update_subscription_amount_updates_donor_selected_amount_metadata[contribution_metadata1] - TypeError: Object of type Url is not JSON serializable
FAILED apps/contributions/tests/test_models.py::TestContributionModel::test_sends_updated_email_when_update_subscription_amount_success - TypeError: Object of type Url is not JSON serializable
```
I think this might indicate real issues? I think the issue is that the `dict()` call is transforming all metadata to the Pydantic-parsed version.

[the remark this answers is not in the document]
benjaminEwhite: @nrh-cklimas ah looks like there was issue converting the pydantic.http url field.

I just updated a line in the suggestion to solve for that (need to output the pydantic instance to JSON string then load that as dict...).

apps/contributions/models.py
benjaminEwhite: Closing this thread cause I don't htink it's relevant anymore.

apps/contributions/models.py
benjaminEwhite: Also resolving this because I believe it's subsumed by other open conversations.

benjaminEwhite (changes requested):
@nrh-cklimas I resolved some conversations and left an updated suggestion for `set_metadata_field` that solves for test issue you encountered.

apps/contributions/models.py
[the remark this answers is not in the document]
nrh-cklimas: Works for me. Added in c11e24e560c7249611814c40c5d13ca6b4349b0d with a check at the top to make sure we're not changing the schema version, as previously discussed.

apps/contributions/models.py
```diff
@@ -764,6 +772,25 @@ def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.
...
+        This doesn't make any changes in Stripe.
+        """
+        if key == "schema_version":
+            raise InvalidMetadataError("Schema version may not be changed")
+        if not (version := self.contribution_metadata.get("schema_version")):
+            raise InvalidMetadataError("No schema version set in metadata")
+        try:
+            schema = STRIPE_PAYMENT_METADATA_SCHEMA_VERSIONS[version]
+        except KeyError as error:
+            raise InvalidMetadataError(f"No schema found for version {version}") from error
```
benjaminEwhite, line 788: Looks like this path doesn't get executed in tests. Think we should add a test for this case, but no need for me to re-review.

benjaminEwhite (approved):
Looks good. Two minor suggestions that you can resolve on your own without re-review from me @nrh-cklimas.

apps/contributions/models.py
nrh-cklimas: 3cf1f6cb88b105cd51fd2c43705989f702f597d0

apps/contributions/models.py
```diff
@@ -764,6 +772,25 @@ def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.
...
+        try:
+            self.contribution_metadata = json.loads(schema(**(self.contribution_metadata | {key: value})).json())
+        except ValidationError as error:
+            raise InvalidMetadataError(f"Change to {key} results in invalid contribution metadata") from error
```
benjaminEwhite, line 792: Might be worth a code comment here -- think this error message is fine (since the attempt to update the key in question would indeed be the immediate dcause), but it's logically possible that the key we're trying to update the value for is not the culprit -- there's no guarantee that the value in db for contribution_metadata is valid, and outstanding invalid dict entries could actually be what trigger this error.

nrh-cklimas: a9b451a65380ae10c9fe3a91760296ca7f39efd7

apps/contributions/models.py
```diff
@@ -764,6 +772,28 @@ def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.
...
+            schema = STRIPE_PAYMENT_METADATA_SCHEMA_VERSIONS[version]
+        except KeyError as error:
+            raise InvalidMetadataError(f"No schema found for version {version}") from error
+        try:
+            self.contribution_metadata = json.loads(schema(**(self.contribution_metadata | {key: value})).json())
+        except ValidationError as error:
+            # To be specific, this might *not* be the root cause. If there is
+            # pre-existing incorrect metadata, then validation would fail here.
+            # too.
+            raise InvalidMetadataError(f"Change to {key} results in invalid contribution metadata") from error
```
njharman, line 795: Consider adding `value` to error message. [Seems useful in diagnosing if bad value vs bad key fails validation]

apps/contributions/models.py
```diff
@@ -764,6 +772,28 @@ def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.
...
         )
 
+    def set_metadata_field(self, key: str, value: Any) -> None:
+        """Set a field in contribution_metadata, ensuring that the result is valid according to its schema version.
+
+        If an invalid key or value is set, this will raise an InvalidMetadataError exception and contribution_metadata will not be changed.
+        This doesn't make any changes in Stripe.
+        """
+        if key == "schema_version":
+            raise InvalidMetadataError("Schema version may not be changed")
```
njharman, line 782: When `InvalidMetadataError `is contextlib.suppressed (below), I assume Sentry never sees it? Thus there is no log of it ever happening.

Do any of error these conditions warrant logging? Perhaps just debug?  And if so, log here? or let the caller decide to log or not (by catching logging and possibly ignoring exception)?

apps/contributions/models.py
```diff
@@ -1271,7 +1301,12 @@ def update_subscription_amount(self, amount: int) -> None:
             raise ValueError("Subscription should have only one item")
 
         item = items["data"][0]
-        self.contribution_metadata["donor_selected_amount"] = amount
+
+        # Set the amount in metadata if the schema supports it.
+
```
njharman, line 1306: no empty line between comment and code it is commenting

njharman (approved):

apps/contributions/models.py
nrh-cklimas: I think this could possibly put data we don't want into logs, so skipping for now.

apps/contributions/models.py
nrh-cklimas: My approach here is to let the caller decide, but could re-evaluate if it ends up being a common pattern.

apps/contributions/models.py
nrh-cklimas: bc3e86db89e74f522f38e00ae89a8cd08a564d5e

apps/contributions/models.py
```diff
@@ -764,6 +772,28 @@ def send_recurring_contribution_email_reminder(self, next_charge_date: datetime.
...
+            raise InvalidMetadataError("No schema version set in metadata")
+        try:
+            schema = STRIPE_PAYMENT_METADATA_SCHEMA_VERSIONS[version]
+        except KeyError as error:
+            raise InvalidMetadataError(f"No schema found for version {version}") from error
+        try:
+            self.contribution_metadata = json.loads(schema(**(self.contribution_metadata | {key: value})).json())
+        except ValidationError as error:
+            # To be specific, this might *not* be the root cause. If there is
+            # pre-existing incorrect metadata, then validation would fail here.
```
x110dc, line 793: super minor: extraneous period in comment

x110dc (approved):
Noted one tiny typo in a comment. Otherwise LGTM.

PR #1647, opened 2024-10-24
DEV-5230: Return 404 for invalid subdomains

GuiMend:

n/a
#### What's this PR do?
- Makes Django return 404 in the browser for unknown for RP slugs (invalid subdomains in portal)
- Making a hard 404 (page status returns 404) so that google stops indexing inexistent pages
- yes
- no
- yes
- n/a
- no
- no
- https://news-revenue-hub.atlassian.net/browse/DEV-5230
-no
- no

### DIFF
--- a/revengine/tests/test_views.py
+++ b/revengine/tests/test_views.py
@@ -19,8 +19,27 @@ def test_spa_revenue_program_does_not_exist(logger_info, client):
         mock.patch("revengine.views.RevenueProgram.objects.get", side_effect=RevenueProgram.DoesNotExist),
     ):
         response = client.get(reverse("index"))
-        # ReactAppView._get_revenue_program_from_subdomain() just logs.
+        # ReactAppView._get_revenue_program_from_subdomain() logs and
+        # ReactAppView._is_valid_rp_subdomain is False, so 404 status code.
         assert logger_info.was_called
+        assert response.status_code == 404
+
+
+@pytest.mark.django_db
+def test_spa_revenue_program_exist(client):
+    with (
+        mock.patch("revengine.views.get_subdomain_from_request", return_value="subdomain"),
+        mock.patch("revengine.views.RevenueProgram.objects.get", return_value=RevenueProgramFactory()),
+    ):
+        response = client.get(reverse("index"))
+        # ReactAppView._is_valid_rp_subdomain is True, so no 404 status code.
+        assert response.status_code == 200
+
+
+def test_no_subdomains(client):
+    with (mock.patch("revengine.views.get_subdomain_from_request", return_value=None),):
+        response = client.get(reverse("index"))
+        # ReactAppView._is_valid_rp_subdomain defaults to True, so no 404 status code.
         assert response.status_code == 200
 
 
--- a/revengine/views.py
+++ b/revengine/views.py
@@ -1,5 +1,4 @@
 import logging
-import sys
 
 from django.apps import apps
 from django.conf import settings
@@ -26,6 +25,17 @@ logger = logging.getLogger(f"{settings.DEFAULT_LOGGER}.{__name__}")
 # Serve Single Page Application
 class ReactAppView(TemplateView):
     template_name = "index.html"
+
+    def get(self, request, *args, **kwargs):
+        # Check if the subdomain exist and is valid
+        is_valid_subdomain = self._is_valid_rp_subdomain()
+
+        # Render index.html but with a 404 status code if the subdomain is invalid.
+        # Let the SPA handle this instead of Django.
+        response = super().get(request, *args, **kwargs)
+        if not is_valid_subdomain:
+            response.status_code = 404
+        return response
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
@@ -57,6 +67,18 @@ class ReactAppView(TemplateView):
     def _add_gtm_id(self, context):
         context["gtm_id"] = settings.HUB_GTM_ID
 
+    def _is_valid_rp_subdomain(self):
+        if subdomain := get_subdomain_from_request(self.request):
+            try:
+                RevenueProgram.objects.get(slug=subdomain)
+            except RevenueProgram.DoesNotExist:
+                return False
+            else:
+                return True
+
+        # If there is no subdomain, default to True
+        return True
+
 
 # Proxies the single page app in local development, parsing any HTML responses
 # using Django's template parser. If this is used, the single page app must have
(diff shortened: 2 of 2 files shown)


revengine/views.py
```diff
@@ -57,15 +67,27 @@ def _add_social_media_context(self, revenue_program, context):
...
+    def _is_valid_rp_subdomain(self):
+        if subdomain := get_subdomain_from_request(self.request):
+            try:
+                RevenueProgram.objects.get(slug=subdomain)
+            except RevenueProgram.DoesNotExist:
+                return False
+            else:
+                return True
+
+        # If there is no subdomain, default to True
```
nrh-cklimas, line 79: I think it'd be helpful to note here that a reason why we might not get a subdomain here is that it's a non-RP subdomain like `engine.fundjournalism.org`. Had to go looking myself because I was curious about it.

revengine/views.py
```diff
@@ -27,6 +26,17 @@
...
     template_name = "index.html"
 
+    def get(self, request, *args, **kwargs):
+        # Check if the subdomain exist and is valid
+        is_valid_subdomain = self._is_valid_rp_subdomain()
+
+        # Render index.html but with a 404 status code if the subdomain is invalid.
+        # Let the SPA handle this instead of Django.
+        response = super().get(request, *args, **kwargs)
+        if not is_valid_subdomain:
```
nrh-cklimas, line 36: Could simplify to

```python
if not self._is_valid_rp_subdomain():
```

Since you don't use `is_valid_subdomain` elsewhere in the method.

nrh-cklimas (approved):

revengine/tests/test_views.py
```diff
@@ -19,8 +19,27 @@ def test_spa_revenue_program_does_not_exist(logger_info, client):
...
+
+@pytest.mark.django_db
+def test_spa_revenue_program_exist(client):
+    with (
+        mock.patch("revengine.views.get_subdomain_from_request", return_value="subdomain"),
+        mock.patch("revengine.views.RevenueProgram.objects.get", return_value=RevenueProgramFactory()),
+    ):
+        response = client.get(reverse("index"))
+        # ReactAppView._is_valid_rp_subdomain is True, so no 404 status code.
+        assert response.status_code == 200
```
benjaminEwhite, line 36: Let's use the `mocker` fixture instead of the imported mock library

```suggestion
@pytest.mark.django_db
def test_spa_revenue_program_exist(client, mocker):

    mocker.patch("revengine.views.get_subdomain_from_request", return_value="subdomain")
    mocker.patch("revengine.views.RevenueProgram.objects.get", return_value=RevenueProgramFactory())
    response = client.get(reverse("index"))
    # ReactAppView._is_valid_rp_subdomain is True, so no 404 status code.
    assert response.status_code == 200

```

benjaminEwhite, line 36: Let's use the `mocker` fixture instead of the imported mock library

```suggestion
@pytest.mark.django_db
def test_spa_revenue_program_exist(client, mocker):

    mocker.patch("revengine.views.get_subdomain_from_request", return_value="subdomain")
    mocker.patch("revengine.views.RevenueProgram.objects.get", return_value=RevenueProgramFactory())
    response = client.get(reverse("index"))
    # ReactAppView._is_valid_rp_subdomain is True, so no 404 status code.
    assert response.status_code == 200

```

revengine/tests/test_views.py
```diff
@@ -19,8 +19,27 @@ def test_spa_revenue_program_does_not_exist(logger_info, client):
...
+
+
+def test_no_subdomains(client):
+    with (mock.patch("revengine.views.get_subdomain_from_request", return_value=None),):
```
benjaminEwhite, line 40: Ditto here, should be able to use `mocker.patch` instead

revengine/views.py
```diff
@@ -57,15 +64,28 @@ def _add_social_media_context(self, revenue_program, context):
...
+        if subdomain := get_subdomain_from_request(self.request):
+            try:
+                RevenueProgram.objects.get(slug=subdomain)
+            except RevenueProgram.DoesNotExist:
+                return False
+            else:
+                return True
+
+        # If there is no subdomain, it is a non-RP subdomain like "engine.fundjournalism.org"
+        # so we default to True
```
benjaminEwhite, line 77: Given implementation of `get_subdomain_from_request` it seems possible that you could end up with edge cases here (but to be clear I haven't pulled thread enough here to see if this is a real problem):

1. `get_subdomain_from_request` could encounter a domain that's in HOST_MAP, but that has an empty value.
2. `get_subdomain_from_request` has this code, which might have some problematic assumptions:

```
if len(split_host) > 2 and split_host[0] not in settings.DASHBOARD_SUBDOMAINS:
        subdomain = split_host[0]
```

May turn out there's nothing required here, but seems worth spending some effort to see if might be any edge cases to consider.

revengine/views.py
```diff
@@ -57,15 +64,28 @@ def _add_social_media_context(self, revenue_program, context):
...
 # Proxies the single page app in local development, parsing any HTML responses
 # using Django's template parser. If this is used, the single page app must have
 # its own dev proxying turned off for this to work.
 #
 # This was cribbed from
 # https://fractalideas.com/blog/making-react-and-django-play-well-together-hybrid-app-model/
-
-
+#
+# This doesn't currently work with Vite as local server.
```
benjaminEwhite, line 88: Does this mean that this function is effectively broken in status quo? If so, we should have a TODO + ticket, IMO.

revengine/views.py
```diff
@@ -86,10 +106,7 @@ def proxy_spa_dev_server(request, upstream="http://localhost:3000"):
     )
 
 
-# Only use the dev proxy if we are running locally AND not in pytest. We want
-# the normal view to be tested in pytest.
-
-index = proxy_spa_dev_server if settings.DEBUG and "pytest" not in sys.modules else never_cache(ReactAppView.as_view())
+index = never_cache(ReactAppView.as_view())
```
benjaminEwhite, line 109: Why is this change necessary now?

benjaminEwhite (changes requested):

revengine/views.py
GuiMend: Done https://news-revenue-hub.atlassian.net/browse/DEV-5405.
c/c @nrh-cklimas

revengine/views.py
GuiMend: I've undone this change and created a new ticket to address this modification

revengine/views.py
GuiMend: I've looked at this and haven't found an issue so far. I added a `logger.warning`  when the subdomain is not found so that we can track it better in case something unexpected happens. What do you think?

benjaminEwhite (approved):

PR #1648, opened 2024-10-25
DEV-5377 | ensure pm is stripe payment method and don't crash when not

benjaminEwhite:

### DIFF
--- a/apps/contributions/management/commands/fix_contributions_missing_provider_payment_method_id.py
+++ b/apps/contributions/management/commands/fix_contributions_missing_provider_payment_method_id.py
@@ -103,14 +103,17 @@ class Command(BaseCommand):
 
     def get_pm_from_subscription(self, subscription: stripe.Subscription) -> stripe.PaymentMethod | None:
         """Get payment method from subscription."""
-        if pm := subscription.get("default_payment_method"):
-            return pm
-        if (invoice_settings := subscription.get("customer", {}).get("invoice_settings")) and (
-            pm := invoice_settings.get("default_payment_method")
-        ):
-            return pm
-        if (invoice := subscription.get("latest_invoice")) and (pi := invoice.get("payment_intent")):
-            return pi.get("payment_method")
+        self.stdout.write(self.style.HTTP_INFO(f"Getting payment method from subscription {subscription.id}"))
+        pm = subscription.get("default_payment_method")
+        if not pm and (invoice_settings := subscription.get("customer", {}).get("invoice_settings")):
+            pm = invoice_settings.get("default_payment_method")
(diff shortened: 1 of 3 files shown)


apps/contributions/management/commands/fix_contributions_missing_provider_payment_method_id.py
```diff
@@ -103,14 +103,17 @@ def _update_and_save(self, contribution: Contribution, pm: stripe.PaymentMethod)
...
+        self.stdout.write(self.style.HTTP_INFO(f"Getting payment method from subscription {subscription.id}"))
+        pm = subscription.get("default_payment_method")
+        if not pm and (invoice_settings := subscription.get("customer", {}).get("invoice_settings")):
+            pm = invoice_settings.get("default_payment_method")
+        if not pm and (invoice := subscription.get("latest_invoice")) and (pi := invoice.get("payment_intent")):
+            pm = pi.get("payment_method")
+        if not isinstance(pm, stripe.PaymentMethod):
+            self.stderr.write(
+                f"Expected a stripe.PaymentMethod object but got {type(pm)} for subscription {subscription.id}. Skipping"
+            )
```
nrh-cklimas, line 115: Is this redundant given line 139?

apps/contributions/management/commands/fix_contributions_missing_provider_payment_method_id.py
```diff
@@ -129,11 +132,18 @@ def process_contribution_via_retrieve_api(self, contribution: Contribution) -> t
                 if contribution.interval != ContributionInterval.ONE_TIME
                 else payment_object.get("payment_method")
             )
-            if pm:
+            if pm and isinstance(pm, stripe.PaymentMethod):
```
nrh-cklimas, line 135: This is a nitpick, but I'd consider something like this:

```python
if pm:
  if isinstance(pm, stripe.PaymentMethod):
    # update and save
  else:
    # warn about bad type
else:
   # warn about faiing to retrieve
```

This is effectively the same as what you've written, but to me, makes it clearer only one of these three clauses could ever run.

nrh-cklimas (approved):

apps/contributions/management/commands/fix_contributions_missing_provider_payment_method_id.py
benjaminEwhite: In principle, yes, but that could change in future, and wanted to be defensive.

apps/contributions/management/commands/fix_contributions_missing_provider_payment_method_id.py
benjaminEwhite: e539cb8a28f5

x110dc (approved):

PR #1650, opened 2024-11-01
DEV-5356: specify python-version

leo-dcfa:

N/A

#### What's this PR do?

Replace runtime.txt with .python-version. (only major version as recommended by Heroku)

Usage of runtime.txt is deprecated and .python-version is recommended https://devcenter.heroku.com/articles/python-runtimes#:~:text=The%20runtime.,tooling%20in%20the%20Python%20ecosystem.

N/A

N/A

N/A

N/A

N/A

N/A

[DEV-5356]

N/A

N/A

[DEV-5356]: https://news-revenue-hub.atlassian.net/browse/DEV-5356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.python-version
```diff
@@ -0,0 +1 @@
+3.10
```
x110dc, line 1: I think we should specify the full version here. If we don't then 1. our deployed python version may not match the python version used for the rest of our tooling. It also means we could be running a different version of python in PROD than all of our Heroku environments.

x110dc (changes requested):

.python-version
leo-dcfa: Fair point 06f0066e6eb34c11fce0ab11db735e890b0c7103 - also specified explicit version on a couple other files

x110dc (approved):
lgtm

njharman (approved):

PR #1656, opened 2024-11-07
Dev 5429 set amount metadata correctly

nrh-cklimas:

n/a

#### What's this PR do?

- Adds a `donor_selected_amount` that is only available when PATCHing an individual contribution. Sending a value here updates the contribution metadata.
- Changes the SPA to send both `amount` and `donor_selected_amount`.

Corrects a problem where donor_selected_amount is getting to set to an amount in cents rather than dollars.

I went with this approach to minimize the amount of math we are doing to amounts, to avoid possible problem sources re: rounding.

Yes.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5429

No.

No.

### DIFF
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -1264,11 +1264,13 @@ class Contribution(IndexedTimeStampedModel):
 
             raise
 
-    def update_subscription_amount(self, amount: int) -> None:
-        """Update Stripe's Subscription Item amount to the new value.
+    def update_subscription_amount(self, amount: int, donor_selected_amount: float) -> None:
+        """Update the item amount and donor-selected amount (in metadata) of the Stripe subscription of this contribution.
 
-        If it's a recurring subscription, update the amount of the next payment
-        without proration (paying difference of existing month).
+        **amount is in cents, but donor_selected_amount is in dollars.** This difference is because amount is tracked in Stripe
+        natively as cents, while donor_selected_amount is a metadata field we added that uses float dollars.
+
+        This doesn't prorate the change (e.g. paying difference of existing month next time).
         """
         # vs circular import
         from apps.contributions.serializers import REVENGINE_MIN_AMOUNT, STRIPE_MAX_AMOUNT
@@ -1302,9 +1304,9 @@ class Contribution(IndexedTimeStampedModel):
 
         item = items["data"][0]
 
-        # Set the amount in metadata if the schema supports it.
+        # Set the donor-selected amount in metadata if the schema supports it.
         with contextlib.suppress(InvalidMetadataError):
-            self.set_metadata_field("donor_selected_amount", amount)
+            self.set_metadata_field("donor_selected_amount", donor_selected_amount)
 
         self.amount = amount
 
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -1267,10 +1267,13 @@ class Contribution(IndexedTimeStampedModel):
     def update_subscription_amount(self, amount: int, donor_selected_amount: float) -> None:
         """Update the item amount and donor-selected amount (in metadata) of the Stripe subscription of this contribution.
 
-        **amount is in cents, but donor_selected_amount is in dollars.** This difference is because amount is tracked in Stripe
-        natively as cents, while donor_selected_amount is a metadata field we added that uses float dollars.
+        **amount is in cents, but donor_selected_amount is in dollars.** This
+        difference is because amount is tracked in Stripe natively as cents,
+        while donor_selected_amount is a metadata field we added that uses float
+        dollars. This field is persisted in Stripe as a string.
 
-        This doesn't prorate the change (e.g. paying difference of existing month next time).
+        This doesn't prorate the change (e.g. paying difference of existing
+        month next time).
         """
         # vs circular import
         from apps.contributions.serializers import REVENGINE_MIN_AMOUNT, STRIPE_MAX_AMOUNT
--- a/apps/contributions/models.py
+++ b/apps/contributions/models.py
@@ -1274,6 +1274,8 @@ class Contribution(IndexedTimeStampedModel):
 
         This doesn't prorate the change (e.g. paying difference of existing
         month next time).
+
+        TODO in DEV-5465: improved validation of donor_selected_amount in Pydantic
         """
         # vs circular import
         from apps.contributions.serializers import REVENGINE_MIN_AMOUNT, STRIPE_MAX_AMOUNT
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -991,6 +991,7 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
                 instance.update_payment_method_for_subscription(
                     provider_payment_method_id=provider_payment_method_id,
                 )
+            # Need to pop donor_selected_amount so it doesn't get sent when saving changes.
             donor_selected_amount = validated_data.pop("donor_selected_amount", None)
             if amount := validated_data.get("amount", None):
                 # We can trust that donor_selected_amount is not None because of
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -993,9 +993,14 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
                 )
             # Need to pop donor_selected_amount so it doesn't get sent when saving changes.
             donor_selected_amount = validated_data.pop("donor_selected_amount", None)
+
             if amount := validated_data.get("amount", None):
-                # We can trust that donor_selected_amount is not None because of
-                # the check in validate().
+                if not donor_selected_amount:
+                    # amount and donor_selected_amount should always be set in tandem in real life thanks to validate(),
+                    # but we want to be certain.
+                    raise serializers.ValidationError(
+                        "If amount is updated, donor_selected_amount must be set as well."
+                    )
                 instance.update_subscription_amount(amount=amount, donor_selected_amount=donor_selected_amount)
             for key, value in validated_data.items():
                 setattr(instance, key, value)
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -963,7 +963,8 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
             "min_value": f"We can only accept contributions greater than or equal to {format_ambiguous_currency(REVENGINE_MIN_AMOUNT)}",
         },
     )
-    # Note that donor_selected_amount is float dollars, not int cents.
+    # Note that donor_selected_amount is float dollars, not int cents. This
+    # value is persisted in Stripe as a string.
     donor_selected_amount = serializers.FloatField(
         required=False,
         min_value=REVENGINE_MIN_AMOUNT / 100,
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -992,7 +992,9 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
                     provider_payment_method_id=provider_payment_method_id,
                 )
             donor_selected_amount = validated_data.pop("donor_selected_amount", None)
-            if donor_selected_amount and (amount := validated_data.get("amount", None)):
+            if amount := validated_data.get("amount", None):
+                # We can trust that donor_selected_amount is not None because of
+                # the check in validate().
                 instance.update_subscription_amount(amount=amount, donor_selected_amount=donor_selected_amount)
             for key, value in validated_data.items():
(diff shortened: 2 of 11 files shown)


apps/contributions/models.py
```diff
@@ -1302,9 +1304,9 @@ def update_subscription_amount(self, amount: int) -> None:
 
         item = items["data"][0]
 
-        # Set the amount in metadata if the schema supports it.
+        # Set the donor-selected amount in metadata if the schema supports it.
         with contextlib.suppress(InvalidMetadataError):
-            self.set_metadata_field("donor_selected_amount", amount)
+            self.set_metadata_field("donor_selected_amount", donor_selected_amount)
```
nrh-cklimas, line 1309: I'm not sure what tests, if any, we need on this.

apps/contributions/models.py
```diff
@@ -1264,11 +1264,13 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
-        """Update Stripe's Subscription Item amount to the new value.
+    def update_subscription_amount(self, amount: int, donor_selected_amount: float) -> None:
+        """Update the item amount and donor-selected amount (in metadata) of the Stripe subscription of this contribution.
 
-        If it's a recurring subscription, update the amount of the next payment
-        without proration (paying difference of existing month).
+        **amount is in cents, but donor_selected_amount is in dollars.** This difference is because amount is tracked in Stripe
+        natively as cents, while donor_selected_amount is a metadata field we added that uses float dollars.
+
+        This doesn't prorate the change (e.g. paying difference of existing month next time).
```
benjaminEwhite, line 1273: It might be noting in this comment the following oddity: although our `StripePaymentMetadataSchema...` classes treat `donor_selected_amount` as a float when representing in revengine, that value gets converted to a string when it's peristed in Stripe's metadata.

benjaminEwhite: This makes me wonder if we should consider a follow up ticket that refactors our conversion/validation logic for donor selected amount in revengine. 

We could have our pydantic model validate that donor selected amount can be be cast to a float before ultimately storing casting it a string value, which is what its terminal state will be when stored in Stripe metadata.

apps/contributions/serializers.py
```diff
@@ -962,10 +963,25 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
             "min_value": f"We can only accept contributions greater than or equal to {format_ambiguous_currency(REVENGINE_MIN_AMOUNT)}",
         },
     )
+    # Note that donor_selected_amount is float dollars, not int cents.
```
benjaminEwhite, line 966: The nuance around string conversion also worth noting here.

apps/contributions/serializers.py
```diff
@@ -974,17 +990,28 @@ def update(self, instance: Contribution, validated_data) -> Contribution:
                 instance.update_payment_method_for_subscription(
                     provider_payment_method_id=provider_payment_method_id,
                 )
-            if amount := validated_data.get("amount", None):
-                instance.update_subscription_amount(
-                    amount=amount,
-                )
+            donor_selected_amount = validated_data.pop("donor_selected_amount", None)
+            if donor_selected_amount and (amount := validated_data.get("amount", None)):
```
benjaminEwhite, line 994: Think we should remove conditionality and just assume that if amount that there is donor selected amount. Add a code comment here that that is expected to be guaranteed by our override of `.validate`, and this code only runs if we have valid data.

benjaminEwhite: Also, noting that the conditionality at line 988 shouldn't be required. This update method will only run if the data is valid. See `.handle_patch` in `apps.contributions.views.py`:

```
def handle_patch(self, serializer: serializers.PortalContributionDetailSerializer) -> Response:
        serializer.is_valid(raise_exception=True)
        try:
            serializer.save()
        except stripe.error.StripeError:
            return Response({"detail": "Problem updating contribution"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
        return Response(serializer.data, status=status.HTTP_200_OK)

```


When .save() gets called, it (by default) calls .update() on the serializer.

apps/contributions/serializers.py
```diff
@@ -974,17 +990,28 @@ def update(self, instance: Contribution, validated_data) -> Contribution:
...
+    def validate(self, data):
+        data = super().validate(data)
+        if "amount" in data and "donor_selected_amount" not in data:
+            raise serializers.ValidationError(
+                {"amount": "If this field is updated, donor_selected_amount must be provided as well."}
+            )
+        if "donor_selected_amount" in data and "amount" not in data:
+            raise serializers.ValidationError(
+                {"donor_selected_amount": "If this field is updated, amount must be provided as well."}
+            )
```
benjaminEwhite, line 1012: Minor, but think it's worthwhile to put this logic into separate named function both so it's self-documenting, and so if we need to add further "late" validation, we can extend that pattern of organizing into separate self-documenting, named functions. 

Otherwise, future maintainer needs to read this entire block of code and then reach the right conclusion that the purpose is to ensure both fields if one is there.

apps/contributions/models.py
nrh-cklimas: 77b994b336c96ed0edea24c24a686b5039c25846 for docs. I think I'm onboard with beefing up the Pydantic model, but what would happen if the metadata was set to a non-float value externally? Or is that just something we need to live with using Pydantic in general.

apps/contributions/serializers.py
nrh-cklimas: 77b994b336c96ed0edea24c24a686b5039c25846

apps/contributions/serializers.py
benjaminEwhite: Looking at this code block more, I see why the conditionality is there at line 988. Looks like I wrote that line, and I'm guessing it's because it's logically possible for validated_data to be an empty dict (we don't stop someone from patching an empty dict since there are no required keys in the serializer for write ops). 

I'm going to close this conversation and will look at that commit in non-stale conversation thread.

benjaminEwhite (commented):
Resolved some of the conversations, and had follow up comments in some others.

apps/contributions/models.py
nrh-cklimas: Not sure why it didn't autolink, but try https://github.com/newsrevenuehub/rev-engine/pull/1656/commits/77b994b336c96ed0edea24c24a686b5039c25846. Yes, I meant a user editing manually in the dashboard or something like that.

apps/contributions/serializers.py
nrh-cklimas: https://github.com/newsrevenuehub/rev-engine/pull/1656/commits/77b994b336c96ed0edea24c24a686b5039c25846

apps/contributions/serializers.py
```diff
@@ -974,17 +990,28 @@ def update(self, instance: Contribution, validated_data) -> Contribution:
                 instance.update_payment_method_for_subscription(
                     provider_payment_method_id=provider_payment_method_id,
                 )
-            if amount := validated_data.get("amount", None):
-                instance.update_subscription_amount(
-                    amount=amount,
-                )
+            donor_selected_amount = validated_data.pop("donor_selected_amount", None)
+            if donor_selected_amount and (amount := validated_data.get("amount", None)):
+                instance.update_subscription_amount(amount=amount, donor_selected_amount=donor_selected_amount)
```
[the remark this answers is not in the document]
nrh-cklimas: I think we need to pop `donor_selected_amount` in any case since it's not a model field?

apps/contributions/models.py
benjaminEwhite: Got it re: external. That makes sense and we'll need to account for it.

I created https://news-revenue-hub.atlassian.net/browse/DEV-5465 as a follow up ticket to pursue this. Do you mind adding as a TODO for this branch?

No need for me to re-review, you can resolve.

apps/contributions/serializers.py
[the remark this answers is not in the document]
benjaminEwhite: Good point. We don't actually need it popped as far as the model is concerned (if i'm not mistaken, you can set arbitrary properties on a model instance and then save it and system won't complain).  But we _do_ need it so we don't send as an update field when calling .save(). 

Think it's worth leaving a code comment for line where donor selected amount is popped from dictionary so it's obvious why we're doing that.

No need for me to re-review to resolve this.

benjaminEwhite (approved):
Leaving two conversations around that you can resolve without re-review from me @nrh-cklimas and approving.

apps/contributions/serializers.py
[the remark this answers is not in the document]
nrh-cklimas: 6a21ad7ceec87fdbd367238d8ca72263bbd07720

apps/contributions/tests/test_models.py
```diff
@@ -1741,7 +1741,7 @@ def test_update_payment_method_for_subscription_card_declined_error_does_not_sen
     def test_update_subscription_amount_when_one_time(self, one_time_contribution: Contribution):
         one_time_contribution.stripe_subscription = MockSubscription("active")
         with pytest.raises(ValueError, match="Cannot update amount for one-time contribution"):
-            one_time_contribution.update_subscription_amount(amount=100)
+            one_time_contribution.update_subscription_amount(amount=100, donor_selected_amount=1)
```
njharman, line 1744: Are `amount `and` donor_selected_amount` required to be identical? All the tests seem to imply so, but I didn't notice any validation for that.

apps/contributions/tests/test_serializers.py
```diff
@@ -1848,7 +1848,16 @@ def test_update_amount_one_time_contribution(self, mocker, one_time_contribution
         mocker.patch("stripe.Subscription.modify")
         serializer = PortalContributionDetailSerializer(instance=one_time_contribution)
         with pytest.raises(ValueError, match="Cannot update amount for one-time contribution"):
-            serializer.update(one_time_contribution, {"amount": 123})
+            serializer.update(one_time_contribution, {"amount": 123, "donor_selected_amount": 1.23})
+
```
njharman, line 1852: tests 

- for minimum, maximum?
- differing values of amount and donor_selected_amount

apps/contributions/serializers.py
```diff
@@ -974,17 +991,34 @@ def update(self, instance: Contribution, validated_data) -> Contribution:
                 instance.update_payment_method_for_subscription(
                     provider_payment_method_id=provider_payment_method_id,
                 )
+            # Need to pop donor_selected_amount so it doesn't get sent when saving changes.
+            donor_selected_amount = validated_data.pop("donor_selected_amount", None)
             if amount := validated_data.get("amount", None):
-                instance.update_subscription_amount(
-                    amount=amount,
-                )
+                # We can trust that donor_selected_amount is not None because of
```
njharman, line 997: `In tests line 1825         updated_contribution = serializer.update(monthly_contribution, {"amount": 123})`

This does not have donor_selected_amount.  Which means it's value is None. So, we can't really trust it, cause .validate() doesn't have to be called before calling .update()?

I'm getting a little lost trying to trace all the code paths. But, it seems the oversight in test (not including donor_selected_amount) should be "caught", no?

njharman: That test `def test_update_amount_monthly_contribution `should probably include assert for `donor_selected_amount`

njharman (changes requested):

apps/contributions/tests/test_models.py
nrh-cklimas: No, they can and often are different. If the donor chooses to cover fees on a $10 donation, `donor_selected_amount` is `10.0` but `amount` would be `1053` ($10.53). I kept them the same in most tests because both args are required now, but not really relevant to most tests.

apps/contributions/serializers.py
nrh-cklimas: I talked with Ben about this because it delves in DRF stuff I don't feel that familiar with yet. The feeling I have is, it falls under "things that should never happen in real life", because DRF ensures validate() is called before update(). However, it's still true that update() could be called directly and it will cause data problems if it was. So I added a check in 9a6b8fda35b93ac06b3a2547bd126c9c6d8adfec.

I also added an assertion on `donor_selected_amount` in the test that runs through a successful update in the same commit.

njharman: Yeah, I had thought that there was DRF "magic" to ensure data was validated before validated_data could be used.  

Like I thought this was automatically called.  https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data

There's probably a good reason update takes validated_data as parameter instead of using self.validated_data. But seems error prone, as this test showed.

njharman (approved):

x110dc (approved):

PR #1661, opened 2024-11-19
DEV-5338 | PoC of subscription reuse on amount/frequency change

benjaminEwhite:

### DIFF
--- a/apps/contributions/serializers.py
+++ b/apps/contributions/serializers.py
@@ -968,7 +968,9 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
     class Meta:
         model = Contribution
         fields = [*PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS, "provider_payment_method_id"]
-        read_only_fields = PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS
+        read_only_fields = [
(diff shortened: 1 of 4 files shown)


apps/contributions/serializers.py
```diff
@@ -962,23 +963,89 @@ class PortalContributionDetailSerializer(PortalContributionBaseSerializer):
...
+                raise ValueError("Amount value must be greater than $0.99")
+            if new_amount > STRIPE_MAX_AMOUNT:
+                raise ValueError("Amount value must be smaller than $999,999.99")
+        if new_interval is not None and new_interval not in (
+            ContributionInterval.MONTHLY.value,
+            ContributionInterval.YEARLY.value,
+        ):
+            raise ValueError("Invalid interval value")
+
+    def create_stripe_price(
```
x110dc, line 999: I'm curious about creating a new price here. Was there an opportunity to update the existing one? I ask because I don't think we've had reason to directly create a price before so not sure how that could/should work.

x110dc (commented):
I know this is a POC so won't make it into PROD as-is, but just had one comment.

apps/contributions/serializers.py
benjaminEwhite: @x110dc may be possible to reususe existing price and modify it:

https://docs.stripe.com/api/prices/update

I didn't explore that specifically and just tried to string together "whatever worked quickest" for purpose of PoC. If we can reuse price in actual implementation, that's a plus.

benjaminEwhite:
Closing this PR because it was around only for PoC, and has served its purpose. We'll keep the branch around so it can be reused in implementation.

PR #1665, opened 2024-11-22
DEV-5416: Handle customer name edge case

GuiMend:

n/a

#### What's this PR do?

Handles when customer doesn't have "name" attribute (vs. having a null value for that attribute).  This can happen when a Stripe customer is initially created without a name. In that case, there won't be a "name" property on the customer object.

- Get rid of bad salutation (Dear None) in transactional emails
- Handle exception so that error isn't thrown

yes

no

yes

n/a

no

no

https://news-revenue-hub.atlassian.net/browse/DEV-5416

If so, open a ticket and link it to the ticket for this PR and list it here:

no

no


apps/emails/tasks.py
```diff
@@ -153,7 +153,7 @@ def generate_email_data(
         ),
         contribution_interval=contribution.interval,
         contributor_email=contribution.contributor.email,
-        contributor_name=customer.name or CONTRIBUTOR_DEFAULT_VALUE,
+        contributor_name=getattr(customer, "name", CONTRIBUTOR_DEFAULT_VALUE) or CONTRIBUTOR_DEFAULT_VALUE,
```
benjaminEwhite, line 156:
```suggestion
        # customer.name could be non truthy, but if customer was created without providing value for `name`,
        # there will not be a `customer.name` attribute. This ensures we get default value in 
        # both cases.
        contributor_name=getattr(customer, "name", CONTRIBUTOR_DEFAULT_VALUE) or CONTRIBUTOR_DEFAULT_VALUE,
```

apps/emails/tests/test_tasks.py
```diff
@@ -64,13 +65,27 @@ def contribution(self, request):
         return request.getfixturevalue(request.param)
 
     @pytest.mark.parametrize("show_billing_history", [False, True])
-    @pytest.mark.parametrize("contributor_name", ["customer_name", None])
+    @pytest.mark.parametrize(
+        ("contributor_name", "expected_name"),
+        [
+            (AttrDict(name="customer_name"), "customer_name"),
+            (AttrDict(name=None), CONTRIBUTOR_DEFAULT_VALUE),
+            (None, CONTRIBUTOR_DEFAULT_VALUE),
```
njharman, line 73: Does AttrDict() not work? It would more close match `customer `not having `name `attribute.

Also, in my opinion easier to understand test / what test is testing.  As is, it appears it's testing when obj` is None`. When it's just relying on fact that `None `has no "name" attribute.

benjaminEwhite: f5e4abe1e 

Also updated the parameter name to customer instead from contributor_name

PR #1672, opened 2024-12-05
DEV-3143: Remove Anymail

nrh-cklimas:

n/a

#### What's this PR do?

Removes the django-anymail package and uses Django's built-in SMTP handling instead.

Reduces complexity, gives us flexibility to change email providers more easily.

Yes.

No.

Updated a few tests that used Anymail-specific mocks.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3143

⚠️ https://news-revenue-hub.atlassian.net/browse/ITS-2688 needs to be done before this merges.

There is also https://news-revenue-hub.atlassian.net/browse/ITS-2689 for cleanup after this merges and we're confident it's working.


revengine/settings/deploy.py
```diff
@@ -16,15 +16,13 @@
...
-EMAIL_BACKEND = "anymail.backends.mailgun.EmailBackend"
-MAILGUN_SENDER_DOMAIN = "fundjournalism.org"
-
-ANYMAIL = {
-    "MAILGUN_API_KEY": os.getenv("MAILGUN_API_KEY", ""),
-    "MAILGUN_SENDER_DOMAIN": MAILGUN_SENDER_DOMAIN,
-}
-
+EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
+EMAIL_HOST = os.getenv("EMAIL_HOST")
```
njharman, line 20: Consider raising exception or critical log if required env vars are not (present and non-empty). os.getenv returns None on default and we'd want to look for empty string.

Those listed in [ITS-2688]

njharman (approved):

revengine/settings/deploy.py
nrh-cklimas: Leaving this here for future reviewers to weigh in on. My feeling is that if we have email misconfigured, we're going to find out pretty fast regardless.

benjaminEwhite (approved):

revengine/settings/deploy.py
x110dc: I do think it's a good practice to fail quickly if expected environment variables don't exist, which is one of the reasons we use `environs` on the Switchboard side (as that's the default with that package). But I'm okay deferring this to a separate ticket where we implement that package for RevEngine (or equivalent changes). If there's not already a ticket for that can you open one and add a code comment about it here @nrh-cklimas ?

x110dc (approved):
I'd also add to the "why are we doing this" section that this change makes it easier for us to change email providers. And that was the main point of this ticket.

PR #1699, opened 2025-01-16
DEV-5564: Add mailchimp_connection_ready

nrh-cklimas:

n/a

#### What's this PR do?

Adds a new property to revenue programs, `mailchimp_integration_ready`. This is true only when all MC setup is done.

Switchboard was using `mailchimp_integration_connected`, which is true even when the integration is partially completed, or if it has become partially disconnected.

Yes.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5564

No.

- https://news-revenue-hub.atlassian.net/browse/DEV-5687
- https://news-revenue-hub.atlassian.net/browse/DEV-5688


apps/organizations/models.py
```diff
@@ -930,9 +930,24 @@ def mailchimp_all_contributors_segment_name(self):
...
     @property
     def mailchimp_integration_connected(self):
-        """Determine mailchimp connection state for the revenue program."""
+        """Whether the Mailchimp integration has an API token and server prefix (e.g. is ready to be configured by a user).
+
+        Referencing this doesn't cause any API requests to Mailchimp.
+        """
         return all([self.mailchimp_access_token, self.mailchimp_server_prefix])
 
+    @property
```
benjaminEwhite, line 939: Given this incurs 3 API calls to Mailchimp and one to Google Secret manager, probably worth using `@cached_property` decorator instead of plain decorator here.  Not sure if it matters much in current calling context, but if for some reason this value gets referenced more than once in a given request cycle, it would prevent from refetching those values.

```
from functools import cached_property
```

apps/organizations/models.py
```diff
@@ -930,9 +930,24 @@ def mailchimp_all_contributors_segment_name(self):
...
+    def mailchimp_integration_ready(self):
+        """Whether the Mailchimp integration is fully configured. Referencing this causes API requests to Mailchimp to occur."""
```
benjaminEwhite, line 941: Arguably pedantic, but also indicate that it incurs a call to Google Secret manager as well (because behind scenes integration connected property is checking for presence of access token).

apps/organizations/serializers.py
```diff
@@ -187,6 +187,7 @@ class MailchimpRevenueProgramForSwitchboard(serializers.ModelSerializer):
     """Primary consumer is switchboard API. This is a read-only serializer."""
 
     mailchimp_integration_connected = serializers.ReadOnlyField()
+    mailchimp_integration_ready = serializers.ReadOnlyField()
```
benjaminEwhite, line 190: Since SB is only consumer, probably could add a TODO above line 189 to remove that field after DEV-5687

apps/organizations/tests/test_models.py
```diff
@@ -897,6 +898,66 @@ def test_mailchimp_integration_connected_property(
...
+    # These disparate parametrize annotations are to exhaustively test all
+    # possible combinations of values.
+
+    @pytest.mark.parametrize(
+        ("mailchimp_integration_connected"),
+        [(True), (False)],
+    )
+    @pytest.mark.parametrize(("mailchimp_store"), [("truthy"), (None)])
+    @pytest.mark.parametrize(("mailchimp_one_time_contribution_product"), [("truthy"), (None)])
+    @pytest.mark.parametrize(("mailchimp_recurring_contribution_product"), [("truthy"), (None)])
```
benjaminEwhite, line 910:
```suggestion
    @pytest.mark.parametrize(
        "mailchimp_integration_connected",
        [True, False],
    )
    @pytest.mark.parametrize("mailchimp_store", ["truthy", None])
    @pytest.mark.parametrize("mailchimp_one_time_contribution_product", ["truthy", None])
    @pytest.mark.parametrize("mailchimp_recurring_contribution_product", ["truthy", None])
```

benjaminEwhite: FWIW, if your parametrizing a single param, ruff doesn't force you to put in tuple.

apps/organizations/tests/test_models.py
```diff
@@ -897,6 +898,66 @@ def test_mailchimp_integration_connected_property(
...
+    def test_mailchimp_integration_ready_property(
```
benjaminEwhite, line 911: Have mixed feelings about this test. On one hand, I can see value in exhausting state space for these values, on the other hand it's arguably tautological in that the `all` assertion is also what the object code does.  I don't feel strongly about this, so no need to change if you want to keep.

benjaminEwhite (changes requested):
A couple of small requests inline.

Also, given recent conversation about trying to do more testing at API layer, seems like it might make sense to add a view layer test for this endpoint (one doesn't currently exist) and assert that expected fields appear. Think we only need happy path for that endpoint, and that's really the only way we get strong guarantee of the business logic here (namely, that SB can depend on there being a mailchimp_integration_ready property being in response object for this endpoint).

apps/organizations/models.py
nrh-cklimas: Good call, 9b9569854acea503555644761ceb5e86bd70abf1

apps/organizations/models.py
nrh-cklimas: 3b0d3c02797278097d9484dac3d96c64d8179577

apps/organizations/serializers.py
nrh-cklimas: 3a9007e68816788130d51eb3876a18b4a69b69bb

apps/organizations/tests/test_models.py
nrh-cklimas: 8bbdd1b6552426f7b9cee8165cbd47d1430d6cf8

apps/organizations/tests/test_models.py
nrh-cklimas: I had similar feelings... other ideas I had were:

- Write a truth table manually (16 lines long I think)
- Generate the truth table programmatically (same issue as in the test itself, repeating logic)

I don't know, the logic here is so simple that it may not be worth thinking too hard about. I'm open to suggestions.

benjaminEwhite: Agree about not overthinking. While Wittgenstein would appreciate us writing truth tables, he's dead anyways and so our efforts would be wasted.

benjaminEwhite (approved):
Noting that Chris and I discussed my request for additional test, and given existing tests, there's no need to add a test.

njharman (approved):

x110dc (approved):

x110dc (approved):

PR #1702, opened 2025-01-17
DEV-5227: Allow updating contribution interval

nrh-cklimas:

Creating review app for testing


apps/contributions/serializers.py
```diff
@@ -871,14 +872,15 @@ class Meta:
             "donor_selected_amount",
             "provider_payment_method_id",
         ]
-        read_only_fields = PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS
+        read_only_fields = [
+            _ for _ in PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS if _ not in ["amount", "interval"]
+        ]
 
     def update(self, instance: Contribution, validated_data) -> Contribution:
```
benjaminEwhite, line 879: This appears to involve potential db commits in two places: this function and in called function update_subscription_price.  MIght be worth using @transaction_atomic decorator on update method so those db commits succeed or fail together.

apps/contributions/models.py
```diff
@@ -1336,19 +1344,55 @@ def update_subscription_amount(self, amount: int, donor_selected_amount: float)
             )
         except StripeError:
             logger.exception(
-                "Encountered a Stripe error while trying to update payment method for subscription on contribution %s",
+                "Encountered a Stripe error while trying to update subscription price on contribution %s",
                 self.id,
             )
             raise
 
         with reversion.create_revision():
```
benjaminEwhite, line 1352: Part of me wonders if this should be happening in webhook handler. Since we're updating price on sub, that should trigger a webhook (I'm guessing customer.subscription_updated, but haven't confirmed that).

apps/contributions/models.py
```diff
@@ -1336,19 +1339,53 @@ def update_subscription_amount(self, amount: int, donor_selected_amount: float)
...
-        with reversion.create_revision():
-            self.save(update_fields={"contribution_metadata", "modified"})
-            reversion.set_comment(
-                f"`Contribution.update_subscription_amount` saved changes to contribution with ID {self.id}"
-            )
-
         # Send amount updated email to contributor
-        self.send_recurring_contribution_amount_updated_email()
+        if amount:
+            self.send_recurring_contribution_amount_updated_email()
```
benjaminEwhite, line 1349: Shouldn't this move into calling context given that this method is meant to only be about updating stripe?

apps/contributions/models.py
```diff
@@ -1336,19 +1344,55 @@ def update_subscription_amount(self, amount: int, donor_selected_amount: float)
...
 
         # Send amount updated email to contributor
-        self.send_recurring_contribution_amount_updated_email()
+        if amount:
+            self.send_recurring_contribution_amount_updated_email()
+
+        # TODO @nrh-cklimas: send interval updated email to contributor
+        # DEV-5691
+
+    def update_subscription_price_validate_args(
```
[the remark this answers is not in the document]
benjaminEwhite: I see this does validate if amount then donor selected amount, but probably should also handle if donor selected then amount.

apps/contributions/serializers.py
```diff
@@ -871,30 +874,53 @@ class Meta:
...
+                instance.update_payment_method_for_subscription(provider_payment_method_id=provider_payment_method_id)
+
+            # Need to pop donor_selected_amount from validated_data so it doesn't get sent when saving changes.
+            # We want to set the donor-selected amount in metadata if the schema supports it.
+            if donor_selected_amount := validated_data.pop("donor_selected_amount", None):
+                with contextlib.suppress(InvalidMetadataError):
+                    instance.set_metadata_field("donor_selected_amount", donor_selected_amount)
+                    update_fields.add("contribution_metadata")
+
+            # Now that we've popped donor_selected_amount, we can now safely add
```
benjaminEwhite, line 895:
```suggestion
            # Now that we've popped donor_selected_amount, we can safely add
```

apps/contributions/serializers.py
```diff
@@ -871,30 +874,53 @@ class Meta:
...
+            update_fields.update(validated_data.keys())
 
             if amount := validated_data.get("amount", None):
-                if not donor_selected_amount:
+                if donor_selected_amount is None:
                     # amount and donor_selected_amount should always be set in tandem in real life thanks to validate(),
                     # but we want to be certain.
                     raise serializers.ValidationError(
-                        "If amount is updated, donor_selected_amount must be set as well."
+                        "If amount is updated, donor_selected_amount must be updated as well."
```
benjaminEwhite, line 904: Would be better for this to live in method that's doing rest of validation, IMO.

apps/contributions/serializers.py
```diff
@@ -871,30 +874,53 @@ class Meta:
             "donor_selected_amount",
             "provider_payment_method_id",
         ]
-        read_only_fields = PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS
+        read_only_fields = [
+            _ for _ in PORTAL_CONTRIBUTION_DETAIL_SERIALIZER_DB_FIELDS if _ not in ["amount", "interval"]
+        ]
 
     def update(self, instance: Contribution, validated_data) -> Contribution:
```
benjaminEwhite, line 881: There's a lot going on in this update method, and a little hard to keep track of it all.


IMO, a quick win would be to group all the stripe update logic into a separate method that gets called here, and to do the same with all the revengine update code. 

Also, i may have missed it, but I didn't see amount or frequency getting updated in revengine contribution metadata or Stripe metadata. Those are not in 1.4 schema but they are in 1.6 and 1.7 in metadata spreadsheet, and IIUC we'd need to support insofar as those can appear in portal. 

Final comment: there's a number of  intermediate places something could go wrong and could leave data in a bad state. For instance, I wonder what happens if say, updating payment method succeeds but updating price fails. I guess the user gets an error and maybe retries, but I don't have a good sense of what the consequences of that would be. I guess one could imagine they quickly resubmit and that causes some sort of race condition possibly involving webhooks firing out of order.

benjaminEwhite: Oh one more thing: def should have a docstring summarizing what all this method does, cause it's a lot.

benjaminEwhite (changes requested):

apps/api/tests/test_views.py
```diff
@@ -533,6 +533,7 @@ def test_response_when_using_noncontributor_token(self):
 
 @pytest.mark.django_db
 @pytest.mark.usefixtures("default_feature_flags")
+@pytest.mark.usefixtures("mock_stripe")
```
benjaminEwhite, line 536: Noting that I pushed up the commit that made this change (and added this fixture in conftest.py). Unclear why, but a test in this class was failing in CI on this branch because we were somehow leaking an attempted call to real stripe. Suppressing stripe here causes the test to pass.

apps/contributions/models.py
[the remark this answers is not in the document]
nrh-cklimas: ba90db82e16d6f989dc106f1768e60ded2fbb9e7

apps/contributions/models.py
nrh-cklimas: I think we decided to let this go for now but let me know if I have that wrong.

apps/contributions/models.py
```diff
@@ -1257,75 +1257,83 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
+            # now and billing the contributor with a new invoice.
+            #
+            # Regardless of the scenario, we never want to prorate changes. We
+            # want to treat the subscription as an entirely new entity, and not
+            # apply previously-contributed money toward a changed amount.
+            #
+            # https://docs.stripe.com/api/subscriptions/update
             stripe.Subscription.modify(
-                sub_id,
+                self.provider_subscription_id,
```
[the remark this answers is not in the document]
nrh-cklimas: Handled in `update_subscription_price_validate_args`, I think

apps/contributions/models.py
```diff
@@ -1257,75 +1256,79 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
+        self.update_subscription_price_validate_args(amount, donor_selected_amount, interval)
+        logger.debug(
             "fetching subscription items from sub %s",
-            sub_id,
+            self.provider_subscription_id,
         )
         try:
-            items = stripe.SubscriptionItem.list(subscription=sub_id, stripe_account=self.stripe_account_id)
+            items = stripe.SubscriptionItem.list(
+                subscription=self.provider_subscription_id, stripe_account=self.stripe_account_id
```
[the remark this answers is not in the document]
nrh-cklimas: Think this is handled in `updated_subscription_price_validate_args`, line 1373

apps/contributions/models.py
```diff
@@ -1257,75 +1256,79 @@ def update_payment_method_for_subscription(self, provider_payment_method_id: str
...
-        # Set the donor-selected amount in metadata if the schema supports it.
-        with contextlib.suppress(InvalidMetadataError):
-            self.set_metadata_field("donor_selected_amount", donor_selected_amount)
-
-        self.amount = amount
-
         logger.info(
-            "Updating Stripe Subscription's %s (item %s), amount to %s",
-            sub_id,
+            "Updating Stripe Subscription's %s (item %s) amount to %s, interval to %s",
```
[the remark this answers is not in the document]
nrh-cklimas: 927845e382b5bc51d078e5a563fbd5a5ce9a1a41

apps/contributions/models.py
nrh-cklimas: Weird, I thought I had but obviously missed it. 4e323e3c663da675cd1e5b16164f4aedc69f0786

apps/contributions/serializers.py
nrh-cklimas: I'm OK with removing this but my memory is this came up in code review in the past, that we wanted coverage at this stage too. But that might be outdated with the re-org of the function.

PR #1705, opened 2025-01-22
Dev 5251 add authed endpoint s to create contributions by switchboard

benjaminEwhite:

#### What's this PR do?

This PR adds an authenticated endpoint for Switchboard with which it can create, retrieve (but not list), and update (via patch) contributions.

This PR also removes a deprecated endpoint at `api/v1/switchboard-contributions`. In an earlier PR that endpoint was aliased to point to `api/v1/switchboard/contributions`.  There was a TO-DO to remove the deprecated endpoint once Switchboard was updated to use the new one, but it turns out that SB never got around to using the deprecated one, so it's retired in this PR.

Also, noting that I discovered we had an old, no longer used, dead view function at `apps.contributions.views.checkout.payment_success`.  That view function was not attached to any URL, and has no caller, so I removed that in this PR.

Finally, I discovered an inaccurate revision comment for our handle_invoice_payment_succeeded webhook handler (copy pasta from handle_payment_intent_succeeded), which I updated in https://github.com/newsrevenuehub/rev-engine/pull/1705/commits/0a2276c4389edbe592c2594e5230e3f4144b8661

This moves us forward on enabling Switchboard to create entities in revengine based on events driven by other systems.

Yes.

No

Yes

N/A

No

N/A

[DEV-5251](https://news-revenue-hub.atlassian.net/browse/DEV-5251)

No

No

[DEV-5251]: https://news-revenue-hub.atlassian.net/browse/DEV-5251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/models.py
```diff
@@ -419,6 +419,10 @@ def revenue_program(self) -> RevenueProgram | None:
             return self.donation_page.revenue_program
         return self._revenue_program
 
+    @revenue_program.setter
+    def revenue_program(self, value: RevenueProgram):
```
benjaminEwhite, line 423: This is added because it makes life much easier on `SwitchboardContributionDetailSerializer`.

apps/contributions/tests/views/test_switchboard_views.py
benjaminEwhite, line 1: We ended up with redundant test file for this set of views. Deleted this one and retained `apps/contributions/tests/views/test_switchboard.py`

apps/contributions/urls.py
```diff
@@ -3,20 +3,13 @@
...
 
-from apps.contributions.views import checkout, orgs, portal, switchboard, webhooks
+from apps.contributions.views import checkout, orgs, portal, webhooks
 
 
 router = routers.DefaultRouter()
 router.register(r"contributions", orgs.ContributionsViewSet, basename="contribution")
 router.register(r"payments", checkout.PaymentViewset, basename="payment")
 router.register(r"contributors", portal.PortalContributorsViewSet, basename="portal-contributor")
-# TODO @BW: Remove deprecated switchboard-contributions endpoint after SB updated to /v1/switchboard/contributions/
```
benjaminEwhite, line 13: As mentioned in PR description, this is opportunistically retired in this PR because it is not a requirement for Switchboard.

apps/contributions/views/switchboard.py
```diff
@@ -3,34 +3,72 @@
...
+    viewsets.mixins.CreateModelMixin,
+    viewsets.mixins.UpdateModelMixin,
+    viewsets.mixins.RetrieveModelMixin,
+    viewsets.GenericViewSet,
+):
     """Viewset for switchboard to update contributions."""
 
     permission_classes = [IsSwitchboardAccount]
-    http_method_names = ["patch"]
+    http_method_names = ["patch", "post", "get"]
```
benjaminEwhite, line 40: Do this because we don't want to allow put.

apps/contributions/views/switchboard.py
```diff
@@ -3,34 +3,77 @@
...
 from apps.api.authentication import JWTHttpOnlyCookieAuthentication
 from apps.api.permissions import IsSwitchboardAccount
 from apps.contributions import serializers
 from apps.contributions.models import Contribution, Contributor
 
 
+# This is a fragment of the error message that is raised when contribution constraint that requries only one of _revenue_program
+# or donation_page to be set is violated. We have a custom exception handler that needs to catch this error. The full constraint
+# name gets truncated in the error message so we use this fragment to identify the error.
+EXCLUSIVE_RP_OR_PAGE_CONSTRAINT_LABEL_FRAGMENT = "exclusive_donation_page_or__revenue"
```
njharman, line 25: It's kind of lame Django doesn't expose (or at least the docs don't explain it does) what the db / psycopg provides https://www.psycopg.org/docs/module.html#psycopg2.Error

Django is very lowest-common-denominator with regards to backends.

apps/contributions/views/switchboard.py
```diff
@@ -3,34 +3,77 @@
...
+            unique_error_found = False
+            for errors in details.values():
+                for error in errors:
+                    if error.code == "unique":
+                        unique_error_found = True
+                        break
+                if unique_error_found:
+                    break
+            if unique_error_found:
+                exc.status_code = status.HTTP_409_CONFLICT
```
njharman, line 74: Eh?

```
if any(x.code == "unique" for x in errors):
  exc.status_code = status.HTTP_409_CONFLICT
  break
```

Although, either solution will hit on any unique violation. Just happens that the three fields mentioned in doc string are the only current unique constraints.

njharman (changes requested):

apps/contributions/models.py
benjaminEwhite: Since this doesn't actually save to the db, I don't think this should raise. In calling context, one might set  attributes one at a time before saving:

```
instance.revenue_program =  some_rp
instance.donation_page = None
instance.save()
```

I don't think we want to change an DP's RP when trying to set the RP on a contribution, because that affects not just the contribution, but the entire RP.

If new a caller uses this setter and then saves and raises an integrity error, I don't think that's the end of the world. Worst case, we find out about in Sentry, and can handle the error in calling context.

If nothing else, a code comment is helpful here, so I will add.

apps/contributions/serializers.py
```diff
@@ -920,44 +922,78 @@ class Meta:
         read_only_fields = PORTAL_CONTRIBUTION_BASE_SERIALIZER_FIELDS
 
 
+class SwitchboardContributionRevenueProgramSourceValues:
+    VIA_PAGE: Final = "via_page"
+    DIRECT: Final = "direct"
+
+    ValidValues = Literal["via_page", "direct"]
```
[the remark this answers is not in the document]
benjaminEwhite: Unfortunately Pylance (which vscode Python extension uses) complains about using var as arg to Literal.

Screenshot 2025-01-23 at 7 16 33 AM

I gather Ruff doesn't care.

apps/contributions/views/switchboard.py
benjaminEwhite: I had a pass where I was inspecting the pg error diag, and even there the full constraint name was being truncated.

apps/contributions/serializers.py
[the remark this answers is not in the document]
benjaminEwhite: This ends up moot since I was able to inherit from str + enum 

fd83ad1a6

apps/contributions/serializers.py
```diff
@@ -920,44 +922,78 @@ class Meta:
         read_only_fields = PORTAL_CONTRIBUTION_BASE_SERIALIZER_FIELDS
 
 
+class SwitchboardContributionRevenueProgramSourceValues:
```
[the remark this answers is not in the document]
benjaminEwhite: We're on 3.10x in rev engine, and StrEnum is introduced in 3.11.

That said, I was able to subclass str, Enum to get same effect,

fd83ad1a6

apps/contributions/views/switchboard.py
benjaminEwhite: 459b93f2e

apps/contributions/views/switchboard.py
njharman: well that's lame as well.  to be clear, that something is truncating (maybe there is a good reason). What you've done is reasonable and practical.

apps/contributions/serializers.py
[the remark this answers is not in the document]
njharman: > We're on 3.10x in rev engine, and StrEnum is introduced in 3.11.

Oooh.

Nice solution!

apps/contributions/serializers.py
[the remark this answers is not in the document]
njharman: > complains about using var as arg to Literal

I can see the logic behind that. 

moot issues are the best kind of issue!

apps/contributions/models.py
njharman: Any concern njharman had is resolved. Leaving conversation "open" to preserve developer, Ben's, PR note.

njharman (approved):

apps/contributions/views/switchboard.py
```diff
@@ -3,34 +3,71 @@
...
+
+        We also look for violation of our exclusive constraint that requires only one of _revenue_program or donation_page to be set.
+        We need a custom handler here to ensure we get a 400 status code for this error, along with an appropriate message.
+        """
+        if isinstance(exc, IntegrityError) and EXCLUSIVE_RP_OR_PAGE_CONSTRAINT_LABEL_FRAGMENT in (
+            exc.args[0] if exc.args else ""
+        ):
+            return Response(
+                {"detail": EXCLUSIVE_RP_OR_PAGE_CONSTRAINT_ERROR_MESSAGE}, status=status.HTTP_400_BAD_REQUEST
+            )
```
x110dc, line 62: This feels brittle. What if a subsequent version of psycopg2 or Postgres changes this error message? The requirement is that the endpoint shouldn't allow both of these to be set. Why can't that be achieved as a validation on the serializer? This part feels unnecessary. If the validation somehow let something through the IntegrityError could bubble up.

x110dc (changes requested):

apps/contributions/urls.py
benjaminEwhite: Yes, the new endpoint supports this. Based on this comment I've explicitly added this scenario to the Jupyter notebook for QA. In the cell that proves that update functionality works, I have us setting donation_page to None and setting revenue_program (mapping to _revenue_program) to a different RP.

Note that as in previous endpoint the following restrictions continue here:

- page OR _revenue_program can and must be set on a contribution, but both may not be
- the rp of a contribution cannot be set to one that is for an org that is different than the one with which it was originally created. If we decide that's too restrictive, we can relax this requirement in a follow up ticket.

apps/contributions/views/checkout.py
```diff
@@ -90,23 +89,3 @@ def destroy(self, request, *args, **kwargs):
             return Response({"detail": "Something went wrong"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
         return Response(status=status.HTTP_204_NO_CONTENT)
-
-
-@api_view(["PATCH"])
-@permission_classes([])
-def payment_success(request, uuid=None):
```
benjaminEwhite, line 97: See PR description for more detail, but this is removed because it was a "dead" view, not connected to any URL.

apps/contributions/serializers.py
```diff
@@ -920,44 +923,76 @@ class Meta:
...
-                instance.donation_page = None
-                update_fields.add("donation_page")
-            with reversion.create_revision():
-                instance.save(update_fields=update_fields)
-                reversion.set_comment("Updated by SwitchboardContributionSerializer.update")
-        return instance
+    def validate_contribution_metadata(self, value: dict) -> dict:
+        """Ensure that the contribution metadata is a valid JSON object."""
+        try:
+            StripeTransactionsImporter.validate_metadata(value)
```
x110dc, line 982: Feels a little weird that we're using the validation function from a class that's unrelated to this. Is this a case for pulling metadata validation out into a class/function that's used here and the importer? If you agree, could just be a todo/ticket.

apps/contributions/views/switchboard.py
x110dc: I'm okay with the uncaught errors here. Again, why not just add validation in the serializer to check for this condition? Like 

```
def validate(self, data):
        if data.get('donation_page') and data.get('_revenue_program'):
                "Cannot set both donation_page and revenue_program"
            )
        return data
```

If we're getting 500s we can't explain I'd rather those surface in Sentry so we can figure out why things aren't working as expected. As this is entirely new there's little/no risk here.

x110dc (changes requested):

apps/contributions/views/switchboard.py
benjaminEwhite: Pivoted to an overridden validate method in this commit 551b5f61946 (which required a bit more nuance to handle all edge cases).

apps/contributions/serializers.py
benjaminEwhite: ~equal work to just make the change vs. creating a ticket: a4ffacaf5

apps/contributions/views/switchboard.py
```diff
@@ -20,17 +21,37 @@
...
-    http_method_names = ["patch"]
+    http_method_names = ["patch", "post", "get"]
     queryset = Contribution.objects.all()
     serializer_class = serializers.SwitchboardContributionSerializer
     # TODO @BW: Remove JWTHttpOnlyCookieAuthentication after DEV-5549
     # DEV-5571
     authentication_classes = [TokenAuthentication, JWTHttpOnlyCookieAuthentication]
 
+    def handle_exception(self, exc):
+        """Ensure select db-level errors get a 400 or 409.
```
x110dc, line 41: I see the 409, but not the 400. Update comment?

x110dc (changes requested):
I think there's a tiny error in a comment. Otherwise LGTM.

apps/contributions/views/switchboard.py
benjaminEwhite: fixed with 560c9c4832 .  As Daniel mentioned in JIRA, this comment is non-blocking, so I'm going to resolve.

PR #1706, opened 2025-01-23
DEV-5272: Add frequency change to contributor portal SPA

nrh-cklimas:

n/a

#### What's this PR do?

Adds UI for changing the frequency of contributions.

Allows users to change their contribution frequency.

Yes.

Yes.

Yes.

Notify clients that this is possible now.

No.

No.

- https://news-revenue-hub.atlassian.net/browse/DEV-5272
- https://news-revenue-hub.atlassian.net/browse/DEV-5227

No.

No.


spa/src/components/portal/ContributionsList/ContributionDetail/BillingDetails/BillingDetails.tsx
```diff
@@ -38,50 +50,91 @@ export const INTERVAL_NAMES: Record<ContributionInterval, string> = {
...
+
+      if (editedInterval === contribution.interval) {
+        onUpdateAmount({ amount: amountInCents, donor_selected_amount: parsedAmount });
+      } else {
+        onUpdateAmount({ amount: amountInCents, donor_selected_amount: parsedAmount, interval: editedInterval });
+      }
+    } else {
+      // We are changing interval only.
+      onUpdateAmount({ interval: editedInterval });
+    }
```
benjaminEwhite, line 112: Minor, but why not init an object called `updateData` and then depending on conditions, add entries, and then have single call to `onUpdateAmount`. Probably just a stylistic thing (and how I'd approach this if code was in Python, which is hardly relevant here). That said, IMO it would be slightly more readable.

spa/vite.config.js
```diff
@@ -35,6 +35,7 @@ const config = defineConfig({
   ],
   server: {
     // Accept connections on any hostname.
+    allowedHosts: true,
```
benjaminEwhite, line 38: This change seems tangential, but guessing it was needed for some reason, yeah?

spa/src/components/portal/ContributionsList/ContributionDetail/BillingDetails/BillingDetails.tsx
nrh-cklimas: TypeScript dislikes this because

- if you don't specify the type of `updateData`, it infers the type as an empty object and then doesn't like it when you add properties to it
- if you declare it as an `AmountChange`, it doesn't like that the object is initially empty

There's might be a TS declaration I don't know about that would facilitate this, though. This is the code I was experimenting with:

```ts
let updateData: AmountChange = {};

if (parsedAmount !== originalAmountInDollars) {
  // We are changing at least amount, but maybe also interval as well.

  // Need "Math.round" because "parseFloatStrictly" * 100 can result in numbers like:
  // amount = 9.45 -> formattedAmount * 100 = 944.9999999999999
  // amount = 9.46 -> formattedAmount * 100 = 946.0000000000001
  const amountInCents = Math.round(parsedAmount * 100);

  if (editedInterval === contribution.interval) {
    updateData = { amount: amountInCents, donor_selected_amount: parsedAmount };
  } else {
    updateData = { amount: amountInCents, donor_selected_amount: parsedAmount, interval: editedInterval };
  }
} else {
  // We are changing interval only.
  updateData = { interval: editedInterval };
}

onUpdateAmount(updateData);
```

PR #1712, opened 2025-01-30
DEV-5550: Remove bespoke Recaptcha code, add additional state around form submission

nrh-cklimas:

n/a

#### What's this PR do?

1. Adds additional state to block a second contribution page form submission while ReCAPTCHA and other processing is taking place.
2. Removes DIY ReCAPTCHA code with a library.

1. Prevents double submission that causes errors.
2. Avoids re-inventing the wheel.

Yes.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5550

No.

No.


spa/src/appSettings.ts
```diff
@@ -35,7 +35,6 @@ export const CSRF_HEADER = 'X-CSRFTOKEN';
 export const PASSWORD_RESET_URL = '/users/password-reset/';
 
 // Google reCAPTCHA
-export const GRECAPTCHA_SCRIPT_URL = 'https://www.google.com/recaptcha/api.js';
```
nrh-cklimas, line 38: No longer needed now that we're using a library to load ReCAPTCHA.

spa/src/components/donationPage/DonationPage.jsx
```diff
@@ -54,48 +53,29 @@ function DonationPage({ page, live = false }, ref) {
...
+  // We do additional work before creating the submission (e.g. POSTing to
+  // create the payment), like verifying reCAPTCHA, that should block form
+  // submission in addition to the payment loading state below. This state
+  // captures that.
+  const [submitting, setSubmitting] = useState(false);
   const { createPayment, deletePaymentMutation, isLoading: paymentIsLoading, payment } = usePayment();
   const { auditAmountChange, auditFrequencyChange, auditPayFeesChange, auditPaymentCreation } = useAmountAuditing();
+  const { executeRecaptcha } = useGoogleReCaptcha();
+
+  useClearbit(live);
```
benjaminEwhite, line 65: What is the significance of this moving earlier in this component definition? Does it effect timing? If so, a code comment would be valuable.

spa/src/components/donationPage/PublishedDonationPage.test.tsx
```diff
@@ -151,6 +168,15 @@ describe('PublishedDonationPage', () => {
       ]);
     });
 
+    it('sets up Google ReCAPTCHA for the page with key and CSP nonce', () => {
```
benjaminEwhite, line 171: I gather that CSP relates to recaptcha, but that linkage is unclear to me. 

Also, I'm curious are the recaptcha changes in this PR opportunistic or are they intrinsic to preventing double-submit? All else equal, touching current recaptcha implementation seems to expand scope and testing surface area considerably, and at least in theory could have down stream effects on fraud prevention (in particular bad actor which looks to recaptcha value).

spa/src/components/donationPage/DonationPage.jsx
nrh-cklimas: No effect--I did it to make the code slightly prettier/group things together more, but it's definitely a personal pref of mine.

spa/src/components/donationPage/PublishedDonationPage.test.tsx
nrh-cklimas: The CSP nonce is needed because the recaptcha library is added via a `<script>` tag. I can add a comment to that effect.

We can peel out the Recaptcha refactor if you think it's a good idea. My thought process while debugging the root issue was:

1. We have zero unit test coverage on the Recaptcha-related code. That's not good.
2. (tap head meme) You don't need to write unit tests on code you don't own.

In defense of my laziness, we do assert in Cypress that there's a Recaptcha token in the request body when creating the contribution. I feel reasonably comfortable that it's something we could verify in QA, and could even test a Recaptcha load failure by editing the API key in the review app.

benjaminEwhite: That all makes sense. I think I can get on board with wrapping this change into this PR given that, as you mention, we don't have testing on the existing code, and (presumably) the third party code is tested, and assuming that QA instructions are updated to include verifying Recaptcha functionality as per your description.

benjaminEwhite (approved):
Approved (with understanding that verification of Recaptcha functionality will be added to QA instructions).

spa/src/appSettings.ts
```diff
@@ -35,8 +35,7 @@ export const CSRF_HEADER = 'X-CSRFTOKEN';
...
-export const GRECAPTCHA_SITE_KEY = '6Lfuse8UAAAAAD9E6tCxKYrxO1IbnXp8IBa4u5Ri';
+export const GRECAPTCHA_SITE_KEY = 'REMOVE_ME_6Lfuse8UAAAAAD9E6tCxKYrxO1IbnXp8IBa4u5Ri';
```
nrh-cklimas, line 38: Note to drop this commit before merging.

nrh-cklimas: Done now

PR #1717, opened 2025-02-06
DEV-5669: Add read-only Switchboard endpoint for users

nrh-cklimas:

n/a

#### What's this PR do?

1. Adds Switchboard-only route for retrieving a user by ID or email.
2. Re-orgs views in the users app to have separate revengine and switchboard modules.

1. Progresses Switchboard/rev-engine integration.
2. Makes it clear which views are SPA-facing and which are for Switchboard.

TBD

No.

Yes.

n/a

No.

Endpoints appear in OpenAPI.

https://news-revenue-hub.atlassian.net/browse/DEV-5669

No.

No.


apps/users/views/revengine.py
nrh-cklimas, line 1: As the diff indicates, nothing changed here.

apps/users/serializers.py
```diff
@@ -145,6 +145,24 @@ class Meta:
         ]
 
 
+class SwitchboardUserSerializer(serializers.ModelSerializer):
+    """Read-only access to users. Organizations and revenue programs are serialized by ID only."""
+
+    organizations = serializers.SerializerMethodField("get_orgs", default=[])
+    revenue_programs = serializers.SerializerMethodField("get_rps", default=[])
```
benjaminEwhite, line 152: I believe we can lean on `serializers.PrimaryKeyRelatedField` and then get rid of the `get_orgs` and `get_rps` methods below:

```
    organizations = serializers.PrimaryKeyRelatedField(many=True, read_only=True, source="permitted_organizations")
    revenue_programs = serializers.PrimaryKeyRelatedField(
        many=True, read_only=True, source="permitted_revenue_programs"
    )
```

apps/users/serializers.py
```diff
@@ -145,6 +145,24 @@ class Meta:
...
+
+    class Meta:
+        model = get_user_model()
+        fields = ["email", "first_name", "id", "job_title", "last_name", "organizations", "revenue_programs"]
+        read_only_fields = ["email", "first_name", "id", "job_title", "last_name", "organizations", "revenue_programs"]
```
benjaminEwhite, line 157: AC9 calls for `role_type` being in response (so on this serializer) looks like that is missing.

Believe you will need to lean on `serializers.SerializerMethodField` for that because not all users have a role assignment, so will need to derive from elsewhere when that's the case.

apps/users/views/switchboard.py
```diff
@@ -0,0 +1,25 @@
...
+from rest_framework.viewsets import GenericViewSet
+
+from apps.api.permissions import IsSwitchboardAccount
+from apps.users import serializers
+from apps.users.models import User
+
+
+class SwitchboardUsersViewSet(RetrieveModelMixin, GenericViewSet):
+    permission_classes = [IsSwitchboardAccount]
+    http_method_names = ["get"]
```
benjaminEwhite, line 16: I don't think you need this here, that should be the only method allowed here by default. The reason you see this being manually set in some views is because we want to exclude PUT when we are using UpdateModelMixin, because out of the box it supports both PUT and PATCH for update.

apps/users/views/switchboard.py
```diff
@@ -0,0 +1,25 @@
...
+    queryset = User.objects.all()
+    serializer_class = serializers.SwitchboardUserSerializer
+    authentication_classes = [TokenAuthentication]
+
+    @action(methods=["get"], url_path="email/(?P<email>[^/]+)", detail=False)
+    def get_by_email(self, request, email=None):
+        user = get_object_or_404(User.objects.all(), email=email)
```
benjaminEwhite, line 23: Just want to call out that `User.email` is currently case sensitive. It's not the responsibility of this PR to solve for making it case insensitive, but we should probably create followup tickets for this.

benjaminEwhite: Also, if you wanted to create a TODO for `SwitchboardContributorsViewSet` to adopt the pattern you're using here instead of separate function based view for retrieve by email, I'd sure would think that's swell.

benjaminEwhite (changes requested):

apps/users/serializers.py
nrh-cklimas: Your comment made me realize I had whiffed on the expected format of the response--I'll link the comment in Jira. I updated in 0f62f005880b269ed3fafa89edc7912c6ef22bd8 but I don't understand, if someone doesn't have a role assignment, how else do we derive role?

apps/users/serializers.py
nrh-cklimas: See https://github.com/newsrevenuehub/rev-engine/pull/1717#discussion_r1946680422

apps/users/views/switchboard.py
nrh-cklimas: 4ae18dccd

apps/users/views/switchboard.py
nrh-cklimas: https://news-revenue-hub.atlassian.net/browse/DEV-5786

apps/users/serializers.py
```diff
@@ -145,6 +145,24 @@ class Meta:
...
+    role = serializers.SerializerMethodField("get_role")
+
+    class Meta:
+        model = get_user_model()
+        fields = ["email", "first_name", "id", "job_title", "last_name", "role"]
+        read_only_fields = ["email", "first_name", "id", "job_title", "last_name", "role"]
+
+    def get_role(self, obj: User):
+        return {
+            "type": obj.role_type[0] if isinstance(obj.role_type, tuple) and len(obj.role_type) == 2 else None,
```
benjaminEwhite, line 160: I think it might be a good idea to create a distinct `User.get_role` method. It can be narrowly unit tested. 

Not sure if this makes sense or not, but we might also want to be able to indicate that a user is a superuser as a role type. That said, I don't think anything logically prevents a user from having a role assignment AND being a superuser.  

Not sure what right decision is here for other role types, but I'm guessing it depends on what expected use case is for this endpoint (which I don't know much about off hand, and am not going to dig around in JIRA to find out at the moment).

apps/users/tests/test_serializers.py
```diff
@@ -408,3 +408,39 @@ def test_has_expected_readable_and_writable_fields(self):
...
+            ("org_user_multiple_rps", "org_admin"),
+            ("rp_user", "rp_admin"),
+            ("user_with_verified_email_and_tos_accepted", None),
+        ]
+    )
+    def user_and_expected_role(self, request):
+        return (request.getfixturevalue(request.param[0]), request.param[1])
+
+    def test_has_expected_fields(self, user_and_expected_role: tuple[User, str | None]):
+        user, expected_role = user_and_expected_role
```
benjaminEwhite, line 429: Slightly cleaner way to do this (that I wasn't aware of a couple years back when i first started using pytest fixtures, so there's plenty of bad examples in code base):

```suggestion
@pytest.mark.parametrize(
        ("user_fixture", "expected_role"),
        [
            ("superuser", "superuser"),
            ("hub_admin_user", "hub_admin"),
            ("org_user_free_plan", "org_admin"),
            ("org_user_multiple_rps", "org_admin"),
            ("rp_user", "rp_admin"),
            ("user_with_verified_email_and_tos_accepted", None),
        ],
    )
    def test_has_expected_fields(self, user_fixture: str, expected_role: str, request):
        user = request.getfixturevalue(user_fixture)
        assert serializers.SwitchboardUserSerializer(user).data == {
            "email": user.email,
            "first_name": user.first_name,
            "id": user.id,
            "job_title": user.job_title,
            "last_name": user.last_name,
            "role": {
                "type": expected_role,
                "organizations": [org.id for org in user.permitted_organizations],
                "revenue_programs": [org.id for org in user.permitted_revenue_programs],
            },
        }
```

apps/users/tests/test_serializers.py
```diff
@@ -408,3 +408,39 @@ def test_has_expected_readable_and_writable_fields(self):
         actual_read_only_fields = set(serializers.MutableUserSerializer.Meta.read_only_fields)
         assert actual_writable_fields == expected_writable_fields
         assert actual_read_only_fields == expected_read_only_fields
+
+
+@pytest.mark.django_db
+class TestSwitchboardUserSerializer:
```
benjaminEwhite, line 414: It would be worthwhile to unit-test .get_role method because it has slightly complex logic in it.

benjaminEwhite (changes requested):

apps/users/tests/views/test_switchboard.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @pytest.fixture
+    def expired_token(self, api_client, switchboard_user, default_password, monkeypatch, settings):
+        token, token_string = AuthToken.objects.create(switchboard_user)
+        token.expiry = now() - timedelta(days=1)
+        token.save()
+        return token_string
+
+    @pytest.mark.parametrize("exists", [True, False])
+    def test_retrieve_by_id(self, api_client, token, org_user_multiple_rps: User, exists):
```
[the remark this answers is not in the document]
nrh-cklimas: We talked this through and we're doing more thorough testing on the serializer.

apps/users/serializers.py
nrh-cklimas: When we talked this through, we decided adding a User model method would be good for a followup.

apps/users/serializers.py
```diff
@@ -145,6 +145,24 @@ class Meta:
...
+    """Read-only access to users. Organizations and revenue programs are serialized by ID only."""
+
+    role = serializers.SerializerMethodField("get_role")
+
+    class Meta:
+        model = get_user_model()
+        fields = ["email", "first_name", "id", "job_title", "last_name", "role"]
+        read_only_fields = ["email", "first_name", "id", "job_title", "last_name", "role"]
+
+    def get_role(self, obj: User):
```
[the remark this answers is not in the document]
nrh-cklimas: This didn't work:
```python
        data = serializers.SwitchboardUserSerializer(user).data
        assert isinstance(data["role"], serializers.UserRole)
```
It said: `TypeError: TypedDict does not support instance and class checks`

But I think the typing makes sense here, and I added it in f158c2d2a

apps/users/tests/views/test_switchboard.py
```diff
@@ -0,0 +1,113 @@
...
+from rest_framework.reverse import reverse
+
+from apps.users.models import User
+
+
+@pytest.mark.django_db
+class TestSwitchboardUsersViews:
+    @pytest.fixture
+    def token(self, switchboard_user):
+        return AuthToken.objects.create(switchboard_user)[1]
```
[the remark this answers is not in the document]
nrh-cklimas: 8ed5ac167

apps/users/tests/views/test_switchboard.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @pytest.fixture
+    def expired_token(self, api_client, switchboard_user, default_password, monkeypatch, settings):
```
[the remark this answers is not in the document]
nrh-cklimas: 8ed5ac167

apps/users/tests/test_serializers.py
nrh-cklimas: I added some tests in 81393b577 but I'm not sure if the approach is 100% right.

apps/users/tests/views/test_switchboard.py
```diff
@@ -0,0 +1,81 @@
+import pytest
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from apps.users.models import User
+
+
+@pytest.mark.django_db
+class TestSwitchboardUsersViews:
```
benjaminEwhite, line 9: Minor (and no need for re-review from me to resolve), but can we call this `TestSwitchboardUsersViewSet` to follow naming convention we're using in other places?

benjaminEwhite (approved):

apps/users/tests/views/test_switchboard.py
nrh-cklimas: 4a1adb3fff631bb5f46d3eb76f20b1c1d35375b8

apps/users/views/switchboard.py
```diff
@@ -0,0 +1,24 @@
...
+
+class SwitchboardUsersViewSet(RetrieveModelMixin, GenericViewSet):
+    permission_classes = [IsSwitchboardAccount]
+    queryset = User.objects.all()
+    serializer_class = serializers.SwitchboardUserSerializer
+    authentication_classes = [TokenAuthentication]
+
+    @action(methods=["get"], url_path="email/(?P<email>[^/]+)", detail=False)
+    def get_by_email(self, request, email=None):
+        user = get_object_or_404(User.objects.all(), email=email)
```
njharman, line 22: Out of Scope. New ticket (First, probably a discussion with other devs. I had thought we already case insensitive email...)?  email.lower() and case insensitive email.

It's very annoying that Django.EmailField isn't already case insensitive, afaict.

apps/users/views/switchboard.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class SwitchboardUsersViewSet(RetrieveModelMixin, GenericViewSet):
+    permission_classes = [IsSwitchboardAccount]
+    queryset = User.objects.all()
+    serializer_class = serializers.SwitchboardUserSerializer
+    authentication_classes = [TokenAuthentication]
+
+    @action(methods=["get"], url_path="email/(?P<email>[^/]+)", detail=False)
+    def get_by_email(self, request, email=None):
```
njharman, line 21: email parameter does not need a default value.  Right? It's always gonna get something from url_path action routing.

apps/users/views/switchboard.py
```diff
@@ -0,0 +1,24 @@
...
+from rest_framework.viewsets import GenericViewSet
+
+from apps.api.permissions import IsSwitchboardAccount
+from apps.users import serializers
+from apps.users.models import User
+
+
+class SwitchboardUsersViewSet(RetrieveModelMixin, GenericViewSet):
+    permission_classes = [IsSwitchboardAccount]
+    queryset = User.objects.all()
```
njharman, line 16: Should is_active=False users be excluded? Line 34 apps/users/models.py

njharman (changes requested):

apps/users/serializers.py
```diff
@@ -145,6 +145,33 @@ class Meta:
...
+
+class SwitchboardUserSerializer(serializers.ModelSerializer):
+    """Read-only access to users. Organizations and revenue programs are serialized by ID only."""
+
+    role = serializers.SerializerMethodField("get_role")
+
+    class Meta:
+        model = get_user_model()
+        fields = ["email", "first_name", "id", "job_title", "last_name", "role"]
+        read_only_fields = ["email", "first_name", "id", "job_title", "last_name", "role"]
```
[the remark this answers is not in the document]
nrh-cklimas: Seems to work. 58686b766ed224280b32f553d3c77b80035504aa

apps/users/views/switchboard.py
nrh-cklimas: Not sure--this wasn't part of the AC on the ticket. How about we leave this comment open and @x110dc can weigh in on this when he reviews, since he was the originator of the ticket?

apps/users/views/switchboard.py
nrh-cklimas: That seems like a safe assumption to me. 17c3f79c5ba796d49a1f7101e74f06ee9b3e9105

apps/users/views/switchboard.py
nrh-cklimas: Ben noticed something similar and we created DEV-5787 to look into it.

apps/users/serializers.py
```diff
@@ -145,6 +145,33 @@ class Meta:
         ]
 
 
+RoleTypeCode = Literal["hub_admin", "org_admin", "rp_admin", "superuser"]
```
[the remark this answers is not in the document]
nrh-cklimas: This makes sense but I hit trouble trying to do anything with code. I tried this which seemed really basic:

```python
RoleTypeCode = Literal[
    RoleTypeReturnValues.HUB_ADMIN[0],
    RoleTypeReturnValues.ORG_ADMIN[0],
    RoleTypeReturnValues.RP_ADMIN[0],
    RoleTypeReturnValues.SUPERUSER[0],
]

class UserRole(TypedDict):
    type: RoleTypeCode | None
    organizations: list[int]
    revenue_programs: list[int]
```

Even this triggered `Variable not allowed in type expression` on the `type` line in `UserRole`. (Before this, I tried a list comprehension to extract out the role names but that didn't seem to work either.) I'm not very experience with type declarations in Python so I may be going about this in the wrong way. Any suggestions on how to approach it? I can fall back to a comment.

apps/users/views/switchboard.py
njharman: If so; Should create a new ticket to look at all instances of user.objects.all() and create a custom UserQuery/Manager [the correct django term is escaping me, we've created these before]. So we can User.ActiveUsers instead of having to (put is_active=False) all over.

apps/users/serializers.py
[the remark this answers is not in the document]
njharman: I don't use Types. I believe @benjaminEwhite  is our local Type expert.

njharman (approved):

apps/users/serializers.py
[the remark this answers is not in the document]
nrh-cklimas: See discussion on Jira--we think it's not feasible right now on this ticket. Added todo in 8cdcc867f2ef8c1c7292c476579cb3f089a220c2

apps/users/views/switchboard.py
x110dc: I think it's fine include all users on this endpoint. Though maybe we'll want a field indicating `is_active` but that can be a followup.

x110dc (dismissed):

apps/users/views/switchboard.py
nrh-cklimas: Created DEV-5875

PR #1722, opened 2025-02-18
DEV-5746: Add send_receipt query param to Switchboard contribution creation

nrh-cklimas:

n/a

#### What's this PR do?

The Switchboard-facing POST endpoint for contribution creation now looks for a query param named `send_receipt`. If present, it triggers a receipt email after creating the contribution.

Allows Switchboard to control when receipt emails are sent.

Yes.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5746

No.

No.


apps/contributions/views/switchboard.py
```diff
@@ -37,6 +37,24 @@ class SwitchboardContributionsViewSet(
     # DEV-5571
     authentication_classes = [TokenAuthentication, JWTHttpOnlyCookieAuthentication]
 
+    def perform_create(self, serializer: serializers.SwitchboardContributionSerializer):
+        """Send a receipt email if requested in a query param.
+
+        The default is to not send it.
+
+        Because we are only creating database models with this serializer, not
+        Stripe objects, duplicate receipt emails shouldn't occur. We assume
```
benjaminEwhite, line 46: Think there could be more context here about why duplicates are even a concern and why we wouldn't get duplpicates (namely, we assume metadata version if not v1.4 here, and that our revengine webhook handlers only send receipt email if dealing with v1.4.

Which also makes me wonder if maybe this should be strictter and only send if a.) param is true, AND b.) metadata schema is not v1.4 (or _is_ 1.6).

apps/contributions/views/switchboard.py
```diff
@@ -37,6 +37,24 @@ class SwitchboardContributionsViewSet(
...
+        callers have already created an appropriate Stripe object (e.g. payment
+        intent or subscription).
+        """
+        send_receipt = self.request.query_params.get("send_receipt")
+        contribution: Contribution = serializer.save()
+        if send_receipt is not None:
+            # send_thank_you_email() handles conditionality around whether
+            # receipt emails for the revenue program are sent by rev-engine.
+            logger.debug("Sending receipt email as requested by query param")
```
benjaminEwhite, line 55: Think this could be a .info so at least we can see it in google cloud

apps/contributions/views/switchboard.py
```diff
@@ -37,6 +37,24 @@ class SwitchboardContributionsViewSet(
...
+        The default is to not send it.
+
+        Because we are only creating database models with this serializer, not
+        Stripe objects, duplicate receipt emails shouldn't occur. We assume
+        callers have already created an appropriate Stripe object (e.g. payment
+        intent or subscription).
+        """
+        send_receipt = self.request.query_params.get("send_receipt")
+        contribution: Contribution = serializer.save()
+        if send_receipt is not None:
```
benjaminEwhite, line 52: Minor, but no need to assign to var at 50, can just check inline here.

apps/contributions/views/switchboard.py
benjaminEwhite: Also, might be good to include contributor ID/email/contribution id and maybe rp slug/id as well to make debugging via logs easier.

apps/contributions/tests/views/test_switchboard.py
```diff
@@ -310,6 +310,26 @@ def test_create_when_invalid_metadata(self, api_client, token, creation_data_rec
...
+            headers={"Authorization": f"Token {token}"},
+        )
+        mock_handle_thank_you_email.assert_not_called()
+
+    def test_create_sends_email_if_requested(
+        self, api_client, creation_data_recurring_with_page, request, token, mocker
+    ):
+        mock_handle_thank_you_email = mocker.patch("apps.contributions.models.Contribution.handle_thank_you_email")
+        api_client.post(
+            reverse("switchboard-contribution-list") + "?send_receipt",
```
benjaminEwhite, line 327: Minor, but i think it would good to set `send_receipt` to a var in switchboard.py, and reference that here instead of hard coding.

benjaminEwhite, line 327: Also minor, but easy enough to combine those two tests into a single one and parametrize `send_receipt` as T/F.

```
@pytest.mark.parametrize("send_email", [True, False])
def test_create_sends_email_behavior(
        self, api_client, creation_data_recurring_with_page, request, token, mocker, send_email
    ):
        url = f"{reverse('switchboard-contribution-list')}{'?send_receipt' if send_email else ''}"
        mock_handle_thank_you_email = mocker.patch("apps.contributions.models.Contribution.handle_thank_you_email")
        api_client.post(url...
        ...
       assert mock_handle_thank_you_email.called is send_receipt
```

apps/contributions/views/switchboard.py
nrh-cklimas: c2b199c36f77d3b975eb0e7f823799080113cad1

apps/contributions/views/switchboard.py
nrh-cklimas: ae6066928d0dcfc13571a98985d7fa67a55b4053

apps/contributions/tests/views/test_switchboard.py
nrh-cklimas: d1e43012cf07c34fc52fb9488044c6293fcf496c

nrh-cklimas: 7077e9a13e83de7ffa195e4b8fe6fbe9aecfa03b

benjaminEwhite (approved):

apps/contributions/views/switchboard.py
x110dc: I don't think it should be restricted to particular metadata values, but I do think it should have a parameter, as in `?send_receipt=true` and then `?send_receipt=false` or if the param is missing would have the same behavior, which is the default of `false`.

x110dc (changes requested):

apps/contributions/views/switchboard.py
nrh-cklimas: Right now it's keyed on the presence of a `send_receipt` param in the request, e.g. `/api/v1/mailchimp/contributions/?send_receipt`. Would you prefer it to only accept `?send_receipt=true` and treat any other value that may be passed as `false`?

x110dc: I think it should check as case-insensitive for "true", that is "true", "True" or any variation. I can also see a case for accepting truthy values like "yes", "1" and "y", but I don't feel strongly about it.  I guess it would be weird for "sent_receipt=yes" to not send a receipt. So my preference would be to accept any truthy value.

x110dc (approved):
Responded.

apps/contributions/tests/views/test_switchboard.py
```diff
@@ -308,7 +308,7 @@ def test_create_receipt_behavior(
     ):
         mock_handle_thank_you_email = mocker.patch("apps.contributions.models.Contribution.handle_thank_you_email")
         api_client.post(
-            f"{reverse('switchboard-contribution-list')}{'?' + SEND_RECEIPT_QUERY_PARAM if send_receipt else ''}",
+            f"{reverse('switchboard-contribution-list')}{'?' + SEND_RECEIPT_QUERY_PARAM + '=true' if send_receipt else ''}",
```
benjaminEwhite, line 311: Ideally, we should parametrize the values being sent in for this query param now that there's > 1 possibility for what is considered true.

apps/common/utils.py
```diff
@@ -173,6 +173,11 @@ def google_cloud_pub_sub_is_configured() -> bool:
     return all([settings.ENABLE_PUBSUB and settings.GOOGLE_CLOUD_PROJECT])
 
 
+def string_has_truthy_value(value: str) -> bool:
+    """Return whether a string value is considered 'truthy'."""
+    return value.lower().strip() in ["y", "yes", "true"]
```
benjaminEwhite, line 178: Should "1" be a value we're converting here?

benjaminEwhite: I'm generally all for DRYing up code, but in this case, I'm not sure it makes sense. It's conceivable one might change the values that convert to True here for some new reason, and since this function gets used in `StripePaymentMetadata_1_0`, we could end up treating a new value therein as Truthy, even if it doesn't adhere to our metadata schema.

To that point, my suggestion about adding "1" above was because it seems like if we're going to have a generic function that is about converting a set of strings, "1" seems to make sense there. But that might cause downstream problems if we allow "1" for a value in metadata. To be clear, I don't have specific knowledge that this is a problem, and it could be that in status quo it won't cause any problems.

apps/contributions/views/switchboard.py
```diff
@@ -51,7 +52,9 @@ def perform_create(self, serializer: serializers.SwitchboardContributionSerializ
         intent or subscription).
         """
         contribution: Contribution = serializer.save()
-        if self.request.query_params.get(SEND_RECEIPT_QUERY_PARAM) is not None:
+        if isinstance(qp := self.request.query_params.get(SEND_RECEIPT_QUERY_PARAM), str) and string_has_truthy_value(
```
benjaminEwhite, line 55: I believe the type check here on string is not necessary. Query params should inherently be strings.

benjaminEwhite (changes requested):

apps/common/utils.py
```diff
@@ -173,6 +173,11 @@ def google_cloud_pub_sub_is_configured() -> bool:
     return all([settings.ENABLE_PUBSUB and settings.GOOGLE_CLOUD_PROJECT])
 
 
+def string_has_truthy_value(value: str) -> bool:
+    """Return whether a string value is considered 'truthy'."""
```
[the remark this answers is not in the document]
benjaminEwhite: But to this point, looking at the test above, it does seem like going against the grain of idiomatic Python to say that "1" and "false" and "unrelated" are not truthy.

apps/common/utils.py
nrh-cklimas: Accepting "1" broke `StripeMetadataSchemaBase.normalize_boolean` because it only accepts "true", "yes", and "y" as "truthy".

apps/common/utils.py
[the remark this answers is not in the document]
nrh-cklimas: de495c3fdcb56c6fdc83cf8c407b1e06ec893765

apps/contributions/tests/views/test_switchboard.py
nrh-cklimas: 8957474cc959d2f2de8136b87bc2f7f32d75b826

apps/contributions/views/switchboard.py
nrh-cklimas: Ah yeah, forgot that `and` would short-circuit. 1f1f24ab2129e673c564ad0308d7879df41963ed

benjaminEwhite (approved):

PR #1727, opened 2025-02-27
DEV-5817 | initial mailchimp integration does not create segments

benjaminEwhite:

#### What's this PR do?

This PR solves for a race condition occurring in our mailchimp connection user flow. 

The problem was as follows:

1. User completes mailchimp connection on their site
2. They get sent back to SPA
3. They're asked to choose a mail chimp list to associate with the RP
4. That triggers a request to the configure mail chimp endpoint on server
5. Server saves mailchimp list ID on the RP
6. A post save signal runs that does NOT use transaction.on_commit() to guarantee finality. The post save signal calls setup_mailchimp_entities_for_rp_mailing_list
7. When setup_mailchimp_entities_for_rp_mailing_list is executed, if it happens soon enough, when it looks to see if rp.mailchimp_list_id is set, it finds it has not been set. 
8. An exception occurs, no segments/store/etc. created, and the system doesn't know to retry later.

The solution here is to use `transaction.on_commit` ahead of triggering side effects that rely on updated db state.

This PR also makes a tiny change to an existing test suite, changing the name of `apps.api.tests.test_mixins.TestViewSet` to `DummyViewset` as the old name was causing error log because Pytest thought that class should be runnable, as it follows test-pickup naming convention. That didn't cause test runs to fail, but it's better to not have mocks with names that pytest think refer to test classes.

Fix a bug that is blocking full setup for mailchimp integration.

Yes.

No

Existing ones updated.

N/A

No. But as noted in linked followup ticket [DEV-5910](https://news-revenue-hub.atlassian.net/browse/DEV-5910), we probably should be using `transaction.on_commit` in other places.

No.

-[DEV-5817](https://news-revenue-hub.atlassian.net/browse/DEV-5817)
-[DEV-5910](https://news-revenue-hub.atlassian.net/browse/DEV-5910)
-[ITS-2958](https://news-revenue-hub.atlassian.net/browse/ITS-2958)

No

-[DEV-5910](https://news-revenue-hub.atlassian.net/browse/DEV-5910)
-[ITS-2958](https://news-revenue-hub.atlassian.net/browse/ITS-2958)

[DEV-5910]: https://news-revenue-hub.atlassian.net/browse/DEV-5910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/organizations/tests/test_signals.py
```diff
@@ -44,12 +44,16 @@ def test_does_not_create_social_meta_if_already_exists(self, mocker):
...
+        delay_mock = mocker.patch("apps.organizations.signals.setup_mailchimp_entities_for_rp_mailing_list.delay")
         handle_rp_mailchimp_entity_setup(sender=mocker.MagicMock(), instance=rp, created=True)
         if expect_task_called:
-            setup_mc_task.delay.assert_called_once_with(rp.id)
+            on_commit_mock.assert_called_once()
+            lambda_func = on_commit_mock.call_args[0][0]
+            lambda_func()
+            delay_mock.assert_called_once_with(rp.id)
         else:
-            assert not setup_mc_task.delay.called
```
njharman, line 52: Consider leaving `assert not setup_mc_task.delay.called`.  Protecting against changes to it being wrapped by on_commit. And as documentation of what is not suppose to be called.

Or add comment explaining what on_comit wraps.

apps/organizations/tests/test_signals.py
```diff
@@ -44,12 +44,16 @@ def test_does_not_create_social_meta_if_already_exists(self, mocker):
     )
     def test_when_new_instance(self, make_rp_kwargs, expect_task_called, mocker):
         rp = RevenueProgramFactory.build(**make_rp_kwargs)
-        setup_mc_task = mocker.patch("apps.organizations.signals.setup_mailchimp_entities_for_rp_mailing_list")
+        on_commit_mock = mocker.patch("django.db.transaction.on_commit")
+        delay_mock = mocker.patch("apps.organizations.signals.setup_mailchimp_entities_for_rp_mailing_list.delay")
         handle_rp_mailchimp_entity_setup(sender=mocker.MagicMock(), instance=rp, created=True)
         if expect_task_called:
-            setup_mc_task.delay.assert_called_once_with(rp.id)
```
njharman, line 50: Doesn't this still get called?  via on_commit?  Suggest leaving in `setup_mc_task.delay.assert_called_once_with(rp.id)`, placing it after on_commit_mock.assert_called_once() [because that's the order they're executed]

njharman (approved):

apps/organizations/tests/test_signals.py
benjaminEwhite: That is still getting called, but with renamed var (delay_mock).

apps/organizations/tests/test_signals.py
benjaminEwhite: Ditto here. Just under new var name (delay_mock).

x110dc (commented):
Just one question.

apps/organizations/signals.py
```diff
@@ -65,7 +66,7 @@ def handle_rp_mailchimp_entity_setup(sender, instance: RevenueProgram, created:
         logger.info(
             "Enqueuing task to setup mailchimp entities for revenue program mailing list for RP %s", instance.id
         )
-        setup_mailchimp_entities_for_rp_mailing_list.delay(instance.id)
+        transaction.on_commit(lambda: setup_mailchimp_entities_for_rp_mailing_list.delay(instance.id))
```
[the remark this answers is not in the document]
benjaminEwhite: Addressed in e02ba136.

@x110dc this comment turned out to be important because it led me to look at unit test again and I realized that the test was possibly trivially true as written. I dug some more and learned there's a simple way to ensure transactions occur in scope of test, which is by setting `@pytest.mark.django_db(transaction=True)`.

x110dc (approved):

PR #1734, opened 2025-03-11
DEV-5864 | Fall back to sub current period end in certain cases

benjaminEwhite:

n/a

#### What's this PR do?

- Changes the serializer for contributions to fall back to the next payment date if the first payment date isn't available due to a migrated recurring contribution which hasn't had a reucrrence yet.
- Changes the SPA to handle the scenario where first payment date isn't set. This will only happen if a contribution has no payments and no future ones.

Handles migrated data better.

Yes with one TBD.

I think no.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5864

No.

No.


apps/contributions/serializers.py
```diff
@@ -750,6 +750,22 @@ def update(self, instance, validated_data):
...
+            obj.status == ContributionStatus.PAID
+            and obj.interval != ContributionInterval.ONE_TIME
+            and (metadata := obj.contribution_metadata)
+            and metadata.get("schema_version") == "1.3"
+            and (date := obj.next_payment_date)
+        ):
+            return date
+        return None
+
+    def to_representation(self, instance):
```
nrh-cklimas, line 766: We want to drop this--not doing anything

spa/src/components/portal/ContributionsList/ContributionItem/ContributionItem.tsx
```diff
@@ -71,7 +71,9 @@ export const ContributionItem = forwardRef<HTMLAnchorElement, ContributionItemPr
           <IntervalIcon aria-label={getFrequencyAdjective(contribution.interval)} />
         </IntervalIconContainer>
         <DateContainer>
-          <CreatedDate data-testid="first-payment-date">{formatDate(contribution.first_payment_date)}</CreatedDate>
+          <CreatedDate data-testid="first-payment-date">
```
nrh-cklimas, line 74: Need to change also on detail view

Probably use something nicer for placeholder dash

apps/contributions/serializers.py
```diff
@@ -750,6 +750,19 @@ def update(self, instance, validated_data):
...
         raise NotImplementedError("update is not supported on this serializer")
 
+    def get_first_payment_date(self, obj: Contribution) -> datetime | None:
+        if (date := obj.first_payment_date) is not None:
+            return date
+        if (
+            obj.status == ContributionStatus.PAID
+            and obj.interval != ContributionInterval.ONE_TIME
+            and (metadata := obj.contribution_metadata)
+            and metadata.get("schema_version") == "1.3"
```
njharman, line 760: Don't understand why it's limited 1.3 schema. Only one vulnerable to this bug?  Wouldn't we want to provide "good" date regardless of version?

If important to only schema 1.3, consider having test that verifies None is returned for other schema versions. something like

`v1_4_contribution_with_next_payment_date_and_no_payment", False, False`

njharman (changes requested):

apps/contributions/serializers.py
nrh-cklimas: I guess the best way to put it is, the fix we're implementing here is only valid for metadata 1.3, aka migrated data. If this situation occurs with other types of contributions, then something different has gone wrong. I guess there's an argument for raising an error in this case, since we think it's a "should never happen" scenario, but I'd rather raise a warning and keep going, if anything, so that users can continue, even if they see odd data. What do you think of that?

Agree with you about regardless of what happens, we should add the test you're suggesting.

njharman (approved):

nrh-cklimas (approved):
Approving on behalf of Ben, who can't do it because he's original author.

PR #1739, opened 2025-03-17
DEV-4688: Make phone character limit 20

nrh-cklimas:

n/a

#### What's this PR do?

Enforces a 20-character limit on phone numbers on both the backend and frontend.

Fixes a validation problem.

Yes.

No.

Yes.

n/a

Hard-coded a validation message that comes from DRF--not sure if there's a better way to do it.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-4688

No.

No.


apps/contributions/serializers.py
```diff
@@ -211,7 +211,7 @@ class AbstractPaymentSerializer(serializers.Serializer):
 
     # Page id is a nice shortcut for getting the page, instead of page_slug + rp_slug
     page_id = serializers.IntegerField(required=False)
-    phone = serializers.CharField(max_length=40, required=False, allow_blank=True)
+    phone = serializers.CharField(max_length=20, required=False, allow_blank=True)
```
benjaminEwhite, line 214: Fine to update this since field already exists and we want to ensure consistency, but AFAICT this field is not used anywhere.  `AbstractPaymentSerializer` base class for `StripeOneTimePaymentSerializer` and `StripeRecurringPaymentSerializer`, which in turn are only used in `StripePaymentManager`, which in turn is only used in quarantine management (accepting and rejecting payments).

I commented out this line and ran tests locally with `pytest app/contributions` and no failing tests. I believe this is safe to delete, based on that fact and looking through usage of this class in code (but don't take my word on that specific point, would be good to have second set of eyes).

If you decide not to remove, then at minimum would be good to define this magic number in a const and refer to it in both places.

Quarantine doesn't care about phone number, and certainly doesn't write to that field, so I think we would be safe to remove this field entirely.

apps/contributions/serializers.py
```diff
@@ -276,7 +276,7 @@ class BaseCreatePaymentSerializer(serializers.Serializer):
         max_length=80, write_only=True, required=False, allow_blank=True, default=""
     )
     agreed_to_pay_fees = serializers.BooleanField(default=False, write_only=True)
-    phone = serializers.CharField(max_length=40, required=False, allow_blank=True, write_only=True, default="")
+    phone = serializers.CharField(max_length=20, required=False, allow_blank=True, write_only=True, default="")
```
benjaminEwhite, line 279: Would be worth it to leave a code comment here explaining significance of 20 here.

apps/contributions/tests/test_serializers.py
```diff
@@ -459,24 +459,29 @@ def test_validate_reason_for_giving(
...
-            ({"phone": ""}, "donation_page", True),
-            ({"phone": "something"}, "donation_page", True),
-            ({"phone": ""}, "donation_page_with_conditionally_required_phone_element", False),
-            ({"phone": "something"}, "donation_page_with_conditionally_required_phone_element", True),
+            ({"phone": ""}, "donation_page", None),
+            ({"phone": "something"}, "donation_page", None),
+            ({"phone": ""}, "donation_page_with_conditionally_required_phone_element", GENERIC_BLANK),
+            ({"phone": "something"}, "donation_page_with_conditionally_required_phone_element", None),
+            (
+                {"phone": "more than 20 characters"},
```
benjaminEwhite, line 469: given reference to "magic" 20 here, might be worthwhile to put that value in an appropriately named const var in serializers and refer to it here instead of hard coding (and having const also would be self-documenting given right name).

benjaminEwhite (changes requested):

apps/contributions/serializers.py
nrh-cklimas: 1707f6283d59907e0ce7d87eddd8e50c4eb95b28

apps/contributions/serializers.py
nrh-cklimas: 1707f6283d59907e0ce7d87eddd8e50c4eb95b28

apps/contributions/tests/test_serializers.py
nrh-cklimas: I was looking at how we define error messages and realized that we could reference the built-in error message in the test. 6deb87094cc4c5bbc5221c3e3d413060b19504ea has my idea but I can also go with the fully custom approach if you think that's better.

benjaminEwhite: I prefer that approach to custom. Leaving this comment here as reminder, but feel free to close once that change is in PR.

benjaminEwhite (approved):

benjaminEwhite (approved):

apps/contributions/serializers.py
```diff
@@ -211,7 +211,6 @@ class AbstractPaymentSerializer(serializers.Serializer):
 
     # Page id is a nice shortcut for getting the page, instead of page_slug + rp_slug
     page_id = serializers.IntegerField(required=False)
-    phone = serializers.CharField(max_length=40, required=False, allow_blank=True)
```
njharman, line 214: This removes phone from the following; which seems out of scope for bug fix.

- class StripeOneTimePaymentSerializer(AbstractPaymentSerializer)
- class StripeRecurringPaymentSerializer(AbstractPaymentSerializer)

apps/contributions/tests/test_serializers.py
```diff
@@ -459,24 +460,29 @@ def test_validate_reason_for_giving(
...
+                "donation_page",
+                CharField.default_error_messages["max_length"].format(max_length=20),
```
njharman, line 472: nice! (The usage of Charfield to match its error message)

apps/contributions/tests/test_serializers.py
```diff
@@ -459,24 +460,29 @@ def test_validate_reason_for_giving(
...
-        self, input_data, page_fixture, expect_valid, minimally_valid_contribution_form_data, request
+        self, input_data, page_fixture, error_message, minimally_valid_contribution_form_data, request
     ):
         data = minimally_valid_contribution_form_data | input_data
         data["page"] = request.getfixturevalue(page_fixture).id
         serializer = self.serializer_class(data=data, context={"request": APIRequestFactory().post("")})
-        assert serializer.is_valid() is expect_valid
-        if expect_valid is False:
+        assert serializer.is_valid() is not bool(error_message)
+        if error_message:
```
njharman, line 483: Suggest

```
if error_message:
   ...
else:
    assert serializer.is_valid()
```

njharman (changes requested):

apps/contributions/tests/test_serializers.py
nrh-cklimas: I tried this and got `AssertionError: You must call .is_valid() before accessing .errors.` Could do this...

```python
 if error_message:
            assert not serializer.is_valid()
            assert set(serializer.errors.keys()) == {"phone"}
            assert serializer.errors["phone"][0] == error_message
        else:
            assert serializer.is_valid()
```

But not sure that's any better?

apps/contributions/serializers.py
nrh-cklimas: See [discussion here](https://github.com/newsrevenuehub/rev-engine/pull/1739#discussion_r2001014715). I don't have much of an opinion around whether we include this or not, other than generally being in favor of removing dead code.

njharman (approved):

PR #1742, opened 2025-03-19
DEV-5929 | Validate AC server URL and access token

benjaminEwhite:

Note that this PR initially has DEV-5899-activecampaign-setup as its base because it merged in those changes.  DEV-5899 will merge before this moves across board.

#### What's this PR do?

This PR adds validation of the user submitted values for AC API URL and AC API KEY when users are going through AC integration connection flow.

The backend makes a request to an AC endpoint using submitted credentials. If that request fails, a 400 error + message is sent back to spa, prompting display of message why the form submission failed.

Better insurance that user-submitted values for AC are correct and that integration will work.

Yes.

No.

Yes.

N/A

No

N/A

[DEV-5929](https://news-revenue-hub.atlassian.net/browse/DEV-5929)

No

No

[DEV-5929]: https://news-revenue-hub.atlassian.net/browse/DEV-5929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spa/src/components/settings/Integration/IntegrationCard/ActiveCampaignIntegrationCard/ConnectionModal/ConnectionModal.tsx
```diff
@@ -54,12 +60,19 @@ export function ConnectionModal({ onClose, open }: ConnectionModalProps) {
...
-    } catch (error) {
-      setUpdateError(
-        <>
-          <strong>Failed to save API information.</strong> Please try again.
-        </>
-      );
+    } catch (error: any) {
+      const axiosError = error as AxiosError<SaveRevenueProgramActiveCampaignUrlAndKeyValidationErrors>;
+      const message =
+        axiosError?.response?.data?.non_field_errors?.[0] === AC_URL_OR_TOKEN_VALIDATION_ERROR ? (
```
nrh-cklimas, line 66:
```suggestion
        axiosError.response.data.non_field_errors?.[0] === AC_URL_OR_TOKEN_VALIDATION_ERROR ? (
```

In all cases, `axiosError` should be defined. I think also `axiosError.response.data` will always have a value?

spa/src/hooks/useConnectActiveCampaign.ts
```diff
@@ -18,6 +18,13 @@ export interface RevenueProgramActiveCampaignStatus {
   slug: string;
 }
 
+export interface SaveRevenueProgramActiveCampaignUrlAndKeyValidationErrors {
+  /**
+   * Non-field error message.
+   */
+  non_field_errors?: Array<string>;
```
nrh-cklimas, line 25:
```suggestion
  non_field_errors?: string[];
```

apps/organizations/serializers.py
```diff
@@ -273,10 +283,45 @@ class Meta(BaseActiveCampaignRevenueProgram.Meta):
             "activecampaign_server_url",
         ]
 
+    def confirm_activecampaign_url_and_token(self, url: str, token: str) -> bool:
+        """Confirm the ActiveCampaign URL and token are valid."""
+        logger.info("Confirming ActiveCampaign URL and token for rp %s", self.instance)
+        try:
+            response = requests.get(
+                url=f"{url}/api/3/users/me/",
```
njharman, line 291: Doesn't `url` potentially come from external "users"? If so, isn't this a security issue in that url could lead anywhere? endpoint that always returns 200.

If not, ~deserves a docstring/comment~ ... the serializer name `...ForSpa...` strongly indicates the endpoint is internal. But that pushes "where does url come from and has it been sanitized" (is SPA getting it from 3rd party) even further away from security concern.

apps/organizations/serializers.py
```diff
@@ -273,10 +283,45 @@ class Meta(BaseActiveCampaignRevenueProgram.Meta):
             "activecampaign_server_url",
         ]
 
+    def confirm_activecampaign_url_and_token(self, url: str, token: str) -> bool:
```
njharman, line 286: Suggest leading underscore as this method is internal to serializer / used only by validate()

njharman (changes requested):

apps/organizations/serializers.py
```diff
@@ -260,11 +262,19 @@ class Meta:
...
+    activecampaign_server_url: str
+
+
 class ActiveCampaignRevenueProgramForSpaSerializer(BaseActiveCampaignRevenueProgram, UpdateFieldsBaseSerializer):
     """A serializer that allows PATCHing of additional fields."""
 
-    activecampaign_access_token = serializers.CharField(max_length=100, write_only=True, required=False)
-    activecampaign_server_url = serializers.URLField(required=False)
+    activecampaign_access_token = serializers.CharField(max_length=100, write_only=True)
+    # should this also have a max_length?
```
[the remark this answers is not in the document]
benjaminEwhite: Removed in 38dbbc74

apps/organizations/serializers.py
benjaminEwhite: I agree with this concern and wondered about it myself. The URL here is meant to be harvested from ActiveCampaign. It's unclear to me whether or not we can do some kind pattern matching to ensure that it's for AC campaign. When I went through the connection flow, the URL api-us1.com as root plus TLD, but I don't know if there's a guarantee of that (also, I was surprised to discover that api-us1.com is in fact owned by active campaign -- would have thought some more generic service would have snatched up that URL).

@nrh-cklimas are you able to weigh in here? Do you happen to know if there is a discrete set of hosts+TLDs that we can validate against?

PR #1753, opened 2025-04-07
DEV-5764: Alphabetize RP lists in Django admin

nrh-cklimas:

n/a

#### What's this PR do?

Changes sorting of revenue programs in Django admin when they're on a dropdown menu as a foreign key.

Makes support a lot easier, as the default sort is by insert order.

Yes.

No.

No. I could use guidance here. I'm not sure if unit tests are necessary because we're just changing a sort order, not the items returned.

n/a

Lack of unit tests as discussed above.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-5764

No.

No.


apps/users/admin.py
```diff
@@ -91,6 +91,10 @@ def get_form(self, request, obj, **kwargs):
         if user_field := form.base_fields.get("user"):
             user_field.limit_choices_to = Q(roleassignment__isnull=True)
 
+        # Alphabetize the organization list
```
benjaminEwhite, line 94: Noting here how I think we can test this. This is quick version (no type hints, would be better to use `reverse(...)` instead of hard coded URL):

```
@pytest.mark.django_db
class TestRoleAssignmentAdmin:

    @pytest.fixture
    def authed_client(self, superuser, client):
        client.force_login(superuser)
        return client

    @pytest.fixture
    def orgs(self):
        org1 = OrganizationFactory(name="Alpha")
        org2 = OrganizationFactory(name="Charlie")
        org3 = OrganizationFactory(name="Delta")
        org4 = OrganizationFactory(name="Bravo")
        return [org1, org2, org3, org4]

    def test_org_dropdown_is_alphabetized(self, authed_client, orgs):
        ra = RoleAssignmentFactory()
        assert Organization.objects.count() >= len(orgs)
        # prove later assertions aren't trivial (i.e., not already sorted)
        assert Organization.objects.all().values_list("name") != Organization.objects.all().order_by(
            "name"
        ).values_list("name")
        response = authed_client.get(f"/nrhadmin/users/roleassignment/{ra.id}/change/")
        soup = BeautifulSoup(response.content)
        select = soup.find("select", {"name": "organization"})
        org_options = [opt.text for opt in select.find_all("option") if opt.attrs.get("value")]
        assert org_options == sorted(org_options)
```

Should be able to use similar approach for testing BenefitAdmin

apps/organizations/admin.py
```diff
@@ -175,6 +175,15 @@ class BenefitAdmin(RevEngineBaseAdmin):
 
     fieldsets = ((None, {"fields": ("name", "description", "revenue_program")}),)
 
+    def get_form(self, request, obj, **kwargs):
+        form = super().get_form(request, obj, **kwargs)
+
+        # Alphabetize the revenue program list
+        if rp_field := form.base_fields.get("revenue_program"):
```
benjaminEwhite, line 182: Doesn't seem like we should need conditionality here.  Unclear to me how we would even be able to test state where this field isn't part of base fields.

apps/users/admin.py
```diff
@@ -91,6 +91,10 @@ def get_form(self, request, obj, **kwargs):
...
+        if org_field := form.base_fields.get("organization"):
```
benjaminEwhite, line 95: Ditto here -- don't think we need the conditionality.

benjaminEwhite (changes requested):

apps/users/admin.py
nrh-cklimas: 392a218c5d820e21be92e1716d2578af4148ed9e

apps/organizations/admin.py
nrh-cklimas: 45f3ced98bdac4ef8def2e707ecff4ede47510cf

apps/users/admin.py
nrh-cklimas: e058629eb221fc020d9c8816f83d41baa03ce7d4

apps/organizations/tests/test_admin.py
```diff
@@ -9,172 +9,83 @@
...
-        assert soup.find("input", {"name": "facebook_pixel_id"}) is not None
-        assert soup.find("input", {"name": "twitter_handle"})
-        assert soup.find("input", {"name": "website_url"})
-        assert soup.find("input", {"name": "allow_offer_nyt_comp"})
+@pytest.mark.django_db
+class TestBenefitLevelAdmin:
+    @pytest.fixture
+    def authed_client(self, superuser, client):
+        client.force_login(superuser)
+        return client
```
benjaminEwhite, line 57: Might as well define this fixture in module-level scope so don't have to repeat in the two test classes that use it.

apps/users/tests/test_admin.py
```diff
@@ -1,48 +1,78 @@
...
-        self.index_url = "/"
-        self.user = user_model.objects.create_superuser(email=email, password=password)
-        self.client = Client()
-        self.client.login(email=email, password=password)
+@pytest.mark.django_db
+class TestRoleAssignmentAdmin:
+    @pytest.fixture
+    def authed_client(self, superuser, client):
+        client.force_login(superuser)
+        return client
```
benjaminEwhite, line 21: Ditto here re: moving authed_client into module scope.

I could even see moving to conftest.py since it's being used in two separate files, spanning two separate apps.

apps/users/tests/test_admin.py
```diff
@@ -1,48 +1,78 @@
...
+
+    def test_views_stand_up(self, authed_client):
+        authed_client.get("/nrhadmin/users/user/")
+        authed_client.get("/nrhadmin/users/user/add/")
+        authed_client.get("/nrhadmin/users/roleassignment/")
+        authed_client.get("/nrhadmin/users/roleassignment/add/")
```
benjaminEwhite, line 56: Would prefer to see `reverse` here with named URLs.

Easiest way to get the URL names would be:

```
python manage.py show_urls | grep /nrhadmin/users/
```

The right most item in each row will be the URL name.

Only going to make this comment here, but would be nice to do throughout.

benjaminEwhite (approved):
Changes look good. Had a couple of minor suggestions re: tests, but those don't require re-review.

apps/organizations/tests/test_admin.py
nrh-cklimas: 9f7574c09b3721af1ce35c37da01d3ab5c1b6876

apps/users/tests/test_admin.py
nrh-cklimas: 9f7574c09b3721af1ce35c37da01d3ab5c1b6876

apps/users/tests/test_admin.py
nrh-cklimas: c8ccb2036ae6267d4aa7af24f1b95c709607be5e

x110dc (approved):

PR #1757, opened 2025-04-11
DEV-4277: Allow deletion of failed contributions

nrh-cklimas:

n/a

#### What's this PR do?

Allows deletion of failed contributions by the SPA.

If the user initially enters an invalid payment method and then closes the payment modal, we should delete the contribution to reflect that the user canceled out of it, as we do for other contributions.

Yes.

No.

Yes.

n/a

No..?

No.

https://news-revenue-hub.atlassian.net/browse/DEV-4277

No.

Discovered https://news-revenue-hub.atlassian.net/browse/DEV-6036 diagnosing this, but it's not directly related.


apps/contributions/models.py
```diff
@@ -667,7 +667,7 @@ def cancel(self):
         # this is specifically used when a user clicks "back" on the second payment form in checkout flow. it's not
         # necessarily intended to be a general-purpose cancellation method (i.e., it is not appropriate for the `.destroy` method
         # on the API endpoint)
```
benjaminEwhite, line 669: Just noticing that this comment is likely stale — in fact the `.destroy` is _exactly_ where we use this method. I'm guessing at some point this point made sense, but now it's just confusing.

Think we should remove the bit about not being appropriate for `destroy`.

apps/contributions/models.py
```diff
@@ -667,7 +667,7 @@ def cancel(self):
...
-        if self.status not in (ContributionStatus.PROCESSING, ContributionStatus.FLAGGED):
+        if self.status not in (ContributionStatus.FAILED, ContributionStatus.PROCESSING, ContributionStatus.FLAGGED):
```
benjaminEwhite, line 670: Think it's worthwhile to have a code comment explaining why these statuses make sense to allow canceling (or at least for FAILED, as it's not immediately obvious why people would be canceling failed status contributions).

apps/contributions/tests/test_models.py
```diff
@@ -627,6 +628,11 @@ def test_cancel_when_one_time(self, status, monkeypatch, mocker):
             (ContributionStatus.PROCESSING, "annual_subscription", False),
             (ContributionStatus.FLAGGED, "monthly_subscription", False),
             (ContributionStatus.FLAGGED, "annual_subscription", False),
+            # Failed contributions can never have a payment method attached,
+            # because they become failed by a payment method failing to attach
+            # (e.g. the card has expired).
```
benjaminEwhite, line 633: While this is true with regards to checkout flow, it's not generally true. For instance, a recurring contribution that was once PAID could move to FAILED if there was an unsuccessful charge. I believe in that case a `payment_intent.failed` event would be emitted.

apps/contributions/views/checkout.py
```diff
@@ -57,10 +57,15 @@ def get_serializer_context(self):
 
     def destroy(self, request, *args, **kwargs):
         contribution = self.get_object()
-        if contribution.status not in (ContributionStatus.PROCESSING, ContributionStatus.FLAGGED):
+        if contribution.status not in (
+            ContributionStatus.FAILED,
+            ContributionStatus.PROCESSING,
+            ContributionStatus.FLAGGED,
```
benjaminEwhite, line 63: I was about to write that it might make sense to store these statuses in a var and reference that var here and in .cancel().

But thinking about it more, it seems odd that we're checking status is equal to one of these in both places, especially since .cancel() is only called in this context.

I think that might argue for getting rid of the model method altogether and putting all the logic in view file.  There are no other callers at the moment for .cancel(), and in that method we're enforcing assumptions that are satisfied (for now) only in context of checkout cancelation.

benjaminEwhite (changes requested):

apps/contributions/views/checkout.py
nrh-cklimas: If I add flagged+has payment method to the permutations, I get a test failure:

```
        elif has_payment_method_id:
>           mock_retrieve_pm.assert_called_once_with(
                contribution.provider_payment_method_id,
                stripe_account=contribution.revenue_program.stripe_account_id,
            )
E           AssertionError: Expected 'mock' to be called once. Called 0 times.
```

I think this stems from this code:

```python
        elif self.status == ContributionStatus.FLAGGED and self.provider_payment_method_id:
            stripe.PaymentMethod.retrieve(
                self.provider_payment_method_id,
                stripe_account=self.stripe_account_id,
            ).detach()
```

I can't tell if we need to modify the test or the actual code here because I'm unsure if we want a payment method to be detached in this scenario.

apps/contributions/tests/test_models.py
nrh-cklimas: If I add flagged+has payment method to the permutations, I get a test failure:

```
        elif has_payment_method_id:
>           mock_retrieve_pm.assert_called_once_with(
                contribution.provider_payment_method_id,
                stripe_account=contribution.revenue_program.stripe_account_id,
            )
E           AssertionError: Expected 'mock' to be called once. Called 0 times.
```

I think this stems from this code:

```python
        elif self.status == ContributionStatus.FLAGGED and self.provider_payment_method_id:
            stripe.PaymentMethod.retrieve(
                self.provider_payment_method_id,
                stripe_account=self.stripe_account_id,
            ).detach()
```

I can't tell if we need to modify the test or the actual code here because I'm unsure if we want a payment method to be detached in this scenario.

apps/contributions/models.py
nrh-cklimas: bcf7914dce3754568e5ed0061bffa604b76408b8

apps/contributions/models.py
nrh-cklimas: Rolled this into bcf7914dce3754568e5ed0061bffa604b76408b8

apps/contributions/views/checkout.py
nrh-cklimas: I agree that it feels weird to check status in both places, but my vote would be to remove logic from the view and trust the model to raise the right error. My main reasoning is that it doesn't look like in other views we do a lot of interaction with Stripe.

apps/contributions/tests/test_models.py
benjaminEwhite: My comment was just about the code comment. I don't think test or object code need to be modified. I just we shouldn't say "Failed contributions can never have a payment method attached" because that is not true.

apps/contributions/views/checkout.py
benjaminEwhite: I can get on board with doing it in either place, but I think not repeating ourselves is a good idea.

benjaminEwhite (changes requested):

apps/contributions/tests/test_models.py
nrh-cklimas: Ah, I misunderstood--thought you thought that scenario should be tested. b8b5d5cd885a006304011caf6595fc23707bd0a1

benjaminEwhite (approved):

apps/contributions/models.py
```diff
@@ -664,10 +664,13 @@ def create_stripe_subscription(
...
     def cancel(self):
-        # this is specifically used when a user clicks "back" on the second payment form in checkout flow. it's not
-        # necessarily intended to be a general-purpose cancellation method (i.e., it is not appropriate for the `.destroy` method
-        # on the API endpoint)
-        if self.status not in (ContributionStatus.PROCESSING, ContributionStatus.FLAGGED):
+        # This is used when a user clicks "back" on the second payment form in
+        # checkout flow. We allow failed contributions to be canceled because if
+        # the user tries to attach an invalid payment method to a contribution
+        # in the second payment form, the contribution goes into that state due
+        # to Stripe emitting a `payment_intent.payment_failed` webhook event.
```
njharman, line 671: I'm not sure if "this" is just this if statement or the method as a whole.  I thought method as there is a blank line after it. But reading/context maybe only about ContributionStatus.FAILED

Consider any/all of the following:

1. Move introductory, "this is the what/why of this method", to docstring.
2. Break out (separate with blank line from other comments) comment specific to one statement (the if). And no blank line between comment and it's subject.
3. Prefer "Method", "ContributionStatus.FAILED" over pronounish "this".

njharman (approved):

apps/contributions/models.py
nrh-cklimas: 0d8517d15c7c6ffba2867ed283bc7026de4ad46c

x110dc (approved):

PR #1760, opened 2025-04-18
DEV-5825 | Activity logs in Django admin for contributor-intitiated recurring contribution cancellations

benjaminEwhite:

#### What's this PR do?

This PR adds a new "Activity Logs" view to the Django admin. Longer term, we expect this view to show a variety of types of activity logs, but for first release, we are showing a log of contributor-initiated recurring contribution cancelations. 

For each instance of a contributor canceling a recurring contribution in the portal, we will see an entry in the Activity Log section of the Django admin.

Screenshot 2025-04-18 at 12 41 37 PM

The activity log model is polymorphic. Its key fields are actor, object, and action.  The actor is the person who did the activity, the object is what they acted on, and the action is the verb describing what they did. 

For the initial activity logs, the actor will be a Contributor, the action will be "canceled" and the object will be a Contribution.  The model will support arbitrary actors, objects, and actions in the future.

Note that this PR contains a migration but it adds a new model and now rows, so run time should be trivial.

Allow support team to more easily provide support to orgs around canceled contributions. 

Yes

No

Yes

We should onboard the key users of this tool once it's done (but it's pretty simple and that won't be very involved).

I don't believe there are smells or tech debt as such, but I do note that we do optimistic activity log creation for contributor initiated contribution cancelation. A contribution's status only gets updated to "canceled" after a Stripe webhook has been processed that is generated in response to RE deleting the subscription. 

While it would be possible to create activity logs concurrent with processing such webhooks that would require transmitting state about a cancellation being contributor-initiated into the processing routine, and that would take considerably more effort. As an initial prototype, the optimistic log creation seems like a good first step, and there is little negative impact if an activity log gets created but ultimately the contribution is not successfully canceled (to this point, the activity log links back to the contribution detail view, where one can easily see that contribution status).

No

[DEV-5825(https://news-revenue-hub.atlassian.net/browse/DEV-5825)

No

No


apps/contributions/views/portal.py
```diff
@@ -195,4 +195,18 @@ def handle_delete(self, contribution: Contribution) -> Response:
...
+            # the activity log entry in the webhook handler, but that would require a more complex implementation and we want to keep this
+            # simple.
+            contribution.create_contributor_canceled_contribution_activity_log()
+        # We're generically catching on Exception here because we don't want to fail the request if the activity log creation fails.
+        except Exception:
+            logger.exception(
+                "create_contributor_canceled_contribution_activity_log returned an exception trying to create activity log for "
+                "contribution %s",
+                contribution.pk,
+            )
```
njharman, line 211: Suggest (strongly) to move this into `create_contributor_canceled_contribution_activity_log()`

The interface for creating activity logs should guarantee "safety", that is not raise. So, we don't have to try/except everytime we want to log action.

apps/activity_log/admin.py
```diff
@@ -0,0 +1,92 @@
...
+        app_label = obj._meta.app_label
+        model_name = obj._meta.model_name
+        url_name = f"admin:{app_label}_{model_name}_change"
+        try:
+            return reverse(url_name, args=[obj.pk])
+        except NoReverseMatch:
+            logger.info(
+                "No reverse match for %s detail view",
+                url_name,
+            )
```
njharman, line 64: Not covered, see comment on test

apps/activity_log/tests/test_admin.py
```diff
@@ -0,0 +1,72 @@
...
+        link = soup.find(class_=LINKED_ACTOR_CLASSNAME)
+        assert link["href"] == reverse(
+            "admin:contributions_contributor_change", args=[activity_log.actor_content_object.pk]
+        )
+        assert link.text == str(activity_log.actor_content_object)
+
+    def test_linked_actor_object_when_no_url(
+        self, activity_log: ActivityLog, mocker: pytest_mock.MockerFixture
+    ) -> None:
+        mocker.patch("apps.activity_log.admin.ActivityLogAdmin._get_admin_url_for_linked_object", return_value=None)
```
njharman, line 50: Consider adding/changing test to one that sets obj.actor_content_object to None.

Mocking private method prevents that method from being covered. Also from the `if related_obj:` line in `linked_actor_object()` from every being False.

apps/contributions/views/portal.py
njharman: Also adding coverage would be more direct, which this exception currently lacks.

njharman (changes requested):

apps/activity_log/admin.py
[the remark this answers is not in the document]
benjaminEwhite: b8614faed

apps/activity_log/admin.py
```diff
@@ -0,0 +1,92 @@
...
+            return None
+
+    def linked_actor_object(self, obj: "ActivityLogAdmin") -> str:
+        related_obj = obj.actor_content_object
+        if related_obj:
+            obj_text = Truncator(str(related_obj)).chars(70)
+            url = self._get_admin_url_for_linked_object(related_obj)
+            if url:
```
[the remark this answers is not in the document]
benjaminEwhite: 581fbe6fc7b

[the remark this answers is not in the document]
benjaminEwhite: 581fbe6fc7b

apps/contributions/views/portal.py
benjaminEwhite: Good point about guarantee for callers. dbafd79b4d4

apps/activity_log/admin.py
benjaminEwhite: d4006f10798

apps/activity_log/tests/test_admin.py
benjaminEwhite: d4006f10798

njharman (approved):

apps/activity_log/models.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    actor_object_id = models.PositiveIntegerField(null=True, blank=True)
+    actor_content_object = GenericForeignKey("actor_content_type", "actor_object_id")
+
+    activity_object_content_type = models.ForeignKey(
+        ContentType, on_delete=models.CASCADE, related_name="activity_object"
+    )
+    activity_object_object_id = models.PositiveIntegerField(null=False, blank=False)
+    activity_object_content_object = GenericForeignKey("activity_object_content_type", "activity_object_object_id")
+    """The object that was acted upon."""
```
x110dc, line 40: No big deal, but just looks weird to me as a triple-quote comment instead of a hash comment.

apps/activity_log/models.py
```diff
@@ -0,0 +1,59 @@
...
+
+    action = models.CharField(max_length=255, choices=ACTIVITY_ACTION_CHOICES)
+
+    def __str__(self):
+        actor_string = (
+            f"{self.actor_content_object.__class__.__name__} #{self.actor_content_object.pk}"
+            if self.actor_content_object
+            else "System"
```
x110dc, line 48: "System" seems too vague to me. System as in RevEngine batch job? As in Switchboard-initiated, as in webhook-initiated? But maybe these distinctions can be left for clarification in further iterations. But as I understand it we are only recording contributor-initiated cancelation at first, so why is this needed at all yet?

apps/contributions/models.py
```diff
@@ -1383,6 +1384,28 @@ def update_subscription_amount(self, amount: int, donor_selected_amount: float)
...
+                logger.warning("Cannot create activity log for contribution %s because it has no contributor", self.pk)
+                return None
+            return ActivityLog.objects.create(
+                actor_content_object=contributor,
+                activity_object_content_object=self,
+                action=ActivityLog.CANCEL,
+            )
+        # We're generically catching on Exception here because we don't want callers to fail if something goes wrong
+        # creating an activity log.
+        except Exception:
```
x110dc, line 1403: Don't we need a #noqa here because we're catching `Exception` or is that not something our CI checks for? If not, should we enable that?

apps/contributions/views/portal.py
```diff
@@ -195,4 +195,10 @@ def handle_delete(self, contribution: Contribution) -> Response:
                 contribution.id,
             )
             return Response({"detail": "Problem canceling contribution"}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+        # Note that we optimistically create an activity log. Transaction finality is not guaranteed and the contribution only gets
+        # marked as canceled after the Stripe webhook is processed. If the webhook fails, we could end up with an activity log
+        # entry for a contribution that is not actually canceled. This should not happen frequently. The alternative would be to create
+        # the activity log entry in the webhook handler, but that would require a more complex implementation and we want to keep this
+        # simple.
+        contribution.create_contributor_canceled_contribution_activity_log()
```
x110dc, line 203: Again, this can probably wait for future iterations, but this feels like an ideal use case for a context manager or maybe a decorator.

apps/activity_log/models.py
benjaminEwhite: d54dd3299

apps/activity_log/models.py
benjaminEwhite: e2845e7a61a

apps/contributions/models.py
benjaminEwhite: That's not something we look for. By default we do look for bare except via [Ruff E722](https://docs.astral.sh/ruff/rules/bare-except/).

apps/contributions/views/portal.py
benjaminEwhite: Agree. Really like the idea of doing it as a decorator.  FWIW, that's how I'm thinking we'll want to do Python-side distributed tracing.

Think once we add a next activity log type, that will be good time to explore this.

x110dc (approved):

PR #1770, opened 2025-04-30
DEV-5562 | expose active campaign disconnection routine from dev 5505 as a management command

benjaminEwhite:

This PR initially takes DEV-5505 as its base, as it branches from those changes.

#### What's this PR do?

This PR wraps the disable ActiveCampaign functionality from DEV-5505 in a management command.

To do this, I refactored the existing `disable_mailchimp_integration` command, renaming it `disable_integration`. The command now takes a required option, `--integrations` which is a list of one or both of the following: mailchimp, activecampaign.

The respective disable function will be run for integrations specified in call args.

Sometimes it's convenient to be able to "zero out" an org's integrations if there's a configuration problem or if something has gone wrong with automatic downgrade.

Yes.

No.

Yes.

N/A

No.

No.

-[DEV-5562](https://news-revenue-hub.atlassian.net/browse/DEV-5562)
-[DEV-5505](https://news-revenue-hub.atlassian.net/browse/DEV-5505)

No

No

[DEV-5562]: https://news-revenue-hub.atlassian.net/browse/DEV-5562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/organizations/management/commands/disable_integration.py
benjaminEwhite, line 1: Although not reflected in git history, I initially renamed `disable_mailchimp_integration.py` to `disable_integration.py` and then made changes to the file to support both integrations.

apps/organizations/management/commands/disable_integration.py
```diff
@@ -0,0 +1,49 @@
+from django.core.management.base import BaseCommand, CommandParser
+
+from apps.organizations.models import RevenueProgram
+
+
+class Command(BaseCommand):
+    """Command to disable select integrations for a revenue program.
+
+    By default will disable both Mailchimp and ActiveCampaign integrations.
```
njharman, line 9: I believe there is no default. Or, the default is to do nothing.

apps/organizations/tests/test_management.py
```diff
@@ -81,23 +81,66 @@ def test_stripe_apple_pay_domain_verify_called(self, mocker):
...
+
+    @pytest.fixture
+    def mock_gcs_provider(self, mocker: pytest_mock.MockerFixture):
+        mocker.patch("apps.organizations.models.GoogleCloudSecretProvider")
+
+    @pytest.mark.usefixtures("mock_gcs_provider")
+    @pytest.mark.parametrize(
+        "integrations",
+        [
+            ("mailchimp",),
```
njharman, line 101: consider adding () empty tuple and verifying that neither integration is run. It's a noop, doesn't "crash".

njharman (approved):

apps/organizations/management/commands/disable_integration.py
benjaminEwhite: I had a commit I forgot to push. 2999c47ad9

apps/organizations/tests/test_management.py
benjaminEwhite: That's already tested via test_requires_integrations below (and it raises an error in that case because this is a required arg).

x110dc (approved):

PR #1777, opened 2025-05-05
DEV-6139: Hide Sentry environment when tearing down review app

nrh-cklimas:

n/a

#### What's this PR do?

Adds a function to hide a Sentry environment, and makes the review app teardown management comman call it.

Saves us the hassle of having to hide the environment manually.

Yes.

No.

Yes.

n/a

I had to use the requests module for this because there didn't seem to be a usable Python package for this.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-6139

No.

No.


apps/common/utils.py
```diff
@@ -17,6 +19,35 @@
...
+    """Hides a Sentry environment for a ticket.
+
+    Requires SENTRY_ORGANIZARTION_SLUG, SENTRY_PROJECT_SLUG, and SENTRY_AUTH_TOKEN
+    environment variables, but if they're not present, doesn't raise an
+    exception.
+    """
+    org_slug = os.getenv("SENTRY_ORGANIZATION_SLUG")
+    project_slug = os.getenv("SENTRY_PROJECT_SLUG")
+    auth_token = os.getenv("SENTRY_AUTH_TOKEN")
+    if not org_slug or not project_slug or not auth_token:
```
njharman, line 32: Not suggesting need for change. **Only** alerting you to another pattern that you may not be aware of.  That maybe "reads" better in some contexts.

```
if not all((org_slug, project_slug, auth_token)):
    ....
```

https://docs.python.org/3/library/functions.html#all and any https://docs.python.org/3/library/functions.html#any

apps/common/utils.py
```diff
@@ -17,6 +19,35 @@
...
+    env_name = ticket_id.lower()
+    headers = {"Authorization": f"Bearer {auth_token}"}
+    hide_request = requests.put(
+        f"https://sentry.io/api/0/projects/{org_slug}/{project_slug}/environments/{env_name}/",
+        headers=headers,
+        json={"isHidden": True},
+        timeout=5,
+    )
+    if hide_request.status_code != 200:
+        logger.error("Failed to hide Sentry environment %s: %s", env_name, hide_request)
```
njharman, line 46: Suggest additionally logging some failure information. status_code and "body"? of response.

apps/common/tests/test_utils.py
```diff
@@ -76,6 +80,43 @@ def test_extract_ticket_id_from_branch_name(branch_name: str, expected: str, moc
...
+def test_hide_sentry_environment_noop_without_env_variable(
+    missing_env_variable: str, mocker: pytest_mock.MockerFixture, monkeypatch
+):
+    mock_put = mocker.patch("requests.put")
+    for env_variable in ("SENTRY_ORGANIZATION_SLUG", "SENTRY_PROJECT_SLUG", "SENTRY_AUTH_TOKEN"):
+        monkeypatch.setenv(env_variable, "test")
+    monkeypatch.delenv(missing_env_variable)
+    hide_sentry_environment("test-123")
+    assert not mock_put.called
+
```
njharman, line 118: If we care about the log.warn('... unset; skipping ...'), should check for it.

njharman (approved):

apps/common/tests/test_utils.py
nrh-cklimas: We haven't been testing logging much in rev-engine--I tested it in the failure case because that was the only thing I could think of to assert on. Leaving this comment open to see what others thing. It's obviously not a big deal either way.

apps/common/tests/test_utils.py
```diff
@@ -76,6 +80,43 @@ def test_extract_ticket_id_from_branch_name(branch_name: str, expected: str, moc
         logger_spy.assert_not_called()
 
 
+def test_hide_sentry_environment_happy_path(mocker: pytest_mock.MockerFixture, monkeypatch):
+    mock_put = mocker.patch("requests.put", return_value=mocker.MagicMock(status_code=200))
+    for env_variable in ("SENTRY_ORGANIZATION_SLUG", "SENTRY_PROJECT_SLUG", "SENTRY_AUTH_TOKEN"):
+        monkeypatch.setenv(env_variable, f"test_{env_variable}")
+    hide_sentry_environment("TEST-123")
+    mock_put.assert_called_once_with(
+        "https://sentry.io/api/0/projects/test_SENTRY_ORGANIZATION_SLUG/test_SENTRY_PROJECT_SLUG/environments/test-123/",
```
benjaminEwhite, line 89: I guess `test-123` is a function of the value 87 is called with (specifically, it's lowering of that value). Might as well make that explicit vs. hard coding.

apps/common/utils.py
```diff
@@ -17,6 +19,40 @@
...
+def hide_sentry_environment(ticket_id: str):
+    """Hide a Sentry environment for a ticket.
+
+    Requires SENTRY_ORGANIZARTION_SLUG, SENTRY_PROJECT_SLUG, and SENTRY_AUTH_TOKEN
+    environment variables, but if they're not present, doesn't raise an
+    exception.
+    """
+    org_slug = os.getenv("SENTRY_ORGANIZATION_SLUG")
+    project_slug = os.getenv("SENTRY_PROJECT_SLUG")
+    auth_token = os.getenv("SENTRY_AUTH_TOKEN")
```
benjaminEwhite, line 31: I think I'd prefer to see these values passed vs. being globals.  org slug, project slug, and auth token are (it seems) easily known from calling context, so would prefer that vs. looking to os vars (particularly in context of this being part of a management command).

apps/common/tests/test_utils.py
```diff
@@ -76,6 +80,43 @@ def test_extract_ticket_id_from_branch_name(branch_name: str, expected: str, moc
         logger_spy.assert_not_called()
 
 
+def test_hide_sentry_environment_happy_path(mocker: pytest_mock.MockerFixture, monkeypatch):
+    mock_put = mocker.patch("requests.put", return_value=mocker.MagicMock(status_code=200))
+    for env_variable in ("SENTRY_ORGANIZATION_SLUG", "SENTRY_PROJECT_SLUG", "SENTRY_AUTH_TOKEN"):
+        monkeypatch.setenv(env_variable, f"test_{env_variable}")
```
benjaminEwhite, line 86: This comment may end up being irrelevant depending on outcome of comment further down re: getting values by passing into function rather than from global env.

If we stick with global references, I would prefer to see fixtures and not relying on monkeypatch. I believe something like this would work:

```
@pytest.fixture
def happy_path_settings(mocker: pytest_mock.MockerFixture):
    mocker.patch.object(
        os,
        "environ",
        {
            "SENTRY_ORGANIZATION_SLUG": "test_organization_slug",
            "SENTRY_PROJECT_SLUG": "test_project_slug",
            "SENTRY_AUTH_TOKEN": "test_auth_token",
        },
    )


@pytest.mark.usefixtures("happy_path_settings")
def test_hide_sentry_environment_happy_path(mocker: pytest_mock.MockerFixture):
    mock_put = mocker.patch("requests.put", return_value=mocker.MagicMock(status_code=200))
    hide_sentry_environment("TEST-123")
    mock_put.assert_called_once_with(
        "https://sentry.io/api/0/projects/test_SENTRY_ORGANIZATION_SLUG/test_SENTRY_PROJECT_SLUG/environments/test-123/",
        headers={"Authorization": "Bearer test_SENTRY_AUTH_TOKEN"},
        json={"isHidden": True},
        timeout=5,
    )
```

apps/common/tests/test_utils.py
```diff
@@ -76,6 +80,43 @@ def test_extract_ticket_id_from_branch_name(branch_name: str, expected: str, moc
...
+        headers={"Authorization": "Bearer test_SENTRY_AUTH_TOKEN"},
+        json={"isHidden": True},
+        timeout=5,
+    )
+
+
+def test_hide_sentry_environment_errors_if_put_fails(mocker: pytest_mock.MockerFixture, monkeypatch, caplog):
+    mock_put = mocker.patch("requests.put", return_value=mocker.MagicMock(status_code=404))
+    for env_variable in ("SENTRY_ORGANIZATION_SLUG", "SENTRY_PROJECT_SLUG", "SENTRY_AUTH_TOKEN"):
+        monkeypatch.setenv(env_variable, f"test_{env_variable}")
```
benjaminEwhite, line 99: Fixture strategy above would allow this to be achieved via `pytest.mark.usefixtures`.

apps/common/tests/test_utils.py
```diff
@@ -76,6 +80,43 @@ def test_extract_ticket_id_from_branch_name(branch_name: str, expected: str, moc
...
+    caplog.set_level(logging.WARNING)
+    hide_sentry_environment("TEST-123")
+    mock_put.assert_called_once()
+    assert "Failed to hide Sentry environment test-123" in caplog.text
```
benjaminEwhite, line 103: Make this function of value passed in 101.

apps/common/tests/test_utils.py
```diff
@@ -76,6 +80,43 @@ def test_extract_ticket_id_from_branch_name(branch_name: str, expected: str, moc
...
+
+@pytest.mark.parametrize(
+    "missing_env_variable", ["SENTRY_ORGANIZATION_SLUG", "SENTRY_PROJECT_SLUG", "SENTRY_AUTH_TOKEN"]
+)
+def test_hide_sentry_environment_noop_without_env_variable(
+    missing_env_variable: str, mocker: pytest_mock.MockerFixture, monkeypatch
+):
+    mock_put = mocker.patch("requests.put")
+    for env_variable in ("SENTRY_ORGANIZATION_SLUG", "SENTRY_PROJECT_SLUG", "SENTRY_AUTH_TOKEN"):
+        monkeypatch.setenv(env_variable, "test")
```
benjaminEwhite, line 114: Ditto re: fixture strategy. But here, we could have a fixture called `broekn_env` or something like that and it could inherit from `happy_path_settings` and delete the entry.

apps/common/utils.py
```diff
@@ -17,6 +19,40 @@
...
+    if not all((org_slug, project_slug, auth_token)):
+        logger.warning(
+            "SENTRY_ORGANIZATION_SLUG, SENTRY_PROJECT_SLUG, or SENTRY_API_KEY unset; skipping Sentry environment cleanup"
+        )
+        return
```
benjaminEwhite, line 36: Seems maybe better for calling context to figure this out and pass in. If nothing else, it makes the testing a bit more straightforward, probably allowing us to not need monkeypatch or (as i have suggested) a pytest fixture.

benjaminEwhite (changes requested):

apps/users/management/commands/teardown-review-app.py
```diff
@@ -19,3 +21,16 @@ def handle(self, *args, **options):
 
         delete_cloudflare_cnames(ticket_id)
         tear_down_hookdeck(ticket_id)
+        org_slug = os.getenv("SENTRY_ORGANIZATION_SLUG")
+        project_slug = os.getenv("SENTRY_PROJECT_SLUG")
+        auth_token = os.getenv("SENTRY_AUTH_TOKEN")
```
benjaminEwhite, line 26: Minor, but I think it would be easier to just pass these as arguments when calling command. That seems easier than setting env vars in review app.

Could also just make them required args when configuring parser, which would mean don't need to have conditional check below.

benjaminEwhite (approved):

x110dc (approved):

PR #1779, opened 2025-05-08
DEV-5953 | Make contribution.contributor non-nullable

benjaminEwhite:

**Before merging into main**: Be sure to check [this metabase query](https://metabase-test.revengine.org/question/346-dev-5953-count-of-contributions-with-contributor-id-null) which shows count of contributions that have a null value for `.contributor`. If that count is > 0, this should not be merged.

**Ahead of merging into Heroku envs**: Confirm no contributions with null value for contributor. If any found in a given env, they can be deleted.

#### What's this PR do?

Makes `Contribuiton.contributor` non-nullable, as there is no use case for a contribution without a contributor.

This PR contains a migration that makes `.contributor` non-nullable. Expected run time is low, as all contributions in prod have a value for .contributor.

Mainly to solve for tech debt and remove conditionality around checking for .contributor in several places.

Yes.

No

Yes

N/A

No

N/A

- [DEV-5953](https://news-revenue-hub.atlassian.net/browse/DEV-5953)

No

N/A

[DEV-5953]: https://news-revenue-hub.atlassian.net/browse/DEV-5953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/contributions/models.py
```diff
@@ -316,9 +316,7 @@ class Contribution(IndexedTimeStampedModel):
     # TODO @BW: Remove Contribution.last_payment_date in favor of derivation from payments
     # DEV-4333
     last_payment_date = models.DateTimeField(null=True)
-    # TODO @BW: Make Contribution.contributor non-nullable
```
benjaminEwhite, line 319: on_delete=protect was a judgment call on my part. Think it's right, but open to other viewpoints.

njharman: PROTECT is certainly safest. Even if there is business case/logic for deleting contributor? Probably better to explicitly delete/archive all their contributions, etc. rather than relying on something like CASCADE.

> [on_delete doesn’t create an SQL constraint in the database](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.ForeignKey.on_delete)
lame!

apps/common/tests/test_utils.py
```diff
@@ -308,12 +308,10 @@ def instance_only_needs_amount_update(self, instance, update_data):
 
     @pytest.fixture(
         params=[
-            {"instance": "instance_is_none", "dont_update": [], "action": "created"},
```
leo-dcfa, line 311: We can remove the fixture `instance_is_none` as well

leo-dcfa (approved):
One small request, no need for re-review

apps/common/tests/test_utils.py
benjaminEwhite: 23c3f55b

njharman (approved):

njharman (approved):

x110dc (approved):

PR #1794, opened 2025-06-04
DEV-6182 | CRUD API for EmailCustomizations

benjaminEwhite:

### DIFF
--- a/apps/emails/views.py
+++ b/apps/emails/views.py
@@ -1,12 +1,18 @@
 import datetime
 from dataclasses import asdict
 
+from django.db import models
 from django.http import HttpResponse
 from django.template import TemplateDoesNotExist
 from django.template.loader import get_template
 
+from rest_framework import viewsets
+
+from apps.api.permissions import IsHubAdmin, IsOrgAdmin, IsRpAdmin
 from apps.contributions.models import BillingHistoryItem, Contribution
 from apps.emails.helpers import get_contribution_receipt_email_customizations
+from apps.emails.models import EmailCustomization
+from apps.emails.serializers import EmailCustomizationSerializer
 from apps.organizations.models import RevenueProgram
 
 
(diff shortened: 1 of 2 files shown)


apps/api/permissions.py
```diff
@@ -51,13 +52,9 @@ class IsOrgAdmin(permissions.BasePermission):
     If the user making the request is an org admin, grant permission.
     """
 
-    def has_permission(self, request, view):
-        return all(
-            [
-                role_assignment := getattr(request.user, "roleassignment", False),
```
benjaminEwhite, line 57: Problem here is that `all` doesn't short circuit if this evaluates to False in which case line 58 raises exception.

apps/api/permissions.py
```diff
@@ -66,13 +63,9 @@ class IsRpAdmin(permissions.BasePermission):
     If the user making the request is a RP admin, grant permission.
     """
 
-    def has_permission(self, request, view):
-        return all(
-            [
-                role_assignment := getattr(request.user, "roleassignment", False),
-                role_assignment.role_type == Roles.RP_ADMIN,
-            ]
-        )
```
benjaminEwhite, line 75: Same issue here

apps/api/permissions.py
njharman: I believe `all()` does short circuit.  Lists, tuples, and list comps are evaluated fully / immediately. In this case before being passed on to all() as a parameter. Iterators are not. https://stackoverflow.com/a/53600865

Example (**not** recommended code) 
```
class Thing:
    pass

def delayed(x):
    yield False
    yield x.missing

x = getattr(Thing(), "missing", False)
all(delayed(x))
print("All good")
all(list(delayed(x)))
print("This will not be printed")
```

apps/emails/tests/test_views.py
```diff
@@ -71,3 +75,308 @@ def test_responds_404_when_rp_doesnt_exist(self):
...
+        response = api_client.get(
+            reverse("email-customization-list")
+            + f"?revenue_program={another_orgs_email_customization.revenue_program.pk}"
+        )
+        assert response.status_code == 200
+        assert response.json()["count"] == 0
+        assert response.json()["results"] == []
+
+    @pytest.mark.parametrize("user_fixture", ["hub_admin_user", "org_user", "rp_user"])
+    def test_can_create(
```
njharman, line 256: Consider adding test (probably as separate test method) that verifies can create when 

- rp and email_type exist but email block different
- rp and email_block exist but email_type different
- email_type and email_block exist but rp is different

apps/api/permissions.py
benjaminEwhite: Good point. Thanks for clarifying.

apps/emails/views.py
```diff
@@ -74,3 +81,54 @@ def preview_contribution_email_template(request, template_name: str):
...
+    serializer_class = EmailCustomizationSerializer
+    http_method_names = ["get", "post", "patch", "delete"]
+
+    def get_queryset(self) -> models.QuerySet[EmailCustomization]:
+        rp_id = (
+            int(self.request.query_params.get("revenue_program"))
+            if self.request.query_params.get("revenue_program")
+            else None
+        )
+        filter_kwargs = {}
```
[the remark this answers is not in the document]
benjaminEwhite: 88602f862

apps/emails/tests/test_views.py
benjaminEwhite: @njharman at present we only have one email type and one email block. I could possibly add a test for the third, but that feels a bit like testing third party code (and wouldn't the best/most efficient test just be to assert that there's a unique together constraint?).

x110dc (approved):

PR #1796, opened 2025-06-06
DEV-6233 | Block double sending of receipt emails

benjaminEwhite:

#### What's this PR do?

This PR adds models and logic to prevent double-sending of receipt emails.

- Adds a new model, `TransactionalEmailRecord` which is for recording when transactional emails are sent for a particular contribution
- Refactors and relocates `Contribution.handle_receipt_email` to `TransactionalEmailRecord.handle_receipt_email`, which retains old behavior with new addition that emails will not be sent if there is an extant `TransactionalEmailRecord`.
- Moves `apps.emails.helpers.get_contribution_receipt_email_customizations` into a class method on `RevenueProgram` (I did this as part of resolving circular dependencies that arose).
- Updates `EmailCustomization` email type and email block choices to use TextChoices inheriting classes defined at EmailCustomization class level so they can be referenced elsewhere more easily.

Note that this PR contains a migration for a new model, but that should have a trivial run time.

Ensure we don't double send receipt emails when reprocessing events.

Yes.

N/A

Yes

N/A

No

N/A

[DEV-6233](https://news-revenue-hub.atlassian.net/browse/DEV-6233)

No

No

[DEV-6233]: https://news-revenue-hub.atlassian.net/browse/DEV-6233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emails/tasks.py
```diff
@@ -90,7 +81,7 @@ class SendContributionEmailData(TypedDict):
     rp_name: str
     contributor_name: str
     non_profit: bool
-    fiscal_status: Literal[FiscalStatuses.FISCALLY_SPONSORED, FiscalStatuses.FOR_PROFIT, FiscalStatuses.NON_PROFIT]
+    fiscal_status: Literal["for-profit", "nonprofit", "fiscally sponsored"]
```
benjaminEwhite, line 84: This change is part of getting rid of circular dependencies. I decided i prefer hard coding here to having inline imports in different places.

apps/emails/models.py
```diff
@@ -48,3 +61,47 @@ def as_dict(self) -> "EmailCustomizationValues":
...
+            contribution=contribution,
+            name=EmailCustomization.EmailTypes.CONTRIBUTION_RECEIPT,
+        ).exists():
+            logger.info(
+                "Skipping sending receipt email for contribution %s, already sent",
+                contribution.pk,
+            )
+            return
+        data = generate_email_data(contribution, show_billing_history=show_billing_history)
+        send_receipt_email(data)
```
njharman, line 100: The original called with `.delay()` async. I assume it's sync now to avoid race conditions / want to only create TransactionalEmailRecord if actual email sent. If async was important...

Haven't thought deeply about it. Perhaps move the Check for and creation of TransactionalEmailRecord into send_receipt_email / celery task.

apps/emails/models.py
```diff
@@ -48,3 +61,47 @@ def as_dict(self) -> "EmailCustomizationValues":
...
+        if TransactionalEmailRecord.objects.filter(
+            contribution=contribution,
+            name=EmailCustomization.EmailTypes.CONTRIBUTION_RECEIPT,
+        ).exists():
+            logger.info(
+                "Skipping sending receipt email for contribution %s, already sent",
+                contribution.pk,
+            )
+            return
+        data = generate_email_data(contribution, show_billing_history=show_billing_history)
```
njharman, line 99: Not sure it's needed, just pointing out difference.  Original logged on attempted email send.

` logger.info("Contribution.handle_receipt_email: the parent org (%s) sends emails with NRE", org.id)`

apps/emails/models.py
```diff
@@ -48,3 +61,47 @@ def as_dict(self) -> "EmailCustomizationValues":
...
+class TransactionalEmailNames(models.TextChoices):
+    RECEIPT_EMAIL = "receipt_email", "receipt email"
+
+
+class TransactionalEmailRecord(IndexedTimeStampedModel):
+    """A record of a transactional email sent for a contribution."""
+
+    contribution = models.ForeignKey("contributions.Contribution", on_delete=models.CASCADE)
+    name = models.CharField(max_length=50, choices=TransactionalEmailNames.choices)
+    sent_on = models.CharField(max_length=50, default="pending")
```
njharman, line 75: Consider datetime auto now(UTC) field.  I don't see the usecase / advantage of making this text (or allowing setting time other than now).  When would TransactionalEmailRecord be created when email not sent?

It's error prone to have timestamp that is not a timestamp field (fixture doesn't convert to string):

- test fixture `sent_on=datetime.datetime.now(tz=datetime.timezone.utc),`
- code `sent_on=datetime.datetime.now(datetime.timezone.utc).isoformat()`

apps/emails/models.py
```diff
@@ -1,7 +1,14 @@
...
 
+from django.conf import settings
 from django.db import models
 
+import reversion
+
+from apps.emails.tasks import generate_email_data, send_receipt_email
+
 
 if typing.TYPE_CHECKING:
```
njharman, line 13: Should  # pragma no cover

apps/emails/models.py
```diff
@@ -48,3 +61,47 @@ def as_dict(self) -> "EmailCustomizationValues":
...
+
+class TransactionalEmailRecord(IndexedTimeStampedModel):
+    """A record of a transactional email sent for a contribution."""
+
+    contribution = models.ForeignKey("contributions.Contribution", on_delete=models.CASCADE)
+    name = models.CharField(max_length=50, choices=TransactionalEmailNames.choices)
+    sent_on = models.CharField(max_length=50, default="pending")
+
+    def __str__(self):
+        return f"TransactionalEmailRecord #{self.pk} ({self.name}) for {self.contribution.pk} sent {self.sent_on}"
```
njharman, line 78: Not covered

apps/emails/models.py
```diff
@@ -48,3 +61,47 @@ def as_dict(self) -> "EmailCustomizationValues":
...
+    @staticmethod
+    def handle_receipt_email(contribution, show_billing_history: bool = True) -> None:
+        """Send a receipt email to the contributor and save a record of the email."""
+        logger.info("Running for receipt email for contribution %s", contribution.pk)
+        if not contribution.revenue_program.organization.send_receipt_email_via_nre:
+            logger.info(
+                "Skipping sending receipt email for contribution %s, organization does not send receipt emails via NRE",
+                contribution.pk,
+            )
+            return
```
njharman, line 89: Not covered

njharman (changes requested):

apps/emails/models.py
benjaminEwhite: That was bad tab completion or bad copy paste (or both). Intended to do just that.

f906f1a43071c

apps/emails/models.py
benjaminEwhite: That is correct re: reasoning for no longer using .delay.

In two places where this method gets called, we're already in async context (`StripeWebhookProcessor.handle_payment_intent_succeeded` and `StripeWebhookProcessor.handle_invoice_payment_succeeded`), so I don't think it's adding much to make doubly-async for those.

I don't think it makes sense to move the check into `send_receipt_email` because that task gets used to send sample receipt emails (regardless of any previous send state) via org dashboard in `send_test_email`.

x110dc (changes requested):

apps/emails/models.py
```diff
@@ -48,3 +62,47 @@ def as_dict(self) -> "EmailCustomizationValues":
...
+        if TransactionalEmailRecord.objects.filter(
+            contribution=contribution,
```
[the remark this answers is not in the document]
x110dc: This is better, but I think the same thing could happen (double sends) as an alternate task/thread/process could act in between the check for the record and creating the record itself. And since there are no uniqueness constraints on the record we'd just end up with two records and two emails. 

Why not just include the check for the record in the transaction? I think that would solve the problem. That is move line 109 to above line 99?

x110dc (changes requested):

apps/emails/models.py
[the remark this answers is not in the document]
benjaminEwhite: @x110dc that's a good point. I took another stab at this, inclusive of adding a uniqueness constraint around contribution + name. I believe this will give us the protection we want, but let me know if you see an edge case here.

1c241c69978

Down the road, when we start tracking event id on this model we'll update the uniqueness constraint, but for the first transactional email record type we have (receipt email), we only want to create one record per contribution, so constraining name + contribution works out.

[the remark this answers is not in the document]
benjaminEwhite: Noting additional commits were pushed, but if you look at most recent commit, this should be fully addressed.

apps/emails/models.py
```diff
@@ -52,3 +62,62 @@ def as_dict(self) -> "EmailCustomizationValues":
...
+
+class TransactionalEmailRecord(IndexedTimeStampedModel):
+    """A record of a transactional email sent for a contribution."""
+
+    contribution = models.ForeignKey("contributions.Contribution", on_delete=models.CASCADE)
+    name = models.CharField(max_length=50, choices=TransactionalEmailNames.choices)
+    sent_on = models.DateTimeField(default=timezone.now, help_text="When the email was sent")
+
+    class Meta:
+        unique_together = (
```
x110dc, line 79: Maybe a comment here that this constraint, in its current form, is going to change as we add more records/types?

x110dc (approved):

PR #1797, opened 2025-06-06
DEV-6185: Add email search

nrh-cklimas:

n/a

#### What's this PR do?

Adds email search to activity log Django admin.

Makes it easier to find relevant entries.

Yes.

No.

No because I'm thinking this is a config change. If there's an approach we usually use here, I'm open to adding tests.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-6185

No.

No.

### DIFF
--- a/apps/activity_log/admin.py
+++ b/apps/activity_log/admin.py
@@ -30,6 +30,7 @@ class ActivityLogAdmin(admin.ModelAdmin):
 
     list_display = ("linked_actor_object", "action", "linked_object_object", "get_on")
     actions = None
+    search_fields = ("email",)
 
     def changelist_view(self, request: HttpRequest, extra_context: dict[Any, Any] | None = None):
         extra_context = extra_context or {}


apps/activity_log/admin.py
```diff
@@ -30,12 +31,18 @@ class ActivityLogAdmin(admin.ModelAdmin):
 
     list_display = ("linked_actor_object", "action", "linked_object_object", "get_on")
     actions = None
+    search_fields = ("actor_content_obj",)
```
benjaminEwhite, line 34: I think maybe this isn't doing anything. There is no field with this name, and `get_search_results` as currently implemented would override (since not calling super().get_search_results).

apps/activity_log/admin.py
```diff
@@ -30,12 +31,18 @@ class ActivityLogAdmin(admin.ModelAdmin):
...
 
     def changelist_view(self, request: HttpRequest, extra_context: dict[Any, Any] | None = None):
         extra_context = extra_context or {}
         extra_context["title"] = "Activity logs"
         return super().changelist_view(request, extra_context)
 
+    def get_search_results(
+        self, request: HttpRequest, queryset: QuerySet[ActivityLog], search_term: str
+    ) -> tuple[QuerySet[ActivityLog], bool]:
+        return queryset.filter(actor__email__icontains=search_term), False
```
benjaminEwhite, line 44: Looks like this all could get simplified to just setting search fields as

```
search_fields = ("actor__email__icontains",)
```

and then removing the get_search_results override.

I tried that locally and tests pass and am get expected search results in admin.

benjaminEwhite (changes requested):

x110dc (approved):

PR #1816, opened 2025-07-02
DEV-5903 | add org slug to contribution detail view for switchboard endpoint

benjaminEwhite:

### DIFF
--- a/apps/contributions/views/switchboard.py
+++ b/apps/contributions/views/switchboard.py
@@ -54,7 +54,9 @@ class SwitchboardContributionsViewSet(
 
     permission_classes = [IsSwitchboardAccount]
     http_method_names = ["patch", "post", "get"]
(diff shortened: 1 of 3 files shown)


apps/contributions/views/switchboard.py
```diff
@@ -54,7 +54,9 @@ class SwitchboardContributionsViewSet(
 
     permission_classes = [IsSwitchboardAccount]
     http_method_names = ["patch", "post", "get"]
-    queryset = Contribution.objects.all().select_related("donation_page__revenue_program")
+    queryset = Contribution.objects.all().select_related(
```
benjaminEwhite, line 57: Selecting related organization because we now have organization slug on the serializer used by this view, and provably solves for n+1 when looking at SQL in Django debug toolbar.

x110dc (approved):

PR #1820, opened 2025-07-16
DEV-6368 | email editor editing buttons should be disabled for Free tier

benjaminEwhite:

### DIFF
--- a/spa/src/components/common/SendTestEmail/SendTestEmail.tsx
+++ b/spa/src/components/common/SendTestEmail/SendTestEmail.tsx
@@ -7,9 +7,9 @@ import { useSnackbar } from 'notistack';
 import { ButtonWrapper, Description, Label, Preview } from './SendTestEmail.styled';
 import SystemNotification from '../SystemNotification';
 
-export type SendTestEmailProps = InferProps<typeof SendTestEmailPropTypes>;
+export type SendTestEmailProps = InferProps<typeof SendTestEmailPropTypes> & { editable?: boolean };
 
-const SendTestEmail = ({ rpId, description }: SendTestEmailProps) => {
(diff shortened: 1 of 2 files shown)


spa/src/components/content/emails/EmailsRoute.tsx
```diff
@@ -76,6 +77,8 @@ export const blocks: EmailBlock[] = [
 
 export function EmailsRoute() {
   const { user } = useUser();
+  const isFreeOrg = user?.organizations?.[0]?.plan?.name === PLAN_NAMES.FREE;
+
```
nrh-cklimas, line 81: Stray newline

spa/src/components/content/emails/EmailsRoute.tsx
```diff
@@ -76,6 +77,8 @@ export const blocks: EmailBlock[] = [
 
 export function EmailsRoute() {
   const { user } = useUser();
+  const isFreeOrg = user?.organizations?.[0]?.plan?.name === PLAN_NAMES.FREE;
```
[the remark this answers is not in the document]
nrh-cklimas: We already compute the org plan in line 88, so we may as well use that. The comparison should be something like this:

```tsx
const isFreeOrg = !orgPlan || orgPlan === PLAN_NAMES.FREE;
```

To catch these cases:
- If the user object is undefined while loading (never happens in practice, but better to catch this scenario)
- A malformed user object with no organizations (probably would hit errors before this, but better to be robust)

spa/src/components/content/emails/EmailsRoute.tsx
benjaminEwhite: e3361626

benjaminEwhite: f3022f13

nrh-cklimas (approved):

PR #1841, opened 2025-07-31
DEV-3944: Add page duplication in Django admin

nrh-cklimas:

n/a

#### What's this PR do?

- Adds a general-purpose `duplicate_name` function that derives a dupe name from an original one, trimming it to keep it under a certain length.
- Adds `duplicate` methods to `DonationPage` and `Style`.
- Adds a Django admin action to duplicate selected pages.

Allows the success team to duplicate pages for customers.

Yes.

No.

Yes.

n/a

No.

No.

https://news-revenue-hub.atlassian.net/browse/DEV-3944

No.

No.


apps/pages/models.py
```diff
@@ -110,6 +115,48 @@ def page_url(self) -> str:
...
+            parsed_path = Path(original_image_field.name)
+            getattr(duplicate, field.name).save(
+                # Most OSes have a 255 character limit on file names.
+                duplicate_name(parsed_path.stem, max_length=200, suffix="_copy", truncation_suffix="_")
+                + parsed_path.suffix,
+                ContentFile(original_image_field.file.read()),
+                save=False,
+            )
+            # We need to reset the file position from the read() above.
+            original_image_field.file.seek(0)
```
nrh-cklimas, line 156: I feel unsure about this line, or if there's another way to do the file read that doesn't require a reset. It came up writing unit tests where reading the file in the unit test returned an empty string after this method was called.

apps/pages/tests/test_models.py
```diff
@@ -142,6 +143,59 @@ def test_cannot_delete_when_related_contributions(self, mocker):
...
+
+    def test_duplicate_has_correct_slug(self):
+        original = DonationPageFactory()
+        dupe = original.duplicate()
+        assert dupe.slug == slugify(dupe.name)
+
+    def test_duplicate_always_unpublished(self):
+        original = DonationPageFactory()
+        original.published_date = "2001-01-01"
+        original.save()
```
benjaminEwhite, line 160: Suggest using `live_donation_page` fixture here instead (it has a published date)

benjaminEwhite (changes requested):

apps/pages/models.py
```diff
@@ -110,6 +115,48 @@ def page_url(self) -> str:
...
+        """
+        duplicate = DonationPage()
+
+        # We need to duplicate image fields last because
+        # _get_screenshot_upload_path() relies on model fields like `name` to
+        # exist.
+
+        image_fields = []
+        for field in self._meta.fields:
+            if isinstance(field, SorlImageField) and (original_image_field := getattr(self, field.name)):
```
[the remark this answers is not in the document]
nrh-cklimas: Works for me 178e7656bc1a03f9f897d4f020d257a3ec4346bc

apps/pages/tests/test_models.py
nrh-cklimas: a78e16ffde1fe6ac9ab16a6f09aed4beaffa9c2c

apps/pages/models.py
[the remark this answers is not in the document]
nrh-cklimas: I'm on board with splitting out image duplication, but I feel like there's some merit in the approach of iterating over `self._meta.fields` instead of hardcoding lists of fields. To me, this futureproofs us so that later model changes don't have to update these methods. Let's discuss when we're both back.

